PR #4, opened 2015-07-26
Client/Server Protocol

benjchristensen:
- made write support async with `write(Publisher<Message>)` signature
- move to MessageTypes defined in draft of protocol
- separated the 4 request handlers
- added client protocol implementation
- everything is still String based while waiting on binary encoding and decision of what byte abstraction we will use


benjchristensen:
Non-deterministic unit tests failing build. Merging for now ... active development, and other projects need the snapshot.

PR #21, opened 2015-08-24
Request_Channel

benjchristensen:
Implementation of Request_Channel interaction type as per https://github.com/ReactiveSocket/reactivesocket/issues/30

### DIFF
--- a/src/main/java/io/reactivesocket/FrameType.java
+++ b/src/main/java/io/reactivesocket/FrameType.java
@@ -21,20 +21,22 @@ package io.reactivesocket;
 public enum FrameType
 {
      SETUP(0x01, true),
-     // Messages from Requestor
+     // Messages from Requestor (start a request)
      REQUEST_RESPONSE(0x11, true),
      FIRE_AND_FORGET(0x12, true),
      REQUEST_STREAM(0x13, true),
      REQUEST_SUBSCRIPTION(0x14, true),
-     REQUEST_N(0x15, true),
-     CANCEL(0x16, true),
+     REQUEST_CHANNEL(0x15, true),
+     // Messages from Requestor (mid-stream) 
+     REQUEST_N(0x20, true),
+     CANCEL(0x21, true),
      // Messages from Responder
-     RESPONSE(0x20, false),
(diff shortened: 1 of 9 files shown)


benjchristensen:
@NiteshKant @stevegury @tmontgomery You okay with me adding requestChannel behavior?

We need to update the protocol to support it if so, which I've superficially begun in https://github.com/ReactiveSocket/reactivesocket/pull/42.

benjchristensen:
Note the method signature:

```java
	/**
	 * @param initialPayload
	 *            The first Payload contained in Publisher<Payload> 'payloads'.
	 *            <p>
	 *            This is delivered each time to allow simple use of the first payload for routing decisions.
	 * @param payloads
	 *            stream of Payloads.
	 * @return
	 */
	public abstract Publisher<Payload> handleChannel(final Payload initialPayload, final Publisher<Payload> payloads);
```

The `initialPayload` is unique to this interaction model. It ***greatly*** simplifies the `channelHandler` implementation for routing decisions to have this `initialPayload` arg. See this unit test setup as an example: https://github.com/ReactiveSocket/reactivesocket-java/blob/b34eca96d2e95d9603a1099e3f955f61d43aad6e/src/test/java/io/reactivesocket/ReactiveSocketTest.java#L107

stevegury:
The use case makes sense.
There'll be one thing that we need to be careful about: because of the propagation lag it's possible that you receive "old events" even after sending another "update request". This seems obvious, but I'm sure you may have subtle bug if you don't pay attention to that.
I'm just saying, it's worth noting in the doc.

benjchristensen:
Thanks for the review. 

Good point on documenting behavior. How do you recommend writing that? I tried writing something and it didn't really help, it just effectively said "this is async messaging" :-)

benjchristensen:
I briefly reviewed the `requestChannel` functionality with @stevegury and @NiteshKant so I'm merging this to allow other stuff to progress in master since master is under active development.

Open questions for later change:

- documentation
- backpressure is not hooked in at all, which the protocol doesn't support right now for a stream of requests

PR #25, opened 2015-08-27
Responder instance per DuplexConnection

benjchristensen:
Responder and Requester now have similar design.
Both now eager start when created.


benjchristensen:
Was committed directly to master in https://github.com/ReactiveSocket/reactivesocket-java/commit/70ba001378dcd1a189f7ab1a12972d52c98dc38f

PR #26, opened 2015-08-28
BROKEN: Example of Unit Tests with RxJava2

benjchristensen:
@akarnokd This may be a good piece of code for you to use with your RxJava 2 work right now. I've ported the code to RxJava 2, but there are various random bugs, NPEs, hangs, etc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,8 +20,7 @@ dependencies {
 	compile 'org.reactivestreams:reactive-streams:1.0.0.final'
 	compile 'uk.co.real-logic:Agrona:0.4.2'
 	
-	testCompile 'io.reactivex:rxjava:1.0.13'
-	testCompile 'io.reactivex:rxjava-reactive-streams:1.0.1'
+	testCompile 'io.reactivex:rxjava:2.0.0-DP0-SNAPSHOT'
 	testCompile 'junit:junit-dep:4.10'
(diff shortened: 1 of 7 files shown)


akarnokd:
What I've found:

  - `RequesterTest.java` doesn't compile
  - `PublisherUtil.just` doesn't cancel its subscription so a request after s.onComplete() sends a new event. This broke `flatMap` because of the contract violation. I'll strengthen `flatMap` for this case. Please use the standard `just`, `error` and `fromCallable` factory methods instead.
  - `testRequestStreamSuccessTake2AndCancel` fails due to a comparison error: `Frame` !== `PayloadImpl` but `PayLoadImpl` === `Frame` I presume. The other tests fail because of this at first run.
```
java.lang.AssertionError: Values at position 0 differ; Expected: Frame => Stream ID: 2 Type: NEXT Payload Data: data: "hello", Actual: io.reactivesocket.TestUtil$PayloadImpl@1d29cf23
	at io.reactivex.subscribers.TestSubscriber.fail(TestSubscriber.java:287)
	at io.reactivex.subscribers.TestSubscriber.assertValueSequence(TestSubscriber.java:510)
	at io.reactivesocket.internal.RequesterTest.testRequestStreamSuccessTake2AndCancel(RequesterTest.java:166)
```

PR #72, opened 2016-03-27
ReactiveSocket interface

robertroeser:
- updated to agrona 0.4.13
- pinned to the correct rxjava 2 snapshot so unit tests pass
- refactored the ReactiveSocket class into an interface and implementation class


src/main/java/io/reactivesocket/ReactiveSocketImpl.java
```diff
@@ -0,0 +1,477 @@
...
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
+
+import static io.reactivesocket.LeaseGovernor.NULL_LEASE_GOVERNOR;
+
+/**
+ * An implementation of {@link ReactiveSocket}
+ */
+public class ReactiveSocketImpl implements ReactiveSocket {
```
stevegury: I like Netty's convention of naming default implementation as `Default*`, in that case, `DefaultReactiveSocket`.

stevegury:
Other that the naming that I would like to improve, LGTM.

PR #82, opened 2016-05-04
adding callback on shutdown

robertroeser:
As a callback that is called when a reactive socket is closed.

### DIFF
--- a/src/main/java/io/reactivesocket/DuplexConnection.java
+++ b/src/main/java/io/reactivesocket/DuplexConnection.java
@@ -15,6 +15,7 @@
  */
 package io.reactivesocket;
 
+import io.reactivesocket.internal.rx.EmptySubscription;
 import io.reactivesocket.rx.Completable;
 import io.reactivesocket.rx.Observable;
 import org.reactivestreams.Publisher;
@@ -32,6 +33,7 @@ public interface DuplexConnection extends Closeable {
 
 	default void addOutput(Frame frame, Completable callback) {
         addOutput(s -> {
+            s.onSubscribe(EmptySubscription.INSTANCE);
             s.onNext(frame);
             s.onComplete();
         }, callback);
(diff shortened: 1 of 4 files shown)


src/main/java/io/reactivesocket/ReactiveSocket.java
```diff
@@ -95,6 +95,13 @@ public void error(Throwable e) {
     void onRequestReady(Completable c);
 
     /**
+     * Invoked when ReactiveSocket is closed
+     *
+     * @throws IllegalStateException if this method is called more than once per instance
+     */
+    void onShutdown(Completable c);
```
stevegury: Instead of returning void, I would return the previous Completable (and allow changing the callback).

src/main/java/io/reactivesocket/DuplexConnection.java
```diff
@@ -15,6 +15,10 @@
  */
 package io.reactivesocket;
 
+<<<<<<< HEAD
+import io.reactivesocket.internal.rx.EmptySubscription;
+=======
+>>>>>>> 119ef1e554ee5e0a195fd44fe7fdc19b77b5e1fd
```
stevegury: Merge conflict!

src/main/java/io/reactivesocket/ReactiveSocket.java
NiteshKant: Actually, why to ask for a `Completable`? This is like a close future for the socket, which is a single instance that anyone can subscriber/listen to. So, why not have something like:

```java
Completable shutdownListener();
```

stevegury: I think it's a good idea.

robertroeser:
@stevegury @NiteshKant changed it so that it returns a Publisher that is completed when the ReactiveSocket is closed

NiteshKant:
@robertroeser doesn't it make sense to have a `Completable` instead of `Publisher<Void>` as you had before?

robertroeser:
@NiteshKant ok switched it back to a Completable

NiteshKant:
ok revelation for me `Completable` here isn't `RxJava` `Completable` 😭

PR #85, opened 2016-05-31
Introduce Connector and availability for Connection

stevegury:
**Problem**
The DuplexConnection class doesn't expose any way to probe its state,
this is somewhat problematic and doesn't fit well in the current
model where the availability is the composable way of chaining state.

The ReactiveSocketFactory create a ReactiveSocket from an `address`
(currently SocketAddress). There are places in the code where the concept
of address is not needed and it leads to implicit dependency on it.

**Solution**
Add a `double availability()` method to the DuplexConnection interface,
most implementation of this method will be straightforward (just
returning 0.0 if the underlying resource is unavailable, 1.0 otherwise).

Split the concept of ReactiveSocketFactory in two, the Factory and the
Connector. The Connector is responsible for creating the ReactiveSocket
from the address, and the factory can create a ReactiveSocket without
any argument.

**Modification**
I added a `ReactiveSocketProxy` helper class, which ease the task of wrapping
a ReactiveSocket instance for extending its behavior.

### DIFF
--- a/src/main/java/io/reactivesocket/util/ReactiveSocketProxy.java
+++ b/src/main/java/io/reactivesocket/util/ReactiveSocketProxy.java
@@ -69,7 +69,6 @@ public class ReactiveSocketProxy implements ReactiveSocket {
                 child.requestStream(payload).subscribe(subscriber);
             };
         }
-
     }
 
     @Override
@@ -82,7 +81,6 @@ public class ReactiveSocketProxy implements ReactiveSocket {
                 child.requestSubscription(payload).subscribe(subscriber);
             };
         }
-
     }
 
     @Override
@@ -95,7 +93,6 @@ public class ReactiveSocketProxy implements ReactiveSocket {
                 child.requestChannel(payloads).subscribe(subscriber);
             };
         }
-
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


NiteshKant:
👍  for removing the address argument from the factory, this is similar to what I converged to in RxNetty!

I did not see the connector class though

stevegury:
@NiteshKant My bad, I added the missing files.

src/main/java/io/reactivesocket/util/ReactiveSocketProxy.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Wrapper/Proxy for a ReactiveSocket.
+ * This is useful when we want to override a specific method.
+ */
+public class ReactiveSocketProxy implements ReactiveSocket {
+    protected final ReactiveSocket child;
+    private final Function<Subscriber<? super Payload>, Subscriber<? super Payload>> subscriberWrapper;
+
+    public ReactiveSocketProxy(ReactiveSocket child, Function<Subscriber<? super Payload>, Subscriber<? super Payload>> subscriberWrapper) {
```
NiteshKant: Wonder if extension is a better approach here as opposed to a single subscriber wrapper function, which only is applicable to methods that return `Publisher<Void>`

NiteshKant:
Apart from the minor comment on `ReactiveSocketProxy` I think this is good.

src/main/java/io/reactivesocket/util/ReactiveSocketProxy.java
stevegury: You're right, it's most of the time a cleaner approach.
That been said, this is pretty convenient to be able to do that here.

PR #86, opened 2016-05-31
error message transformer

robertroeser:
Added a transformer in the Frame.Error class to suck in the whole stack trace and send that as an error message payload optionally.


src/main/java/io/reactivesocket/Frame.java
```diff
@@ -321,6 +323,31 @@ public static String dataMimeType(final Frame frame)
...
+        // TODO Make this plugin, and also make it so that a error message transformer can be
+        // TODO passed in with an application
+        static {
+            if (Boolean.getBoolean("io.reactivesocket.debugError")) {
+                DEFAULT_ERROR_MESSAGE_TRANSFORMER = throwable -> {
+                    StringWriter writer = new StringWriter();
+                    PrintWriter printWriter = new PrintWriter(writer);
+                    throwable.printStackTrace(printWriter);
+                    String data = writer.toString();
+                    byte[] bytes = data.getBytes(Charset.forName("UTF-8"));
```
NiteshKant: nit: You can use `StandardCharsets.UTF_8` instead of charset lookup.

NiteshKant:
I am not a big fan of sending stacktraces back to the caller over the network but if you think it is useful 👍

src/main/java/io/reactivesocket/Frame.java
```diff
@@ -321,6 +323,31 @@ public static String dataMimeType(final Frame frame)
...
+
+        // TODO Make this plugin, and also make it so that a error message transformer can be
+        // TODO passed in with an application
+        static {
+            if (Boolean.getBoolean("io.reactivesocket.debugError")) {
+                DEFAULT_ERROR_MESSAGE_TRANSFORMER = throwable -> {
+                    StringWriter writer = new StringWriter();
+                    PrintWriter printWriter = new PrintWriter(writer);
+                    throwable.printStackTrace(printWriter);
+                    String data = writer.toString();
```
stevegury: I would prepend `throwable.getMessage() == null ? "" : throwable.getMessage();` to the `data` String (sperated by new line) so that in both cases the data starts by the same thing.

stevegury:
👍

NiteshKant:
Closing this in preparation for the latest code push (ref #175)

PR #87, opened 2016-06-02
Unknown Stream Id frame shutdown connection.

NiteshKant:
If a frame with non-existent (terminated or unknown) stream Id arrives on a connection, the connection was shutdown.
This change sends the error to the error stream for the connection instead and hence ignores such a frame.

### DIFF
--- a/src/main/java/io/reactivesocket/internal/Requester.java
+++ b/src/main/java/io/reactivesocket/internal/Requester.java
@@ -923,23 +923,8 @@ public class Requester {
                         streamSubject = streamInputMap.get(streamId);
                     }
                     if (streamSubject == null) {
-                        if (streamId <= streamCount) {
-                            // receiving a frame after a given stream has been cancelled/completed,
-                            // so ignore (cancellation is async so there is a race condition)
-                            return;
-                        } else {
-                            // message for stream that has never existed, we have a problem with
-                            // the overall connection and must tear down
-                            if (frame.getType() == FrameType.ERROR) {
-                                String errorMessage = getByteBufferAsString(frame.getData());
-                                onError(new RuntimeException(
-                                    "Received error for non-existent stream: "
-                                        + streamId + " Message: " + errorMessage));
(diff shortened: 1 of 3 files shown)


stevegury:
Do you have more context about the bug, did the code in the `else` branch was executed?

NiteshKant:
@stevegury I have added a UnitTest to test this behavior. The test fails without this change.

stevegury:
My question was about the bug (I want to understand the problem), what message did you see in the log? "Received error for non-existent stream" or "Received message for non-existent stream".

The code currently seems correct, it's invalid to receive a message for a streamId which hasn't yet been generated, we should tear down the connection.
I think the problem is that we reuse the Responder/Reactivesocket while we shouldn't, it could be that we are not closing resources properly or that the calling library doesn't cope well with failures.

NiteshKant:
>>> The code currently seems correct, it's invalid to receive a message for a streamId which hasn't yet been generated, we should tear down the connection.

I don't believe that is correct. The [protocol suggests](https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md#handling-the-unexpected):

>This protocol attempts to be very lenient in processing of received frames and SHOULD ignore conditions that do not make sense given the current context.

As for this bug, it came up during an internal test and I am unsure whether it was an error frame or a data frame. The code appeared wrong to me based on the spec, so I removed that section and just sent the ignore message to the error stream instead.

>>> it could be that we are not closing resources properly or that the calling library doesn't cope well with failures.

You are correct, the issue was _also_ that the library was reusing a close socket, which I intend to fix next. Although, I do feel that this change here is valid according to the spec, i.e. we should not tear down the connection if we did not recognize a stream Id.

NiteshKant:
Offline discussion result:

- We should close the connection if we see a stream ID that was never generated.
- We should ignore the frame if we see a stream ID that was cancelled.

The above is what the current code does, closing this PR.

PR #88, opened 2016-06-03
Frame instance was referenced longer than required.

NiteshKant:
Problem:

`handleXXX` methods in `Responder` were closing over the passed `requestFrame` and using it later in the lifecycle of request processing.
 `Frame` objects and the underlying buffers are not designed to be retained after the scope of the parent method as these objects are threadlocal and reused. This causes issues when the frame object is referenced later in the request processing (eg: `cleanup()`)

 Solution:

 The only reason frame object was retained was to get the stream Id. This change pre-fetches the `streamId` and uses that from within the processing closure.

 Result:

 No more issue with frame access.


stevegury:
👍 This looks good, thanks for the fix!

PR #89, opened 2016-06-08
Aggregation of reactivesocket-java-impl into reactivesocket-java

stevegury:
**Problem**
The current multi repo organization of reactivesocket is annoying to use, it
forces us to publish jars to test libraries.

**Solution**
Migrate all the subprojects from reactivesocket-java-impl into reactivesocket-java.
To avoid having version number conflict, and to start a convention, I renamed the
projects like this:
- reactivesocket => rs-core
- reactivesocket-aeron => rs-transport-aeron
- reactivesocket-jsr-356 => rs-transport-jsr-356
- reactivesocket-local => rs-transport-local
- reactivesocket-mime-types => rs-mime-types
- reactivesocket-netty => rs-transport-netty
- reactivesocket-test => rs-test

**Note**
Both git histories are kept.


NiteshKant:
This is cool and will reduce a bunch of headaches :)

Few questions:

- Are we using rs-jsr-356 anywhere? 
- rs-transport-netty sounds wrong as it is along with rs-transport-aeron which is a protocol and -netty is not. I suggested in a different PR that we should split it into rs-transport-tcp and rs-transport-websocket. -websocket can depend on -tcp (since it is over tcp) if we need to share common code.
- Should we create a rs-skunkworks (rs-experimental?) module/project which can contain -jsr-356 and other experimental projects that we have been talking about.
- Should we keep "-java" in the module names?

You do not have to do all the changes in this PR, we can merge this (I don't think anyone can review :)) and then do incremental changes. I just thought it will be good to discuss this since we are doing the restructuring.

stevegury:
> Are we using rs-jsr-356 anywhere?

I don't think so, @rdegnan can confirm.
I agree that if it's the case, we should move it elsewhere.

> rs-transport-netty sounds wrong as it is along with rs-transport-aeron which is a protocol and -netty is not. I suggested in a different PR that we should split it into rs-transport-tcp and rs-transport-websocket. -websocket can depend on -tcp (since it is over tcp) if we need to share common code.

Again, I agree. I hesitated to do exactly this but I wanted to limit the already large PR.
I will submit a subsequent PR with your proposal.

> Should we create a rs-skunkworks (rs-experimental?) module/project which can contain -jsr-356 and other experimental projects that we have been talking about.

Do you mean a subdirectory containing multiple submodules?

- Should we keep "-java" in the module names?

We don't have "-java" in the module name, only in the name of the repo.
Under the groupId "io.reactivesocket" we currently generate those artifact-id:

- rs-core
- rs-transport-aeron
- rs-transport-jsr-356
- rs-transport-local
- rs-mime-types
- rs-transport-netty
- rs-test

It will change after I split `-netty` into `-tcp` and `-websocket`

stevegury:
Per offline discussion, I renamed the 'rs-*' project into 'reactivesocket-*'

PR #90, opened 2016-06-11
Remove the jsr-356 Transport.

stevegury:
Despite being functional, we don't use the transport and
we don't plan to in a near future.


stevegury:
Merging myself since it isn't anything contentious.

PR #91, opened 2016-06-11
Split reactivesocket-transport-netty into 2 projects.

stevegury:
**Problem**
`reactivesocket-transport-netty` isn't an implementation of a transport but
two "tcp" and "websocket". Furthermore, Netty is an implementation details
rather than a transport.

**Solution**
Split `reactivesocket-transport-netty` into two transport libraries:

- `reactivesocket-transport-tcp`
- `reactivesocket-transport-websocket`

Introduce a common library `reactivesocket-netty` which only contains two
classes `MutableDirectByteBuf` and `NettyDuplexConnection` that are used in
both transport libraries.

**Note**
I didn't touch `ReactiveSocketServerHandler` which is discussed in a PR
on the old repo.


NiteshKant:
I would suggest we do not have a different module `reactivesocket-netty` specially since we only have two modules now tcp and websocket and websocket can depend on tcp (logically that is the case). So, the common classes can reside in tcp module. Furthermore, I expect to move these modules to use RxNetty in the near future which will further reduce any requirement of a common module.

stevegury:
@NiteshKant Refactored as you suggested.

stevegury:
I fixed the tests (I forgot to update the imports).
This is now compiling on TravisCI.

PR #92, opened 2016-06-14
more reactor re-factoring

robertroeser:


NiteshKant:
@robertroeser I think we can close this one since you are working on this in a different branch?

PR #94, opened 2016-06-16
Refactor ReactiveSocketServerHandler to be not shareable.

stevegury:
**Problem**
There's a memory leak in `ReactiveSocketServerHandler`, it keep adding entries
in the `duplexConnections` map but never remove them.

**Solution**
Instead of having only one `ReactiveSocketServerHandler`, and manage resources
manually, we let Netty allocate one instance per Channel. Then no resource
management is necessary.

**Modifications**
I refactored all the uages of `ReactiveSocketServerHandler` whithout changing
the logic. I also got rid of the method
```
public void channelReadComplete(ChannelHandlerContext ctx) {
    ctx.flush();
}
```
since we only use `writeAndFlush` in the DuplexConnection (we're sure there's
nothing to flush).

### DIFF
--- a/reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/server/ReactiveSocketServerHandler.java
+++ b/reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/server/ReactiveSocketServerHandler.java
@@ -16,9 +16,7 @@
 package io.reactivesocket.transport.tcp.server;
 
 import io.netty.buffer.ByteBuf;
-import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
-import io.netty.channel.ChannelId;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelPipeline;
 import io.netty.handler.codec.LengthFieldBasedFrameDecoder;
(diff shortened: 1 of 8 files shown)


reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/server/ReactiveSocketServerHandler.java
```diff
@@ -28,51 +26,51 @@
...
     public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
         ChannelPipeline cp = ctx.pipeline();
         if (cp.get(LengthFieldBasedFrameDecoder.class) == null) {
-            ctx
-                .pipeline()
-                .addBefore(
-                    ctx.name(),
-                    LengthFieldBasedFrameDecoder.class.getName(),
-                    new LengthFieldBasedFrameDecoder(Integer.MAX_VALUE >> 1, 0, BitUtil.SIZE_OF_INT, -1 * BitUtil.SIZE_OF_INT, 0));
+            LengthFieldBasedFrameDecoder frameDecoder =
```
NiteshKant: Minor: It doesn't make much of a difference but we can move this to server initialization i.e. add these handlers upfront using `ChannelInitializer` instead of doing it on `handlerAdded`

NiteshKant:
👍  I will merge this change, my minor comment can be addressed later.

reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/server/ReactiveSocketServerHandler.java
stevegury: Good idea, we'll improve that later.

PR #95, opened 2016-06-20
Introducing `PublisherFunctions`

NiteshKant:
#### Background

I was looking at a few functions like `retry` and `timeout` and saw a few issues related to cancellation and insufficient guard against emissions post termination. Instead of fixing them locally (which isn't accessible to the rest of the project), I tried to provide a utility class that contains commonly used functions that can be applied on to a `Publisher`.

This PR is more or less a Proof of concept and I see the following value in this change:

- We have more visibility into which kind of functions are we using internal to our code base.
- This can be considered as the first phase before introducing a function composition library (like reactor which is pursued in a different branch).

The positives of the above two things are that we do not need to commit to a function composition library upfront without knowing how much of that library will we be actually using. This is more important because we do not intend to expose a function composition library in our public APIs and if we use one, it will be only for internal use.

This approach was also briefly mentioned offline by @stevegury earlier.

#### Change

This is just a proof of concept and hence I have not gone about changing all the other custom functions in the code. If we get some agreement on this change, I think we should do the following:

- Get rid of any custom APIs (`Completable`, `Disposable`, etc.) from our code and replace with a `Publisher`.
- Refactor the code to use common functions from `PublisherFunctions` instead of defining inline.

The biggest benefit I see of this is that it is a general cleanup which is anyways required and hence can be done in the master branch. 
Without taking this approach, I can see the reactor branch drifting considerably from master and hence  become more and more difficult to merge back.


NiteshKant:
PS: I have intentionally not added unit tests for `PublisherFunctions` because currently the tests use RxJava 2.x DP artifacts which I do not want to use any further. I will send another PR to move to using RxJava 1.x in the meanwhile with `RxReactiveStreams` bridge, only for unit tests.

reactivesocket-core/src/main/java/io/reactivesocket/internal/PublisherFunctions.java
```diff
@@ -0,0 +1,316 @@
...
+        @Override
+        public void onComplete() {
+            if (done.compareAndSet(false, true)) {
+                doOnComplete();
+            }
+        }
+
+        @Override
+        public void cancel() {
+            done.set(true);
```
stevegury: Shouldn't it be:
```
if (done.compareAndSet(false, true)) {
  super.cancel();
}
```

reactivesocket-examples/src/main/java/io/reactivesocket/examples/EchoClient.java
```diff
@@ -45,10 +46,10 @@ public static void main(String... args) throws Exception {
...
             new TcpReactiveSocketConnector(new NioEventLoopGroup(8), setupPayload, System.err::println);
 
-        ReactiveSocket client = Builder.instance()
-            .withSource(source(address))
-            .withConnector(tcp)
-            .build();
+        ReactiveSocket client = ClientBuilder.instance()
+                                             .withSource(source(address))
+                                             .withConnector(tcp)
+                                             .build();
```
stevegury: I'm voting for the previous code-style (less horizontal space wasted).

reactivesocket-core/src/main/java/io/reactivesocket/internal/PublisherFunctions.java
```diff
@@ -0,0 +1,316 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+
+/**
+ * A set of utility functions for applying function composition over {@link Publisher}s.
+ */
+public final class PublisherFunctions {
```
stevegury: Why not using the Java convention of pluralizing, i.e. renaming `Publishers`.

reactivesocket-core/src/main/java/io/reactivesocket/internal/PublisherFunctions.java
```diff
@@ -0,0 +1,316 @@
...
+     * timeout does <em>not</em> apply for further emissions.
+     *
+     * @param source Source to apply the timeout on.
+     * @param timeoutSignal Source after termination of which, {@code source} will be cancelled, if not already done.
+     *
+     * @param <T> Type of items emitted by {@code source}.
+     *
+     * @return A new {@code Publisher} with timeout applied.
+     */
+    public static <T> Publisher<T> timeout(Publisher<T> source, Publisher<Void> timeoutSignal) {
```
stevegury: An override with `Publisher<T> source, long interval, TimeUnit timeUnit, ScheduledExecutorService scheduler` would be easier to use.

NiteshKant: Agreed it will be easier to use. However, my intention is to keep the API surface area smaller and hence I did not add overrides. Instead I provided a `timer()` method which in connection with this can be used as 
`timeout(source, timer(executor, interval, unit))`.

Having said that, I don't think we should provide ways to specify timeouts/retries on public APIs. Any practical application can not use ReactiveSocket without using a function composition library. When they use one, they can use that to provide timeouts and retries. We should just make sure that we have cold sources with no side-effects.

That is my opinion, but I am open to adding more methods here when convinced 😄

reactivesocket-examples/src/main/java/io/reactivesocket/examples/EchoClient.java
NiteshKant: After we are done with considerable cleanups, I think we should add a checkstyle file and re-format everything.

reactivesocket-core/src/main/java/io/reactivesocket/internal/PublisherFunctions.java
stevegury: Fair enough.

reactivesocket-core/src/main/java/io/reactivesocket/internal/PublisherFunctions.java
NiteshKant: Done

reactivesocket-examples/src/main/java/io/reactivesocket/examples/EchoClient.java
NiteshKant: Reverted formatting.

reactivesocket-core/src/main/java/io/reactivesocket/internal/PublisherFunctions.java
NiteshKant: Done

NiteshKant:
@stevegury I have incorporated the review comments.

PR #96, opened 2016-06-20
Replace usage of RxJava2 with RxJava1 and RxReactiveStreams.

NiteshKant:
#### Problem

Tests in reactivesocket-core were using RxJava2 with a DP release. RxJava2 is currently not stable.

#### Solution

Moved to RxJava1 with RxReactiveStreams. When RxJava2 becomes stable, we can see whether we need to use it again.

### DIFF
--- a/reactivesocket-core/build.gradle
+++ b/reactivesocket-core/build.gradle
@@ -1,3 +1,2 @@
 dependencies {
-    testCompile 'io.reactivex:rxjava:2.0.0-DP0-20151003.214425-143'
 }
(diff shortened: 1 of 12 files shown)


benjchristensen:
ugh

NiteshKant:
@benjchristensen this is just for unit tests.

benjchristensen:
Yes I know. That's why I chose v2 as it wasn't production critical but allowed cleaner code.

NiteshKant:
Modified the description to better depict the change.

@benjchristensen @stevegury if we aren't ok with this, lets close it.

stevegury:
There was tons of progress in RxJava 2 recently, my hope is that it will be usable sooner than I'd have expected.
I would prefer sticking to RxJava 2, and if it's too late, temporarily use Reactor.

benjchristensen:
Even until RxJava 2 matures, I didn't find any issues while writing this code originally with just using the snapshot it currently uses. Since ReactiveSocket is all about ReactiveStreams, it was incredibly annoying always going to and from RxJava v1 Observables, and made debugging difficult as it was transformed so many times.

NiteshKant:
Ok, closing the PR. Thanks guys for the feedback!

PR #99, opened 2016-06-22
Wire `onSubscribe` in *Connector.

stevegury:
**Problem**
Instead of manually requesting in the `onSubscribe`, it's better
to actually pass the `Subscription` to the `Subscriber`.
The *Connector code was a little bit confusing because the `Subscriber`
like the `Subscription` were names `s`.

**Solution**
Rename the `Subscriber` -> `subscriber`, and pass the subscription.


NiteshKant:
Ok so this is working today because no where in the code we are actually validating backpressure for this invocation i.e. if we receive an `onNext` without actually requesting 😞 
So, it is the correct thing to do but doesn't really make a difference today, rite?

stevegury:
The main problem here, is that we're requesting 1 DuplexConnection without ensuring that the Publisher chain of the ReactiveSocket is initialized. We were breaking the chain by introducing a `request(1)`.

For instance, this fixes the problem when you receive an `onError` without having an `onNext` which could fail due to the lack of initialization.

NiteshKant:
Yes I understood that. My surprise was "why was it working till now?" :) Which I think is because we do not do this sanity check: "Did I get an `onNext` before `onSubscriber()`?"

Anyways, this change looks good, merging..

PR #101, opened 2016-06-22
Using RxNetty for tcp transport

NiteshKant:
#### Problem

Current TCP transport implementation lacks a few critical features around insights and network flow control.

#### Solution

Since RxNetty already has these features, it makes sense to use it.

__This is a backward incompatible change as it removes some public classes__

`EchoClient` in examples module contains the usage example.


reactivesocket-examples/src/main/java/io/reactivesocket/examples/EchoClient.java
```diff
@@ -15,46 +15,59 @@
...
             .build();
 
         Unsafe.awaitAvailability(client);
 
         Payload request = TestUtil.utf8EncodedPayload("Hello", "META");
-        Payload response = Unsafe.blockingSingleWait(client.requestResponse(request), 1, TimeUnit.SECONDS);
-
-        System.out.println(response);
+        RxReactiveStreams.toObservable(client.requestResponse(request))
+                         .map(payload -> {
```
yschimke: This example would be a lot nicer with a utility function

.map(PayloadUtil::dataAsString)

Also 2000 error nit that new String() isn't safe without specifying encoding

reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/client/TcpReactiveSocketConnector.java
```diff
@@ -1,81 +1,50 @@
...
 import java.util.function.Consumer;
+import java.util.function.Function;
 
-/**
- * An implementation of {@link ReactiveSocketConnector} that creates Netty TCP ReactiveSockets.
- */
 public class TcpReactiveSocketConnector implements ReactiveSocketConnector<SocketAddress> {
-    private final ConnectionSetupPayload connectionSetupPayload;
+
+    private final ConcurrentMap<SocketAddress, TcpReactiveSocketFactory> socketFactories;
```
stevegury: I don't understand that part, the Connector shouldn't do any caching, its role is to asynchronously create a ReactiveSocket from an identifier (e.g. SocketAddress).

reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/ReactiveSocketFrameCodec.java
```diff
@@ -0,0 +1,63 @@
...
+        try {
+            if (msg instanceof ByteBuf) {
+                buffer.wrap((ByteBuf) msg);
+                frame.wrap(buffer, 0);
+                ctx.fireChannelRead(frame);
+            } else {
+                super.channelRead(ctx, msg);
+            }
+        } finally {
+            ReferenceCountUtil.release(msg);
```
yschimke: Should you be releasing in the case where you call super.channelRead?

reactivesocket-transport-websocket/build.gradle
```diff
@@ -1,6 +1,7 @@
 dependencies {
     compile project(':reactivesocket-core')
     compile project(':reactivesocket-transport-tcp')
+    compile 'io.netty:netty-codec-http:4.1.0.Final'
```
yschimke: Woot!

reactivesocket-examples/src/main/java/io/reactivesocket/examples/EchoClient.java
NiteshKant: Agreed. 
My just another rant about `ByteBuffer` API, as compared to netty's `ByteBuf` API which enables [this](https://github.com/ReactiveX/RxNetty/blob/0.5.x/rxnetty-examples/src/main/java/io/reactivex/netty/examples/http/helloworld/HelloWorldClient.java#L81):

```java
.map(bb -> bb.toString(Charset.defaultCharset())))
```
😞

reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/client/TcpReactiveSocketConnector.java
NiteshKant: This code is caching `HttpClient` instead of creating a new one for every connect. `HttpClient` has 1:1 mapping to a `SocketAddress` or a stream of `SocketAddress` but you can not connect to an arbitrary socket.

We can discuss this later but IMO we do not need `ReactiveSocketConnector` contract, we can just have `ReactiveSocketFactory`.

reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/ReactiveSocketFrameCodec.java
NiteshKant: Good point, it is only needed when emitting a `Frame`.

reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/ReactiveSocketLengthCodec.java
```diff
@@ -0,0 +1,28 @@
...
+
+package io.reactivesocket.transport.tcp;
+
+import io.netty.handler.codec.LengthFieldBasedFrameDecoder;
+import org.agrona.BitUtil;
+
+public class ReactiveSocketLengthCodec extends LengthFieldBasedFrameDecoder {
+
+    public ReactiveSocketLengthCodec() {
+        super(Integer.MAX_VALUE >> 1, 0, BitUtil.SIZE_OF_INT, -1 * BitUtil.SIZE_OF_INT, 0);
```
yschimke: This doesn't match the spec.  I think Integer.MAX_VALUE is enough here because its 0 left padded with 1 bit.

reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/client/TcpReactiveSocketConnector.java
stevegury: I understand now that this is coming from a limitation of RxNetty API, but the code you wrote for `TcpReactiveSocketFactory` should actually be used for `TcpReactiveSocketConnector`, and you don't need `TcpReactiveSocketFactory` anymore.

NiteshKant: You are correct, for this change it is not required.

In the (slightly?)longer term, I think we don't need the connector as up the chain `LoadBalancer` keeps a list of `ReactiveSocketFactory` and never really calls `connect(SocketAddress)` which is the correct thing to do. The only reason `ReactiveSocketConnector` is required today because it provides the source (list of Sockets) in `ClientBuilder` to `ReactiveSocketFactory` mapping. 
I suggest that we use the source as `Publisher<List<ReactiveSocketFactory>>` which will remove the need of `ReactiveSocketConnector` and hence the ugly mapping inside _any_ `ReactiveSocketConnector` implementation as generally clients bind to an address. 
As I said, we can discuss this in a separate PR. 

If you feel strongly about removing this abstraction in this PR, I will remove it but I think `ReactiveSocketFactory` should be the correct abstraction which is bound to a single `SocketAddress`

stevegury: API-wise I think it's more user-friendly to ask for:
- `Publisher<List<SocketAddress>>` as the source
- A `ReactiveSocketConnector` as the code responsible for transforming the address into a `ReactiveSocket`

Rather than `Publisher<List<ReactiveSocketFactory>>`.

It allows separation of concern between something like the discovery mechanism (Eureka, DNS...) and the Transport (TCP, WebSocket, Aeron).

You're right than the `ReactiveSocketFactory` provides barely more abstraction than a raw `ReactiveSocketConnector` (it only hides the SocketAddress). But I believe it makes the API cleaner.

reactivesocket-examples/src/main/java/io/reactivesocket/examples/EchoClient.java
NiteshKant: Done.

reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/client/TcpReactiveSocketConnector.java
NiteshKant: I have removed the factory for now, will send another PR to explain what I have in mind.

reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/ReactiveSocketFrameCodec.java
NiteshKant: Done

reactivesocket-transport-tcp/src/main/java/io/reactivesocket/transport/tcp/ReactiveSocketLengthCodec.java
NiteshKant: Done

NiteshKant:
@yschimke @stevegury addressed all comments and updated to [latest RxNetty](https://github.com/ReactiveX/RxNetty/releases/tag/v0.5.2-rc.3)

reactivesocket-test/src/main/java/io/reactivesocket/test/TestUtil.java
```diff
@@ -63,6 +63,13 @@ public static Payload utf8EncodedPayload(final String data, final String metadat
         return new PayloadImpl(data, metadata);
     }
 
+    public static String dataAsString(Payload payload) {
+        ByteBuffer data = payload.getData();
+        byte[] dst = new byte[data.remaining()];
+        data.get(dst);
+        return new String(dst);
```
stevegury: We'll add a charset in another commit ;-)

yschimke: Thanks Steve - I'm looking out for the EBCDIC users of the world

PR #105, opened 2016-06-26
Ignoring `ClosedChannelException` from the error stream.

NiteshKant:
#### Problem

Since `ReactiveSocket` reads and writes from a transport connection, it is completely expected that at some point the underlying connection may have been severed. In such a case, both the read and write will fail on the connection. This pollutes logs of applications that directly log the error stream (as expected).

#### Modification

Since, this is kind of an "expected" exception, there is no point to send this to the error stream.
This change adds a filter to the error stream to filter any known exceptions (as of today, it is only `ClosedChannelException`)

_PS: We currently do not unsubscribe from the input (and stop writing) when `ReactiveSocket` is explicitly closed. This change does not mean, we do not clean that up._


stevegury:
You beat me to it, I also had a fix for this!

PR #106, opened 2016-06-27
Local transport now uses tcp.

NiteshKant:
#### Problem

Local transport was using a custom local socket implementation.
Netty provides a `LocalChannel` abstraction that can be used.

#### Modification

Modified `reactivesocket-transport-local` to use `reactivesocket-transport-tcp`.
Refactored common test classes to `reactivesocket-test` module so they can be used in both tcp and local transport without code duplication.


reactivesocket-core/src/main/java/io/reactivesocket/util/PayloadImpl.java
```diff
@@ -0,0 +1,80 @@
...
+import io.reactivesocket.Frame;
+import io.reactivesocket.Payload;
+
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
+
+/**
+ * An implementation of {@link Payload}
+ */
+public class PayloadImpl implements Payload {
```
yschimke: This being available is very helpful

reactivesocket-core/src/main/java/io/reactivesocket/util/PayloadImpl.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private final ByteBuffer data;
+    private final ByteBuffer metadata;
+
+    public PayloadImpl(String data) {
```
yschimke: I'd suggest dropping the String constructors as you already have the static factory methods.  These 4 constructors make it seem like Payload carries Strings or byte[]+Charset

stevegury:
This looks good, but one of the test is failing.

reactivesocket-core/src/main/java/io/reactivesocket/util/PayloadImpl.java
NiteshKant: @yschimke are you referring to the `TestUtil` static methods? This class doesn't have static factory methods.

>> These 4 constructors make it seem like Payload carries Strings or byte[]+Charset

hmm.. not sure I agree. It is just different ways of constructing the payload with metadata being optional. Speaking of which, may be we should have a `NULL_PAYLOAD` constant.

NiteshKant:
@stevegury yes, there is some flakiness in the tests with local channel. I will get back to this tomorrow may be.

reactivesocket-core/src/main/java/io/reactivesocket/util/PayloadImpl.java
yschimke: My bad, I misread the private static methods for a single String.

NiteshKant:
Closing this in favor of #134 which picks all changes from here other than actually changing local transport.
I will start another PR just for local transport changes.

PR #107, opened 2016-06-27
Unsubscribe from the connection before `shutdown` the RS.

stevegury:
**Problem**
When we close a ReactiveSocket, it closes the underlying DuplexConnection
which trigger an unnecessary onError on the Requester/Responder.

**Solution**
Store the disposable of the DuplexConnection subscription as an instance
variable, and dispose it before closing the DuplexConnection.

**Modification**
I also added a `name` method in Requester/Responder to simplify debugging
code where we have client and server in the same JVM.
It now shows which Requester/Responder is generating an error, the client's
one or the server's one.

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/internal/Requester.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/internal/Requester.java
@@ -52,24 +52,23 @@ import org.agrona.collections.Int2ObjectHashMap;
  * Concrete implementations of {@link DuplexConnection} over TCP, WebSockets, Aeron, etc can be passed to this class for protocol handling.
  */
 public class Requester {
-
-    private final static Disposable CANCELLED = new EmptyDisposable();
-    private final static int KEEPALIVE_INTERVAL_MS = 1000;
+    private static final Disposable CANCELLED = new EmptyDisposable();
+    private static final int KEEPALIVE_INTERVAL_MS = 1000;
+    private static final long DEFAULT_BATCH = 1024;
+    private static final long REQUEST_THRESHOLD = 256;
 
     private final boolean isServer;
     private final DuplexConnection connection;
     private final Int2ObjectHashMap<UnicastSubject<Frame>> streamInputMap = new Int2ObjectHashMap<>();
     private final ConnectionSetupPayload setupPayload;
     private final Consumer<Throwable> errorStream;
-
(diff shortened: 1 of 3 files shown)


reactivesocket-core/src/main/java/io/reactivesocket/internal/Requester.java
```diff
@@ -115,8 +114,10 @@ public static Requester createServerRequester(
     }
 
     public void shutdown() {
-        // TODO do something here
-        System.err.println("**** Requester.shutdown => this should actually do something");
+        Disposable disposable = connectionSubscription.getAndSet(CANCELLED);
```
NiteshKant: nit: `EmptyDisposable` (`CANCELLED` here) ignores `dispose` call, so you don't need `&& disposable != CANCELLED` check here.

reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
```diff
@@ -348,22 +350,24 @@ public void success() {
...
-				onError(new RuntimeException("Connection Setup Failure", se));
+				onError(new RuntimeException(name() + ": Connection Setup Failure", se));
 			}
 
 			@Override
 			public void onError(Throwable t) {
 				// TODO validate with unit tests
 				if (childTerminated.compareAndSet(false, true)) {
-					errorStream.accept(t);
+					if (!(t instanceof ClosedChannelException)) {
```
NiteshKant: Is this required since we know have the `KnownErrorFilter`?

yschimke: nit: KnownErrorFilter is subtly different regarding subclasses of ClosedChannelException, but I expect you have observed ClosedChannelException in the wild only.

stevegury: @NiteshKant It's not required anymore.
@yschimke indeed, I'm not sure of any efficient way to achieve KnownErrorFilter purpose and properly dealing with subclasses without reflexion.

NiteshKant: @yschimke valid point. I guess we can iterate over known errors and check `instanceof` in `KnownErrorFilter`

PR #112, opened 2016-06-28
Loadbalancer: Refactor socket/factory Map into Lists + bugfix.

stevegury:
**Problem**
The loadbalancer uses internally 2 maps, one for the list of `ReactiveSocketFactory`
it can use for creating a new `ReactiveSocket`, and one for the active `ReactiveSocket`.
The link is made between the 2 maps through the `SocketAddress` of the associated
remote server. This is kind of clunky and requires explicit knowledge of
the `SocketAddress` (result type of `remote()`).

Also, when we select a new factory to connect to or when we select the slowest
ReactiveSocket to quick out, the sorting is made via the `sorted()` method
on the java stream. The comparison function is actually unstable because a
tcp socket can be closed while we do the sorting.

**Solution**
Use two lists, one for factories and one for ReactiveSocket. Selecting a factory
is made using the "Power of 2 Choices" technique. The WeightedSocket now contains
a reference to the associated factory, and closing it add back the factory to the
factory list.

**Modification**
Move the `WeightedSocket` class inside the `LoadBalancer`, no

I created a proper constructor to `LoadBalancer` with javadoc explaining all
parameters the algorithm requires.

The `LatencySubscriber` inside the LoadBalancer treats some exceptions
particularly:

- `TimeoutException` latency is used for predicating the next latency
- `TransportException` & `ClosedChannelException` removes the ReactiveSocket
   from the active list.

`Publishers.onError` now also `cancel` the subscription.

I created a StressTest file, which stresses most of the part of the Client:

- Adding/Removing new servers
- High request concurrency
- Dealing with failing/black-hole servers


yschimke:
Apologies for some basic questions, I'm getting up to speed on the code, but will ask anyway.

1) It doesn't look like you a replenishing both activeFactories, so eventually you will start using the bad factories anyway.  i.e. given 100 factories, 2 of which are unhealthy (low availability), ideally those 2 are statistically rarely used unless both unhealthy factories are chosen at the same time. 

2) Doesn't Power of 2 Choices rely on those initial 2 being chosen randomly? You seem to pick the first two from the Deque.

stevegury:
No worries about your basic questions.

1. The activeFactories is replenished when we close a ReactiveSocket or when a new server is coming from the discovery mechanism. ReactiveSocket beeing a multiplexed protocol we don't necessarily need tons of connection, and we don't need multiple connections to the same server. 
Furthermore, if 2 out of 100 connections are bad, this algorithm tends (still possible that they are in consecutive positions in the queue) to only use them when no other options are possible.

2. Yes, I should have said that the algorithm is based on "the power of 2 choices" but doesn't really respect it. Right now it is "aging" (picking in the front, adding back to the end) the factories via the Deque, so that we guarantee considering all factories.

reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
```diff
@@ -194,7 +194,7 @@ public void onNext(List<SocketAddress> socketAddresses) {
                         }
                     }
 
-                    List<ReactiveSocketFactory<SocketAddress>> factories =
+                    List<ReactiveSocketFactory<?>> factories =
```
NiteshKant: This isn't related, but do we need a type argument for `ReactiveSocketFactory`?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -70,25 +75,74 @@
...
+        this.lastApertureRefresh = Clock.now();
+        this.refreshPeriod = Clock.unit().convert(15L, TimeUnit.SECONDS);
         this.lastRefresh = Clock.now();
 
         factories.subscribe(factoryRefresher);
     }
 
+    public LoadBalancer(Publisher<List<ReactiveSocketFactory<?>>> factories) {
+        this(factories,
+            4.0,
```
NiteshKant: Should we have these as constants so that if someone has to change one value, he can use the remaining constants?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -121,18 +174,43 @@ public LoadBalancer(Publisher<List<ReactiveSocketFactory<SocketAddress>>> factor
...
-            .sorted((a, b) -> -Double.compare(a.availability(), b.availability()))
-            .limit(numberOfNewSocket)
-            .forEach(factory -> {
-                pendingSockets += 1;
-                factory.apply().subscribe(new SocketAdder(factory.remote()));
-            });
+        int n = Math.min(activeFactories.size(), numberOfNewSocket);
+        int maxEffort = 1 + activeFactories.size()/2;
+        int effort = 0;
+        while (n > 0 && effort < maxEffort) {
```
NiteshKant: (I remember we discussing it offline, so if this is in your TODOs, then let me know)

If the number of hosts < aperture size, should we favor creating more connections on the same host?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -618,4 +661,284 @@ public void shutdown() {}
...
+            private final WeightedSocket socket;
+
+            CountingSubscriber(Subscriber<U> child, WeightedSocket socket) {
+                this.child = child;
+                this.socket = socket;
+            }
+
+            @Override
+            public void onSubscribe(Subscription s) {
+                pendingStreams.incrementAndGet();
```
NiteshKant: It will be better, if you access `WeightedSocket` member variables using `socket.*` instead of direct access. It is easier to understand.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -618,4 +661,284 @@ public void shutdown() {}
...
+            @Override
+            public void onNext(U u) {
+                child.onNext(u);
+            }
+
+            @Override
+            public void onError(Throwable t) {
+                pendingStreams.decrementAndGet();
+                child.onError(t);
+                removeSocket(socket);
```
NiteshKant: `RemoveItselfSubscriber` was removing itself on `TransportException`, but this removes the socket on any exception. Intentional?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -618,4 +661,284 @@ public void shutdown() {}
...
+        @Override
+        public Publisher<Payload> requestSubscription(Payload payload) {
+            return subscriber ->
+                child.requestSubscription(payload).subscribe(new CountingSubscriber<>(subscriber, this));
+        }
+
+        @Override
+        public Publisher<Void> fireAndForget(Payload payload) {
+            return subscriber ->
+                child.fireAndForget(payload).subscribe(new CountingSubscriber<>(subscriber, this));
```
NiteshKant: This may use `LatencySubscriber`?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -618,4 +661,284 @@ public void shutdown() {}
...
+        }
+
+        @Override
+        public Publisher<Void> metadataPush(Payload payload) {
+            return subscriber ->
+                child.metadataPush(payload).subscribe(new CountingSubscriber<>(subscriber, this));
```
NiteshKant: This may use `LatencySubscriber`?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -211,40 +289,47 @@ private synchronized void quickSlowestRS() {
...
+                slowest = socket;
+                if (load == 0.0) {
+                    break;
                 }
-            });
+            }
+        }
+
+        if (slowest != null) {
+            removeSocket(slowest);
```
NiteshKant: So, it looks like the socket is removed in the following cases:

- Removed from discovery.
- Receives error from any request.
- Slowest amongst others.

I do not see that we are hooking into the `close()` of the socket. 
Since, we do not listen to keep-alive failures here, we will most likely not know if a connection is closed and is slow for a long time, till we attempt to use it. Appears to me that listening to socket close will be a good thing here.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -121,18 +174,43 @@ public LoadBalancer(Publisher<List<ReactiveSocketFactory<SocketAddress>>> factor
...
+            }
+
+            if (factory0.availability() == 0.0 && factory1.availability() == 0.0) {
+                activeFactories.addLast(factory0);
+                activeFactories.addLast(factory1);
+                effort++;
+            } else if (factory0.availability() < factory1.availability()) {
+                n--;
+                pendingSockets++;
+                activeFactories.addLast(factory0);
```
NiteshKant: I do not quiet understand the reason for aging the relatively slower factory. The slower factory here may be the fastest amongst others, so why to penalize?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -618,4 +661,284 @@ public void shutdown() {}
...
+    private class WeightedSocket extends ReactiveSocketProxy {
+        private static final double STARTUP_PENALTY = Long.MAX_VALUE >> 12;
+
+        private final ReactiveSocket child;
+        private ReactiveSocketFactory<?> factory;
+        private final Quantile lowerQuantile;
+        private final Quantile higherQuantile;
+        private final long inactivityFactor;
+
+        private volatile int pending;       // instantaneous rate
```
NiteshKant: Since all access is `synchronized`, may be we can make the `getter` also `synchronized` and remove this `volatile`?

reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
stevegury: The type argument represents the identifier of a server, which is, in our implementation, always `SocketAddress`.
I can refactor the code to remove it (later).

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: Good point!

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: In practice, I don't think it's necessary.
If `aperture` is bigger than the number of hosts, it will just mean that we'll have (on average) more than 2 outstanding messages.

The only case where this could be problematic is when you have a very small number of clients/servers, and because of how Netty assign connections to Threads, if a 32 cores server only has 2 connections, it will only use 1/16th of its CPU.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: If a socket is closed its `availability` instantaneously becomes `0`, so it will never be selected. 
It will be removed on the next refresh.

But you're right that we should remove it on socket close.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: `getPending()` isn't synchronized.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: There's no response to measure the latency.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: Ditto, no response.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: This is a bug, actually, that part of the code (every interaction-model other than request/response) isn't tested.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: Basically, what I'm trying to achieve here is selecting a good factory without having to scan the entire list.
If I don't care about having a "good" factory, I can just select the top of the queue, but, by looking at the first 2 elements it gives me almost enough information for making a smart decision (this isn't exactly what the power of two choices paper proved but the idea is similar).

The reason why I'm "aging" (adding to the end) the unselected factory, is to avoid having a bad instance in the front, which will always lead to selecting the second one.

The other option here would be to use literally the Power of two choices algorithm (pick two factories randomly, not only the first two), which is something I'm ok to consider. It would have roughly the same characteristic. 

If you think it's simpler to understand, I'll make the change.

yschimke: It's hard to reason about this load balancer because without the random element it seems affected by its environment.  

e.g. it seems like FactoriesRefresher will add new elements to the tail of the list, so something like a cluster roll will likely result in waves of traffic to the same shards at roughly the same time.

stevegury: That's a good point, I'll switch to pure "Power of two choice" implementation.

stevegury:
Here's a summary of the last commit:

#### Address comments:
- Add static named default initial values
- Add type parameter to loadBalancer (matching ReactiveSocketFactory)
- Move from a queue implementation to a list
- Refactor the queue/aging into proper "Power of Two Choices"

#### Fix Bug:
- Computing the server list delta (added/removed) was wrong.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
NiteshKant: This will include write latency to the socket, which _may_ be useful as we do not currently listen to `keep-alive` latency? Probably more if this is the primary interaction model for a client.

I think you said that the only _really_ supported model is request-response, so this may just be moot.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
NiteshKant: Yeah I was saying, that may be we can also synchronize the getter.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: Yes, so far, only request-response interaction are used for making a load-balancing decision.
I'll improve this in the future.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: I'm not focusing on performance here, but that's a micro-optimization that barely impact the readability of the code.
So I would like to keep it.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -121,18 +179,59 @@ public LoadBalancer(Publisher<List<ReactiveSocketFactory<SocketAddress>>> factor
...
-            .sorted((a, b) -> -Double.compare(a.availability(), b.availability()))
-            .limit(numberOfNewSocket)
-            .forEach(factory -> {
-                pendingSockets += 1;
-                factory.apply().subscribe(new SocketAdder(factory.remote()));
-            });
+        int n = numberOfNewSocket;
+        if (n > activeFactories.size()) {
+            n = activeFactories.size();
+            logger.info("addSockets(" + numberOfNewSocket
```
NiteshKant: nit: May be we can use templated messages here?

```java
logger.info("addSockets({}) restricted", numberOfNewSocket)
```

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -121,18 +179,59 @@ public LoadBalancer(Publisher<List<ReactiveSocketFactory<SocketAddress>>> factor
...
+                break;
+            }
+            ReactiveSocketFactory<T> factory0 = null;
+            ReactiveSocketFactory<T> factory1 = null;
+            int i0 = 0;
+            int i1 = 0;
+            for (int i = 0; i < EFFORT; i++) {
+                i0 = rng.nextInt(size);
+                i1 = rng.nextInt(size - 1);
+                if (i1 >= i0) {
```
NiteshKant: Can you explain why this is `>=` and not `==`?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -121,18 +179,59 @@ public LoadBalancer(Publisher<List<ReactiveSocketFactory<SocketAddress>>> factor
...
+                    i1++;
+                }
+                factory0 = activeFactories.get(i0);
+                factory1 = activeFactories.get(i1);
+                if (factory0.availability() > 0.0 && factory1.availability() > 0.0)
```
NiteshKant: Should we choose one if the other has availability <= `0.0`?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -121,18 +179,59 @@ public LoadBalancer(Publisher<List<ReactiveSocketFactory<SocketAddress>>> factor
...
+                    break;
+            }
+
+            if (factory0.availability() < factory1.availability()) {
+                n--;
+                pendingSockets++;
+                if (i1 < size - 1) {
```
NiteshKant: I think it will be good to put a comment here as to why we are doing this as opposed to doing `.remove(i1)`.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -151,22 +250,29 @@ private synchronized void refreshAperture() {
 
         long now = Clock.now();
         boolean underRateLimit = now - lastApertureRefresh > APERTURE_REFRESH_PERIOD;
```
NiteshKant: This is named `underRateLimit` because this refresh is not to be rate limited because of the refresh period?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -151,22 +250,29 @@ private synchronized void refreshAperture() {
...
-        int previous = targetAperture;
         if (avgPending < 1.0 && underRateLimit) {
```
NiteshKant: Is the idea here to refresh `pendings` EWMA even if we don't refresh the aperture?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -151,22 +250,29 @@ private synchronized void refreshAperture() {
...
+        targetAperture = newValue;
+        targetAperture = Math.max(minAperture, targetAperture);
+        int maxAperture = Math.min(this.maxAperture, activeSockets.size() + activeFactories.size());
         targetAperture = Math.min(maxAperture, targetAperture);
+        lastApertureRefresh = now;
+        pendings.reset((minPendings + maxPendings)/2);
 
         if (targetAperture != previous) {
-            logger.info("Current pending=" + avgPending
+            logger.info("Current pending=" + pendings.value()
```
NiteshKant: Should these logs be on debug level?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -211,40 +317,47 @@ private synchronized void quickSlowestRS() {
...
+        double lowestAvailability = Double.MAX_VALUE;
+        for (WeightedSocket socket: activeSockets) {
+            double load = socket.availability();
+            if (socket.getPredictedLatency() != 0) {
+                load *= 1.0 / socket.getPredictedLatency();
+            }
+            if (load < lowestAvailability) {
+                lowestAvailability = load;
+                slowest = socket;
+                if (load == 0.0) {
```
NiteshKant: If the availability is `0.0` then all the above calculations are thrown away, so looks like this can be the first condition in this loop? Should we also handle negative availability?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -211,40 +317,47 @@ private synchronized void quickSlowestRS() {
...
+        }
+
+        if (slowest != null) {
+            removeSocket(slowest);
+        }
+    }
+
+    private synchronized void removeSocket(WeightedSocket socket) {
+        try {
+            logger.info("Removing socket: -> " + socket);
```
NiteshKant: `DEBUG` level?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -211,40 +317,47 @@ private synchronized void quickSlowestRS() {
...
+            activeSockets.remove(socket);
+            activeFactories.add(socket.getFactory());
+            socket.close();
+        } catch (Exception e) {
+            logger.warn("Exception while closing a ReactiveSocket", e);
```
NiteshKant: Not sure if this matters, so should we reduce the log level?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -311,8 +423,8 @@ private synchronized ReactiveSocket select() {
             if (i2 >= i1) {
                 i2++;
             }
-            rsc1 = buffer.get(i1);
-            rsc2 = buffer.get(i2);
+            rsc1 = activeSockets.get(i1);
+            rsc2 = activeSockets.get(i2);
             if (rsc1.availability() > 0.0 && rsc2.availability() > 0.0)
```
NiteshKant: Same comment as the previous P2C code, should we select one with non `0.0` availability if the other is `0.0`

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -426,37 +502,58 @@ public void onSubscribe(Subscription subscription) {
         }
 
         @Override
-        public void onNext(List<ReactiveSocketFactory<SocketAddress>> newFactories) {
-            List<ReactiveSocketFactory<SocketAddress>> removed = computeRemoved(newFactories);
+        public void onNext(Collection<ReactiveSocketFactory<T>> newFactories) {
             synchronized (LoadBalancer.this) {
```
NiteshKant: I know you haven't optimized for concurrency, but this seems excessive as for every refresh, this would mean we would not be serving any requests, is my understanding correct?
How would you think it will behave with clusters of thousand servers and a discovery client which isn't great in pushing deltas?

NiteshKant: This also means that the `select()` will be blocking the eventloop for the duration of the refresh, correct?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -618,4 +691,286 @@ public void shutdown() {}
...
+
+        int getPending() {
+            return pending;
+        }
+
+        private synchronized long instantaneous(long now) {
+            return duration + (now - stamp0) * pending;
+        }
+
+        private synchronized long incr() {
```
NiteshKant: If `incr()` and `decr()` are only used by the inner class `LatencySubscriber`, should we move them there?

NiteshKant:
@stevegury I have done one pass of the changes and apart from the `synchronized` block during refresh, nothing is major. Once you address/answer these comments, I think we will be ready to merge.
Apologies for the delay in reviewing.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: I want to find two different random numbers here in [0, n].
I pick the first one randomly, let's call it `i0`.
`[.....i0.....]`
I want to pick another one `i1` which has to be different from x, so I pick a random number in [0, n-1]
`[.....i0.....]`
`[...........]`

Two cases here:
either `i1` is < `i0` and that's fine
`[.....i0.....]`
`[.i1..i0.....]`

or `i1` is >= `i0` then I need to increment `i1` because of the spot that `i0` is already taking.
`[.....i0.....]`
`[.....i0..i1.]`

If you don't do that, and only check for the case where you pick the same number twice, you bias the second random number to `i0 + 1` with a 2/(n-1) probability. (There's two cases where `i1 = i0+1`, when you secondly pick `i0` and `i0+1`)

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: I decided not to, because it will complexify the code.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: Yes, it limits the number of aperture refresh, to 1 every 15 seconds to be resilient to small spikes.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: yeah, `pendings` is the average (ewma) number of outstanding messages, it is updated all the time.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: They could, in practice it's not that chatty in steady-state.
I'll update the code.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: By undefined (other than in my head) contract we should never have negative availability.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: This should be exceptional, I'd like to know if we experience those exception.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: Same answer.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: Correct.
I'll plan to optimize that in the future (but it didn't show up as a problem on my "limited" stress test)

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: It makes more sense (I think) to have them here, they are modifying members of the parent class.

NiteshKant:
@stevegury is the failed test flaky or `2000` millis is too aggressive a timeout?

PR #113, opened 2016-06-29
use charset consistently

yschimke:
I don't like this PR, so using it for discussion.

The spec says

> Encoding MIME Type: MIME Type for encoding of Data and Metadata. This MAY be a US-ASCII string that includes the Internet media type specified in RFC 2045. Many are registered with IANA such as CBOR. Suffix rules MAY be used for handling layout. For example, application/x.netflix+cbor or application/x.reactivesocket+cbor or application/x.netflix+json. The string may or may not be null terminated.

Does this actually mean:  "This MAY be a Internet media type specified in RFC 2045.  If present, the value MUST be US-ASCII.".  If it can be UTF-8 then string.length() may not be the byte length.

Also can we change the spec to say that the string MUST NOT be null terminated.  we need to string before creating strings.  I assume this is useful if we want to have 0-copy strings in C++.  But we create these garbage strings in Java anyway...

I'll add tests once this is agreed.

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/exceptions/Exceptions.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/exceptions/Exceptions.java
@@ -30,9 +30,15 @@ public class Exceptions {
         final int errorCode = Frame.Error.errorCode(frame);
         String message = "<empty message>";
         final ByteBuffer byteBuffer = frame.getMetadata();
-        if (byteBuffer.hasArray()) {
-            message = new String(byteBuffer.array(), UTF_8);
+
+        // TODO nasty, but it seems strange to assume ByteBuffer is array backed
+        byte[] bytes = new byte[byteBuffer.remaining()];
+        byteBuffer.get(bytes, 0, bytes.length);
+        int length = bytes.length;
+        if (length > 1 && bytes[length - 1] == 0) {
+            length--;
         }
+        message = new String(bytes, 0, length, UTF_8);
 
         Throwable ex;
         switch (errorCode) {
--- a/reactivesocket-core/src/main/java/io/reactivesocket/exceptions/Exceptions.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/exceptions/Exceptions.java
@@ -29,7 +29,7 @@ public class Exceptions {
     public static Throwable from(Frame frame) {
         final int errorCode = Frame.Error.errorCode(frame);
         ByteBuffer dataBuffer = frame.getData();
-        String message = dataBuffer.remaining() == 0 ? "<empty message>" : ByteBufferUtil.toUtfString(dataBuffer);
+        String message = dataBuffer.remaining() == 0 ? "<empty message>" : ByteBufferUtil.toUtf8String(dataBuffer);
 
         Throwable ex;
         switch (errorCode) {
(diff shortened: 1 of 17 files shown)


reactivesocket-core/src/main/java/io/reactivesocket/exceptions/Exceptions.java
```diff
@@ -30,9 +30,15 @@ public static Throwable from(Frame frame) {
         final int errorCode = Frame.Error.errorCode(frame);
         String message = "<empty message>";
         final ByteBuffer byteBuffer = frame.getMetadata();
-        if (byteBuffer.hasArray()) {
```
yschimke: hasArray here seems problematic in two ways.

1) assuming array backed
2) assuming that the byte buffer is not a sliced with a shared larger backing array

package io.reactivesocket.exceptions;

```
import java.nio.ByteBuffer;

public class AAAA {
        ByteBuffer b = ByteBuffer.allocate(100);
        b.position(20).limit(50);
        ByteBuffer b2 = b.slice();

        System.out.println(b.capacity()); // 100
        System.out.println(b.remaining()); // 30
        System.out.println(b.array()); // [B@d716361
        System.out.println(b.array().length); // 100

        System.out.println();

        System.out.println(b2.capacity()); // 30
        System.out.println(b2.remaining()); // 30
        System.out.println(b2.array()); // [B@d716361
        System.out.println(b2.array().length); // 100
    }
}
```

NiteshKant:
@yschimke I am not quiet sure if we want to use US-ASCII over UTF-8, do you want to open an issue for the protocol?

Since, mime-length is part of the frame, I am sure we don't intend it to be null terminated. It would help to make it clear in the spec.

reactivesocket-core/src/main/java/io/reactivesocket/exceptions/Exceptions.java
NiteshKant: Agreed, it is wrong to assume the buffer is array backed.

NiteshKant: This code is defaulting to `<empty message>` if the buffer is not array backed, which isn't correct but is safe, rite?

Moreover, it is assuming `UTF-8` which isn't correct at all and it is converting entire metadata frame into the message. So, I guess ignoring error message if the buffer isn't array backed, isn't the bigger problem here 😄

yschimke:
@NiteshKant I agree about using UTF-8, I'll fix the bug with string.length and remove the null termination handling.

yschimke:
@NiteshKant anything blocking this merging? AFAIK it only fixes bug, no breaking behaviour.

NiteshKant:
@yschimke just time, been busy today. I should be able to get to this today.

reactivesocket-core/src/main/java/io/reactivesocket/internal/frame/ByteBufferUtil.java
```diff
@@ -44,4 +45,10 @@ public static ByteBuffer preservingSlice(final ByteBuffer byteBuffer, final int
         byteBuffer.limit(savedLimit).position(savedPosition);
         return result;
     }
+
+    public static String toUtfString(ByteBuffer byteBuffer) {
```
NiteshKant: nit: `toUtf8String`?

yschimke: fixed.  thanks

stevegury:
👍 LGTM

PR #114, opened 2016-06-29
Tabs => Spaces

NiteshKant:
#### Problem

I don't agree with Richard Hendricks to use tabs over spaces.

#### Solution

Gimme back my spaces.

#### Result

Awesomeness .....

PS: I don't use vim or emacs... Idea rocks!

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/ConnectionSetupPayload.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/ConnectionSetupPayload.java
@@ -1,17 +1,14 @@
-/**
- * Copyright 2015 Netflix, Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


tmontgomery:
For what it is worth, I am spaces >> tabs.

stevegury:
Uniformity is more important than one choice or the other.

yschimke:
+1 to uniformity.  I think it would be helpful if you bulk formatted the code according to a default Intellij settings as well.  Inconsistency of braces have been a grating on me :)

NiteshKant:
@yschimke I thought to do that but may be we should agree on which formatting style to use first 😄

yschimke:
Yep.  FWIW I find this sort of setup really easy to use https://github.com/square/java-code-styles/tree/master/configs

I'm not religious about styles, just want to be able to reformat code without thinking too much.  e.g. no standards for aligning javadoc params manually etc.

stevegury:
If we all use IntelliJ, I propose to use the default Java code style.
AFAIK, it matches Sun/Oracle's recommended Java code style.

In `Editor > Code Style > Java > Scheme` choose `Default`

NiteshKant:
I am fine with that so may be when the pending PRs are merged we can bulk reformat the code and remove warnings which are so annoying when reading the code.

tmontgomery:
Uniformity is very good.

I have found that [Kevlin Henney's points](https://vimeo.com/74316116) make a lot of sense and are good guidelines for format, naming, etc. Also, good reasoning behind them.

PR #115, opened 2016-06-29
Completing subscriber after sending error frame (caused memory leak).

NiteshKant:
#### Problem

In some cases, `Responder` was converting an `onError()` to `onNext(Frame.Error)` but was not sending an `onComplete()` after the `onNext()`.
This causes the `subscriber` to wait for a terminal event, which never arrives, or if it arrives, does not find the `streamId` to be valid since the code was invoking `cleanUp()`.
This although subtle, but causes memory build up as the `Subscriber` in the case of a server, is the subscriber which is writing the response. If the response does not complete, all state associated with the request is kept alive forever.

#### Solution

Call `onComplete()` after sending an error frame.

This code does not call `cancel()` after `onComplete()` as reactive streams spec does not mandate cancellation post a terminal event, unlike `RxJava`.

#### Result

No more memory leak!

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
@@ -75,10 +75,10 @@ public class Responder {
         this.isServer = isServer;
         this.connection = connection;
         this.connectionHandler = connectionHandler;
-        clientRequestHandler = requestHandler;
+        this.clientRequestHandler = requestHandler;
         this.leaseGovernor = leaseGovernor;
         this.errorStream = errorStream;
-        timeOfLastKeepalive = System.nanoTime();
+        this.timeOfLastKeepalive = System.nanoTime();
         this.setupCallback = setupCallback;
     }
 
--- a/reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
@@ -10,6 +10,7 @@
  *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
  *  specific language governing permissions and limitations under the License.
  */
+
 package io.reactivesocket.internal;
 
 import io.reactivesocket.ConnectionSetupHandler;
@@ -36,7 +37,6 @@ import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
 
-import java.nio.channels.ClosedChannelException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiFunction;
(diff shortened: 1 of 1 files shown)


reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
```diff
@@ -483,16 +485,16 @@ private void cleanup() {
                 }
 
             };
-            synchronized(Responder.this) {
+            synchronized(this) {
```
stevegury: `Responder.this` is correct here, `this` represents the `new Subscriber` line 417.

reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
```diff
@@ -636,7 +638,7 @@ private void cleanup() {
...
```
stevegury: ditto

reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
```diff
@@ -822,7 +824,7 @@ private void cleanup() {
                     }
 
                 };
-                synchronized(Responder.this) {
+                synchronized(this) {
```
stevegury: ditto

reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
```diff
@@ -75,10 +75,10 @@ private Responder(
         this.isServer = isServer;
         this.connection = connection;
         this.connectionHandler = connectionHandler;
-        this.clientRequestHandler = requestHandler;
+        clientRequestHandler = requestHandler;
```
stevegury: I know that `this.` is not required, but I like the uniformity when we use it all the time.

reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
NiteshKant: Actually not, `new Subscriber` ends at line 464 and `Publisher` at 487.
This `synchronized` block belongs to `handleRequestResponse` function.

This and other changes were suggested by Intellij for me!

reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
NiteshKant: Hmmm ... ok, I would leave it status quo for now.

reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
NiteshKant: Ditto 😄

reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
NiteshKant: Ditto 😄

reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
stevegury: My bad.

PR #116, opened 2016-06-29
Add encrypted keys for Sonatype/Bintray

stevegury:
***Problem***
The Maven Central synchronization is not working (Travis CI is building the jars, but can't copy them to maven central).

***Solution***
This is the result of removing the previously encrypted keys, and running this script (authenticated as ReactiveSocketAdmin).

```
travis encrypt bintrayUser=*** --add
travis encrypt bintrayKey=*** --add
travis encrypt sonatypeUsername=*** --add
travis encrypt sonatypePassword=*** --add
```


NiteshKant:
Yay!

PR #117, opened 2016-06-30
Statically create `TimeoutException`.

NiteshKant:
#### Problem

If a host is latent and most of the requests happen to timeout, creating a new `TimeoutException` for every timeout produces garbage and wastes CPU fetching the stack, which is useless.

#### Solution

Create a static instance of `TimeoutException` and fill it with empty stack.

#### Result

Less garbage, less work, more happiness.


stevegury:
What about moving the `TimeoutException` from `io.reactivesocket.client.exception` (in the `reactivesocket-client` project) to `io.reactivesocket.exceptions` (in the `reactivesocket-core` project) and use this one.
(still generate a static instance).

stevegury:
The advantage is that when you see a `reactivesocket TimeoutException` you're sure of the source of the Exception, while you're not really sure of the source of the `java TimeoutException`. You can then take particular action based on this knowledge of the source.

c.f. `io.netty.handler.timeout.TimeoutException` or `com.twitter.finagle.TimeoutException`.

NiteshKant:
@stevegury I think that argument is more valid when you are writing a generic catch code where you can get exceptions from various different sources. In this case, since the `TimeoutException` is a result of subscribing to a `Publisher`, I don't see a reason why `TimeoutException` from a java library must be considered different from the `TimeoutException` from ReactiveSocket. In either cases, the timeout is the result of the `Publisher`.

Anyways, I think this isn't an important point for this change, so I have modified code to use ReactiveSocket's `TimeoutException` (after moving it to `reactivesocket-core`). More discussions are reserved for another day 😉

PR #118, opened 2016-07-01
ClientBuilder: Disable default timeout, remove retries.

stevegury:
***Problem***
There's presently no way to disable timeouts/retries other than manually
composing the stack.

***Solution***
Default to disable any timeout in the ClientBuilder, remove completely retries.

***Modification***
Also remove unused fields `backupQuantile`, `counter` and `name`.

### DIFF
--- a/reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
+++ b/reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
@@ -34,9 +34,6 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 public class ClientBuilder<T> {
-    private static AtomicInteger counter = new AtomicInteger(0);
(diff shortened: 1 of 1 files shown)


reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
```diff
@@ -151,13 +121,19 @@ public ReactiveSocket build() {
         }
 
         ReactiveSocketConnector<T> filterConnector = connector
-            .chain(socket -> new TimeoutSocket(socket, requestTimeout, requestTimeoutUnit, executor))
+            .chain(socket -> {
+                if (requestTimeout > 0) {
```
NiteshKant: Does this mean that the `ClientBuilder` instance will always be around? May be a better idea to conditionally add `chain`?

PR #121, opened 2016-07-01
use remaining instead of capacity generally

yschimke:
Replace use of capacity which is only appropriate if you know exactly what the client has given you (a whole ByteBuffer representing the relevant data).  Uses remaining instead which assumes you trust your caller has given you a buffer that won't be change during the method call, but uses position and limit to support sliced buffers.


yschimke:
Some of these look like definite bugs (FrameHeaderFlyweight), and others seem really dependent on the Contextual use of ByteBuffers.

stevegury:
👍 
Just FYI: We try to always write a small description in the PR, rather than adding comments.

NiteshKant:
Tell me about this AS I hit on this bug in our internal code yesterday! 👍  for this @yschimke !

PR #122, opened 2016-07-01
Add meaningfull toString() methods to most objects.

stevegury:
***Problem***
Debuging and Logging could be improved by adding meaningful `toString()`
methods (that expose internal state).

***Solution***
Add meaningful `toString()` methods.

***Modification***
- Remove unused chaining functions.
- Change visibility (to protected) of the `child` factory in ReactiveSocketFactory.

### DIFF
--- a/reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
+++ b/reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
@@ -219,5 +219,15 @@ public class ClientBuilder<T> {
             null
         );
     }
+
+    @Override
+    public String toString() {
+        return "ClientBuilder("
+            + "source=" + source.toString()
+            + ", connector=" + connector.toString()
+            + ", requestTimeout=" + requestTimeout + " " + requestTimeoutUnit
+            + ", connectTimeout=" + connectTimeout + " " + connectTimeoutUnit
+            + ")";
+    }
 }
 
--- a/reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
+++ b/reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
@@ -223,11 +223,10 @@ public class ClientBuilder<T> {
     @Override
     public String toString() {
         return "ClientBuilder("
-            + "source=" + source.toString()
-            + ", connector=" + connector.toString()
+            + "source=" + source
+            + ", connector=" + connector
             + ", requestTimeout=" + requestTimeout + " " + requestTimeoutUnit
             + ", connectTimeout=" + connectTimeout + " " + connectTimeoutUnit
             + ")";
     }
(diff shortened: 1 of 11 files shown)


reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
```diff
@@ -219,5 +219,15 @@ public void onNext(Collection<T> socketAddresses) {
             null
         );
     }
+
+    @Override
+    public String toString() {
+        return "ClientBuilder("
+            + "source=" + source.toString()
```
yschimke: nit: toString() is redundant, assume the same since unlikely to be null

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -867,6 +867,7 @@ public String toString() {
                 + " quantile-high:" + higherQuantile.estimation()
                 + " inter-arrival:" + interArrivalTime.value()
                 + " duration/pending:" + (pending == 0 ? 0 : (double)duration / pending)
+                + " pending:" + pending
```
yschimke: I would *NOT* suggest commans-lang ToStringBuilder, but it would be nice to be consistent.

ClientBuilder(a=1, b=2)
vs
WeightedSocket@123434234 [median: a quantile: b]

reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
NiteshKant: Even if it is null, string concatenation will not throw `NullPointerException`. Try:

```java
System.out.println("hello" + null);
```

yschimke: I know that is safe, but the current null.toString() in the diff is unsafe.  There is a bizarro parallel universe, where Steve intended a NPE if you call toString with nulls.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
NiteshKant: Default `toString()` generator for Intellij? Although it uses `StringBuilder` I think. So, it becomes 2 steps, generate `toString()` and then accept intention of using string concatenation 😄

stevegury:
I fix all consistency issues, I'm now using the convention:
`ClassName(state)->` or `ClassName(name=a, name2=b)->` (depending on the number of variables).

I don't like using StringBuilder since the JVM is already doing the optimization.

I removed the explicit call to `toString()`.

NiteshKant:
@stevegury looks like your other PRs conflicted with this PR, you need to rebase.

stevegury:
I just merged master.

PR #126, opened 2016-07-04
Filter non-up instances in Eureka

NiteshKant:
#### Problem

Eureka source does not currently filter messages if `status != UP`. This means that on graceful shutdown of servers, `LoadBalancer` will not sending requests to the server, till it is shutdown.

#### Modification

Filter all messages that do not have status as `UP`

#### Result

More graceful removal of shutting down instances.


NiteshKant:
\cc @stevegury @yschimke

reactivesocket-discovery-eureka/src/main/java/io/reactivesocket/discovery/eureka/Eureka.java
```diff
@@ -51,6 +52,9 @@ public void subscribe(Subscriber<? super Collection<SocketAddress>> subscriber)
             private synchronized void pushChanges(Subscriber<? super List<SocketAddress>> subscriber) {
                 List<InstanceInfo> infos = client.getInstancesByVipAddress(vip, secure);
                 List<SocketAddress> socketAddresses = infos.stream()
+                    .filter(instanceInfo -> {
+                        return instanceInfo.getStatus() == InstanceStatus.UP;
```
yschimke: nittiest of nits: 

Can be simplified down to 1 line

`.filter(i -> i.getStatus() == InstanceStatus.UP)`

yschimke:
Change looks sensible to me, but first I've heard of Eureka to be honest. 

If the timeout of 60 seconds for the test suggests they take 20-30 seconds, then this could get annoying quick.  Otherwise looks good.

NiteshKant:
@yschimke Thanks for the review!

The `60` seconds timeout is just my template, I will modify this to put a smaller timeout, 10 seconds. The test doesn't have any wait, so shouldn't take any time.

NiteshKant:
Ended up removing the timeout for the test.

PR #127, opened 2016-07-06
Publishers cleanup and new functions

NiteshKant:
#### Problem

`Publishers.retry()` had the following bugs:

- Retry subscription was sending duplicate `onSubscribe` callbacks to the downstream subscribers. This is illegal according to spec.
- Flow control was not properly managed if the error happens after a few emissions.

Since we have removed retry functionality from the `ClientBuilder` there is no reason to maintain this code.

Apart from the above, `Publishers` was not wiring the downstream subscriber cancellation to the cancellation of the `Subscriber`. This means that a cancel followed by `onNext` would not have stopped the emission, which although not necessary (because of the inherent race between cancel and emissions) but is good to have.

#### Modification

 - Removed `RetrySocket` and `Publishers.retry()` since they are not used.
 - Added (moved from `PublisherUtils`) `.empty()`, `.just()` and `.error()` to `Publishers`
 - Added `concatEmpty()` to `Publishers`. This is required in places where we concat `Publisher<Void>` in custom crafter code. (I will send another PR which is to use this operator in `ReactiveSocket.close()`
 - Added tests for functions in `Publishers`.


NiteshKant:
@stevegury @yschimke I am merging this as I need these changes for another fix.

Feel free to review and comment and I will send another PR to address the comments. There isn't anything controversial in this PR I believe.

reactivesocket-core/src/main/java/io/reactivesocket/internal/SafeCancellableSubscriber.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public void onNext(T t) {
+        if (!done.get()) {
+            doOnNext(t);
+        }
+    }
+
+    @Override
+    public void onError(Throwable t) {
```
yschimke: nit: for this sort of class making the AtomicBooleans private and methods that access them final seems prudent.

NiteshKant: That is a good point. I changed this error prone approach to remove inheritance and use lambdas for callbacks. PR #137 does this cleanup, thanks!

reactivesocket-core/src/main/java/io/reactivesocket/internal/SafeCancellableSubscriber.java
```diff
@@ -0,0 +1,74 @@
...
+    protected final AtomicBoolean subscribed = new AtomicBoolean();
+    protected final AtomicBoolean done = new AtomicBoolean();
+
+    @Override
+    public void onSubscribe(Subscription s) {
+        if (subscribed.compareAndSet(false, true)) {
+            super.onSubscribe(s);
+        } else {
+            onError(new IllegalStateException("Duplicate subscription."));
+        }
```
ktoso: this is violates the reactive streams spec Rule 2.5: 

> A `Subscriber` MUST call `Subscription.cancel()` on the given `Subscription` after an `onSubscribe` signal if it already has an active `Subscription`.

https://github.com/reactive-streams/reactive-streams-jvm#2.5

Please use the TCK each time you implement a custom Subscriber / Publisher, as there's plenty cases like this which are well defined in the spec but easy to forget if one does not test for them. The TCK covers the vast majority of such cases: https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck

NiteshKant: Thanks @ktoso for the review!

Yes, one of my intentions behind this refactor was to centralize the creation of `Subscriber` and `Publisher` in the project so that we can easily verify the behaviors. Currently we create lambdas all over the code base which are hard to test and have bugs related to violating the spec (eg: #136 ). I intend to add tck validation tests for these subscribers and publishers to easily validate these conditions.

ktoso: Very glad to hear that 👍 
I'll keep my eye on the repo, hopefully being able to provide some useful feedback :)

PR #133, opened 2016-07-07
Fixes #132 (Cancel before `RequestN`)

### LINKED ISSUE #132: Calling cancel() before request(n) causes null pointer exception
While writing the TCK for reactive sockets, my TCK code found the bug mentioned above.

```
val s4 = requestResponse("g", "h")
    s4 cancel()
    s4 assertCanceled()
    s4 assertNoErrors()
    s4 assertNotCompleted()
    s4 assertReceivedCount 0
```

The line `s4 cancel()` causes the NPE on the client side, as it seems like the current implementation always expects request(n) to be called before cancel.

NiteshKant:
#### Problem

As described in #132, All `Subscription` implementations in`Requester` were not handling the condition where a `cancel` is issued before `requestN`.

#### Modification

Correctly handle this condition and also short-circuit for multiple cancellations and `requestN()` post `cancel()`.
Added Unit tests to verify this for all interaction models.

#### Result

No `NullPointerException`

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/internal/Requester.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/internal/Requester.java
@@ -306,6 +306,7 @@ public class Requester {
         return (Subscriber<? super Payload> child) -> {
             child.onSubscribe(new Subscription() {
 
+                private boolean cancelled;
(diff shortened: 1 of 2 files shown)


stevegury:
👍

PR #134, opened 2016-07-07
Refactor of TCP transport.

NiteshKant:
#### Problem

This is a split of PR #106 to make local transport use TCP.
There seems to be some issue in netty's local transport which creates race-conditions and causes test failures.
These refactor has some changes that are not necessarily related to local transport but makes the usage better.

#### Modification

Refactored common test classes to `reactivesocket-test` module so they can be used in both tcp and local transport without code duplication.
Added `PayloadImpl` as a utility implementation of `Payload`.


NiteshKant:
Since this is a subset of PR #106 which got a 👍  I am just merging this now.

yschimke:
Are you sure the issues for Local Transport were not because you were reusing the same local DuplexConnection over all tests?

NiteshKant:
@yschimke `ClientSetupRule` reconnects to a new socket (new ID generated randomly) for every test. The code is [here](https://github.com/NiteshKant/reactivesocket-java/blob/d4047ddca1f243e374d4bcacd7718922e26cbf8c/reactivesocket-test/src/main/java/io/reactivesocket/test/ClientSetupRule.java#L51) and [here](https://github.com/NiteshKant/reactivesocket-java/blob/local-split-transport/reactivesocket-transport-local/src/test/java/io/reactivesocket/local/LocalClientSetupRule.java#L38)

So, it doesn't use the same connection AFAICT.

yschimke:
I was looking at the tests before changes

```
public class ClientServerTest {

    static ReactiveSocket client;

    static ReactiveSocket server;

    @BeforeClass
    public static void setup() throws Exception {
        LocalServerReactiveSocketConnector.Config serverConfig = new LocalServerReactiveSocketConnector.Config("test", new ConnectionSetupHandler() {
            @Override
            public RequestHandler apply(ConnectionSetupPayload setupPayload, ReactiveSocket rs) throws SetupException {
                return new RequestHandler() {
                    @Override
```

PR #136, opened 2016-07-08
`FailingReactiveSocket` violates spec

NiteshKant:
#### Problem

`FailingReactiveSocket` was invoking `Subscriber.onError()` without calling `onSubscribe()` which is invalid as per reactive-streams spec:

https://github.com/reactive-streams/reactive-streams-jvm#api-components

```
This means that onSubscribe is always signalled, followed by a possibly unbounded number of onNext signals (as requested by Subscriber) followed by an onError signal if there is a failure, or an onComplete signal when no more elements are available—all as long as the Subscription is not cancelled.
```

#### Modification

Modified to use `Publishers.error()` which follows the spec correctly.

#### Result

Better adherence of spec, happy users :)


stevegury:
👍

PR #141, opened 2016-07-12
`ClientBuilder.build` is now asynchronous.

stevegury:
***Problem***
The ClientBuilder `build` API return a ReactiveSocket.
This ReactiveSocket is returned while connections are still in establishing
mode, then it will fail until the connections established.
The current solution is to call `Unsafe.awaitAvailability` which is cluncky.

***Solution***
Create a more robust API that return a `Publisher<ReactiveSocket>`, it is now
obvious that the API is asynchronous.

***Modification***
I found a bug in `sourceToFactory`, where I was calling `subscriber.onSubscribe`
before initializing the `current` Map.

### DIFF
--- a/reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
+++ b/reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
@@ -118,6 +118,10 @@ public class ClientBuilder<T> {
                         subscriber.onError(new IllegalStateException("Please configure the source!"));
                         return;
                     }
+                    if (executor == null) {
+                        subscriber.onError(new IllegalStateException("Please configure the executor!"));
+                        return;
+                    }
                     if (connector == null) {
                         subscriber.onError(new IllegalStateException("Please configure the connector!"));
                         return;
--- a/reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
+++ b/reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
@@ -23,7 +23,9 @@ import org.reactivestreams.Subscription;
 import java.util.*;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public class ClientBuilder<T> {
     private final ScheduledExecutorService executor;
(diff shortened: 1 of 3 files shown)


reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
```diff
@@ -104,26 +106,55 @@ private ClientBuilder(
...
+                        filterConnector = filterConnector
+                            .chain(socket -> new TimeoutSocket(socket, requestTimeout, requestTimeoutUnit, executor));
+                    }
+                    filterConnector = filterConnector.chain(DrainingSocket::new);
+
+                    Publisher<? extends Collection<ReactiveSocketFactory<T>>> factories =
+                        sourceToFactory(source, filterConnector);
+                    LoadBalancer<T> loadBalancer = new LoadBalancer<>(factories);
+
+                    scheduledFuture = executor.scheduleAtFixedRate(() -> {
```
NiteshKant: May be we should do null check for executor in `withExecutor`, since you are assuming it to be non-null.

reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
```diff
@@ -104,26 +106,55 @@ private ClientBuilder(
...
+                        if (loadBalancer.availability() > 0 && !cancelled.get()) {
+                            subscriber.onNext(loadBalancer);
+                            subscriber.onComplete();
+                            if (scheduledFuture != null) {
+                                scheduledFuture.cancel(true);
+                            }
+                        }
+                    }, 1L, 50L, TimeUnit.MILLISECONDS);
```
NiteshKant: nit: If we have a outage and no servers are available, this will spin. Should we exponentially backoff?

I think it is "good enough" for now.

reactivesocket-client/src/test/java/io/reactivesocket/client/ClientBuilderTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+public class ClientBuilderTest {
+
+    @Test(timeout = 10_000L)
+    public void testIllegalState() throws ExecutionException, InterruptedException {
+        // you need to specify the source and the connector
+        Publisher<ReactiveSocket> socketPublisher = ClientBuilder.instance().build();
+
+        CountDownLatch latch = new CountDownLatch(1);
+        socketPublisher.subscribe(new Subscriber<ReactiveSocket>() {
```
NiteshKant: you can use RxJava2 `TestSubscriber`, makes the code cleaner.

NiteshKant:
👍  apart from the small comments.

reactivesocket-client/src/main/java/io/reactivesocket/client/ClientBuilder.java
stevegury: Yes, I think this is good enough.

PR #142, opened 2016-07-12
Remove reactivesocket-local-transport

stevegury:
***Problem***
Nobody is using this code, and I start to be tired of maintaining it.

***Solution***
Remove it, git keep the code in its history if we need it.


NiteshKant:
👍 

@robertroeser ok?

NiteshKant:
Based on offline discussions with @robertroeser we _may_ need this. 

@stevegury lets close this one and keep your branch in case we decide to remove.

stevegury:
Ok

PR #144, opened 2016-07-12
Cleanup of unused classes in `io.reactivesocket.internal.rx`

stevegury:
***Problem***
We have a lot of copy/pasted classes from RxJava 2, that we're not using.

***Soltion***
Remove them.


NiteshKant:
Yay 🎉

PR #145, opened 2016-07-13
Remove `ReactiveSocketFactory.remote()`

stevegury:
***Problem***
Since the refactoring of the `LoadBalancer` to use `List` instead of `Map`,
there's no need for a `remote` method that identify a remote server.
Furthermore, that's the only reason why LoadBalancer is generic.

***Solution***
Delete the method `ReactiveSocketFactory.remote()` as well as all the
references to it.


NiteshKant:
Yay 🎉

PR #146, opened 2016-07-13
int overflow in `requestN` for `Responder`

NiteshKant:
#### Problem

`ReactiveSocket` expects `requestN` to be an `int` whereas `ReactiveStreams` expects it to be a `long`.
`Responder` does not correctly convert `long` to `int` for values greater than `Integer.MAX_VALUE`. This sends `-1` as the `requestN` value of the `RequestN` frame.

#### Modification

If the value is greater than `Integer.MAX_VALUE`, use `Integer.MAX_VALUE` instead.

#### Result

We will not send invalid values over the wire for `RequestN`.

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
@@ -735,11 +735,11 @@ public class Responder {
                                     if (rn.intValue() > 0) {
                                         // initial requestN back to the requester (subtract 1
                                         // for the initial frame which was already sent)
-                                        child.onNext(Frame.RequestN.from(streamId, rn.intValue() - 1));
+                                        child.onNext(Frame.RequestN.from(streamId, Math.min(Integer.MAX_VALUE, rn.intValue() - 1)));
                                     }
                                 }, r -> {
                                     // requested
-                                    child.onNext(Frame.RequestN.from(streamId, r.intValue()));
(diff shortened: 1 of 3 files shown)


reactivesocket-core/src/main/java/io/reactivesocket/internal/Responder.java
```diff
@@ -735,11 +735,11 @@ public void request(long n) {
                                     if (rn.intValue() > 0) {
                                         // initial requestN back to the requester (subtract 1
                                         // for the initial frame which was already sent)
-                                        child.onNext(Frame.RequestN.from(streamId, rn.intValue() - 1));
+                                        child.onNext(Frame.RequestN.from(streamId, Math.min(Integer.MAX_VALUE, rn.intValue() - 1)));
```
stevegury: Actually, based on the spec, the `request-n` is a 64bits value (it should actually be a 63bits, but that's another story).
https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md#request-n-frame

So instead of downcasting everything to int, we upcasting everything to long.
The `RequestNFrame` is bogus.

stevegury: I take it back, ReactiveSocket protocol defines the request-n as a 32bits.

NiteshKant: Yes it is an integer value. So this is ready to merge then?

stevegury: Yes it is.

PR #148, opened 2016-07-14
Loadbalancer: closing doesn't subscribe to the underlying

stevegury:
***Problem***
Closing the loadbalancer doesn't properly subscribe to the `Publisher`s
returned by the `close()` methods of the underlying `ReactiveSocket`.
Thus, the close event is lost at the LoadBalancer level.

***Solution***
Properly subscribe to the close `Publisher`s and propagate the `onComplete`
events when all `ReactiveSocket` are closed. (I choose to ignore any exception
that happened during the close, i.e. only propagate 1 `onComplete`).

### DIFF
--- a/reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
+++ b/reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
@@ -27,6 +27,8 @@ import io.reactivesocket.exceptions.TransportException;
 import io.reactivesocket.internal.EmptySubject;
 import io.reactivesocket.internal.Publishers;
 import io.reactivesocket.internal.Publishers;
+import io.reactivesocket.internal.rx.EmptySubscriber;
+import io.reactivesocket.internal.rx.EmptySubscription;
 import io.reactivesocket.rx.Completable;
 import io.reactivesocket.client.stat.FrugalQuantile;
 import io.reactivesocket.client.stat.Quantile;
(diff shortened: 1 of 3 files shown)


reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -468,22 +471,44 @@ public synchronized String toString() {
...
+                            s.request(Long.MAX_VALUE);
+                        }
+
+                        @Override
+                        public void onNext(Void aVoid) {}
+
+                        @Override
+                        public void onError(Throwable t) {
+                            if (n.decrementAndGet() == 0) {
+                                subscriber.onComplete();
```
nickmahilani: can propagate the onError to the subscriber here?

nickmahilani:
one minor comment, LGTM

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -468,22 +471,44 @@ public synchronized String toString() {
...
+                                closeSubject.subscribe(EmptySubscriber.INSTANCE);
+                                closeSubject.onComplete();
+                            }
+                        }
+
+                        @Override
+                        public void onComplete() {
+                            if (n.decrementAndGet() == 0) {
```
robertroeser: you're doing the same thing in the onComplete and the onError - why not just call onComplete in the onError method?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: Good point.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
stevegury: Good point, it makes the code simpler.

PR #151, opened 2016-07-15
Fix Error flags to match the spec + review the `frame` package.

stevegury:
***Problem***
There is an error in the Error frame, the error codes used for `APPLICATION_ERROR`,
`CONNECTION_ERROR`, `REJECTED`, `CANCELED` and `INVALID` were invalid.

***Solution***
Update the constants.
I don't think it requires doing any version bump, since the error code was not used
prior to #150.

***Modification***
I also took the oportunity to review the `frame` package and update it to the Java
code style.

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/internal/frame/ErrorFrameFlyweight.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/internal/frame/ErrorFrameFlyweight.java
@@ -39,11 +39,11 @@ public class ErrorFrameFlyweight {
     public static final int INVALID_SETUP = 0x0001;
     public static final int UNSUPPORTED_SETUP = 0x0002;
     public static final int REJECTED_SETUP = 0x0003;
-    public static final int CONNECTION_ERROR = 0x0011;
-    public static final int APPLICATION_ERROR = 0x0021;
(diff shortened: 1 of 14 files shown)


reactivesocket-core/src/main/java/io/reactivesocket/internal/frame/ErrorFrameFlyweight.java
```diff
@@ -39,11 +39,11 @@ private ErrorFrameFlyweight() {}
     public static final int INVALID_SETUP = 0x0001;
     public static final int UNSUPPORTED_SETUP = 0x0002;
     public static final int REJECTED_SETUP = 0x0003;
-    public static final int CONNECTION_ERROR = 0x0011;
-    public static final int APPLICATION_ERROR = 0x0021;
+    public static final int CONNECTION_ERROR = 0x0101;
+    public static final int APPLICATION_ERROR = 0x0201;
     public static final int REJECTED = 0x0022;
```
yschimke: why not fix Rejected?

stevegury: oversight.

PR #152, opened 2016-07-15
Fix REJECTED error code to match the spec.

stevegury:
It should have been fixed in #151, that was an oversight from my part.


stevegury:
Thanks @yschimke for spotting this!

PR #153, opened 2016-07-19
merging tck drivers with new code

xytosis:
I would like to integrate the TCK Java Drivers into this repository, as I think it would be the best organization to have each driver for each implementation live inside their own repos.

The problem I am having is that, for some reason, the channel tests that I wrote that worked for the 2.2 snapshot of the TCP transport does not work with the new TCP code, and the toy server and client in test just doesn't work at all (the client seems to randomly timeout before sending or receiving data from the server).

There is probably something wrong with the way I wrote the code. I think if I can get SimpleServer and Main in the `test` directory working, the driver code should start working too.


reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.concurrent.CountDownLatch;
+
+import static java.net.InetSocketAddress.createUnresolved;
+
+// this client should parse the test cases we wrote and use them to
+public class JavaTCPClient {
+
+    public static void main(String[] args) throws MalformedURLException, URISyntaxException {
+        // we pass in our reactive socket here to the test suite
+        String file = "/Users/mjzhu/dev/reactivesocket-java/reactivesocket-tck-drivers/clienttest$.txt";
```
yschimke: avoid hardcoded paths, a relative path assuming you run the build just inside of reactivesocket-java is probably safer for IDE and gradle use.

reactivesocket-tck-drivers/clienttest$.txt
```diff
@@ -0,0 +1,111 @@
+!
```
yschimke: Can you move these under src/(main|test)/resources?

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/test/PayloadImpl.java
```diff
@@ -0,0 +1,49 @@
+package io.reactivesocket.tckdrivers.test;
+
+import io.reactivesocket.Payload;
+
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+
+public class PayloadImpl implements Payload // some JDK shoutout
```
yschimke: maybe for a separate change, but I believe this move allows us to use io.reactivesocket.util.PayloadImpl

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/server/JavaTCPServer.java
```diff
@@ -0,0 +1,41 @@
...
+package io.reactivesocket.tckdrivers.server;
+
+import io.reactivesocket.transport.tcp.server.TcpReactiveSocketServer;
+
+
+public class JavaTCPServer {
+
+    public static void main(String[] args) {
+
+        String file = "/Users/mjzhu/dev/reactivesocket-java/reactivesocket-tck-drivers/servertest$.txt";
```
yschimke: as above, relative path and hopefully not in module root.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/PayloadImpl.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+package io.reactivesocket.tckdrivers.common;
+
+import io.reactivesocket.Payload;
+
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+
+public class PayloadImpl implements Payload {
```
yschimke: as below, I think we can reuse the util.PayloadImpl now.  Could be for a later change.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/ParseThread.java
```diff
@@ -0,0 +1,34 @@
...
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *  <p>
+ *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ *  specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivesocket.tckdrivers.common;
+
+public class ParseThread implements Runnable {
```
yschimke: I'd avoid code that manually creates threads and even implementing runnable.  This class could just be an inline lambda passed to an ExecutorService.

yschimke:
Is it working as is? Or you still have the issue with channel tests?

Sorry for the driveby review, I'm just back from PTO, getting back up to speed.  I haven't gone through the whole thing and not sure if it previously went through review and discussion in this form in other repo and we just want to do an initial code drop.

reactivesocket-tck-drivers/clienttest$.txt
xytosis: Will do, I left them here temporarily to test the driver

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
xytosis: That makes sense, I was going to allow users to input a command line argument for a path, but I was testing and forgot to change it. Will do it soon.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/ParseThread.java
xytosis: I've mostly worked with threads in C before; is ExecutorService the preferred way to do so in Java? I can change this after I get the drivers working again with channel, just so I don't have to potentially debug two things at once.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/PayloadImpl.java
xytosis: Sounds good, I'll keep this here for now and integrate later.

xytosis:
@yschimke No worries, there's been a few changes since the last time you've looked at the code, but we haven't gone through a rigorous review/discussion yet.

Also, the code is still not working with channel tests after switching to the newest code. Everything works with the 2.2 snapshot, but there seems to be a problem with the ReactiveSocket not waiting and the main thread just exits. In the old code I used ReactiveSocket.startAndWait() but in the new code I'm not sure what to use. If anyone could take a look at the toy server and client in the `test` package and help me get that working, I think I can fix everything else.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/TestSubscriber.java
```diff
@@ -0,0 +1,1113 @@
...
+     */
+    public final boolean await(long time, TimeUnit unit) throws InterruptedException {
+        if (done.getCount() == 0) {
+            return true;
+        }
+        return done.await(time, unit);
+    }
+
+    public final boolean awaitAtLeast(long n, long time, TimeUnit unit) throws InterruptedException {
+        numOnNext.await(time, unit);
```
yschimke: I suggest checking the return values for both these await calls, and making sure the test terminates cleanly if there is some failure causing a missing event.

yschimke:
I suggest also stealing the logging configuration from reactivesocket-cli and making that optionally shown when running the test, so its easier to understand the failure without stepping through a debugger

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    public static ReactiveSocket createClient() {
+        try {
+            String target = "tcp://localhost:4567/rs";
+            URI uri = new URI(target);
+
+            Publisher<ReactiveSocket> socketPublisher = buildConnection(uri);
+            SocketWrapper client = new SocketWrapper();
```
stevegury: You could replace everything below by:
`Unsafe.blockingSingleWait(socketPublisher, 5, TimeUnit.SECONDS);`

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/TestSubscriber.java
xytosis: I can do that; I can also look to creating global timeout values as well

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
xytosis: Thanks for reviewing my code @stevegury! I used similar and also cleaner line of code from Yuri's CLI project, which has the added benefit of allowing me to add a logger to see the individual frames being sent.

xytosis:
@yschimke thanks for all the support you gave me earlier, I was able to find the source of the bug, and the code should be working as is. I will implement the changes you've suggested

yschimke:
@xytosis No problem, hopefully this failure you had was a useful test of future implementors experience using the TCK.  

One cherry on top if you have a couple of days free would be to implement the Client Tests as Junit4 Parameterized tests e.g. http://stackoverflow.com/a/359004/1542667  This would be really nice integration in an IDE and allow you to run the tests and just re-run or debug a single failing test case.

But not essential, and specifically not for this review.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
```diff
@@ -0,0 +1,517 @@
+/*
+ * Copyright 2016 Facebook, Inc.
```
yschimke: I believe we agreed to put Netflix as copyright holder for submissions to this repo for simplicity.  cc @benjchristensen

yschimke:
Could you add running instructions to reactivesocket-tck-drivers/README.md

Generally I'm ok with this landing sooner rather than later, as in a separate repo we are likely to hit issues continually do to refactoring changes.  But I'd like a thumbsup from @NiteshKant or @stevegury rather than just mine.  They are likely to suffer more of the maintenance burden when changes to the public java API affect the TCK.

NiteshKant:
I am on a PTO, so haven't looked at the code. 
Philosophically, what is the reason for having multiple TCK drivers and co-location with respective repos?
If the intent to co-locate to reduce instability with breaking changes, would this continue to exist after we stabilize the implementation?

xytosis:
@yschimke Thanks for suggesting the parameterized tests, it would definitely be neat to have and I can look into it once I have documentation set up and when I stabilize the code. I will also start adding documentation today, as well as continue to add documentation to https://github.com/ReactiveSocket/reactivesocket-tck

@NiteshKant Yes, separating the drivers to live with their respective repos would increase maintainability as well as keep the TCK repo polyglot. So ideally we would have wanted to use a stable implementation to implement this driver, but since you are still making changes to the code, every breaking change would have to be reflected within the driver. This would require people to understand how the driver works, but as of now I think the implementation of the driver is mostly straightforward enough for that to not be too big of an issue.

reactivesocket-tck-drivers/gradlew
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env bash
```
yschimke: you should only need gradlew at the top level

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
```diff
@@ -29,26 +29,14 @@
...
-    @Option(name = "--file", description = "test file")
-    public static String realfile;
-
-    public static void main(String[] args) throws MalformedURLException, URISyntaxException {
-
-        SingleCommand<JavaTCPClient> cmd = SingleCommand.singleCommand(JavaTCPClient.class);
-        cmd.parse(args);
+    public static void run(String realfile, String host, int port, boolean debug2)
+            throws MalformedURLException, URISyntaxException {
+        debug = debug2;
```
yschimke: rename param and use "this.debug = debug;"

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/main/Main.java
```diff
@@ -0,0 +1,45 @@
...
+    public static void main(String[] args) {
+        SingleCommand<Main> cmd = SingleCommand.singleCommand(Main.class);
+        cmd.parse(args);
+        if (server) {
+            JavaTCPServer.run(file, port);
+        } else if (client) {
+            try {
+                JavaTCPClient.run(file, host, port, debug);
+            } catch (Exception e) {
+                System.out.println("Could not launch client");
```
yschimke: this seems anti-helpful.  Either catch specific Exceptions where you can provide a simpler and expected error e.g. FileNotFoundException -> "input file not found", or at least print the exception stacktrace.

reactivesocket-tck-drivers/run
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+java -cp build/libs/reactivesocket-tck-drivers-0.2.2-SNAPSHOT.jar io.reactivesocket.tckdrivers.main.Main $@
```
yschimke: For scripts the convention is "$@" which keeps the same arguments that were passed in.  I know its counter intuitive, since you expect it to munge them into one.

Try using: https://www.shellcheck.net/ or install with "brew install shellcheck"

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
```diff
@@ -0,0 +1,78 @@
...
+public class JavaTCPClient {
+
+    private static URI uri;
+    private static boolean debug;
+
+    public static void run(String realfile, String host, int port, boolean debug2)
+            throws MalformedURLException, URISyntaxException {
+        debug = debug2;
+        // we pass in our reactive socket here to the test suite
+        String file = "reactivesocket-tck-drivers/src/main/test/resources/clienttest$.txt";
```
yschimke: I was using a pseudo regex format in my comment, I meant either test or main.

> Can you move these under src/(main|test)/resources?

reactivesocket-tck-drivers/gradlew
xytosis: I'll get rid of this and try to match the other subprojects as closely as I can.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/main/Main.java
xytosis: That's a great point, I'll change this so it's more helpful.

reactivesocket-tck-drivers/run
xytosis: Sounds good, I'll change this and check out shellcheck.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
xytosis: My bad, I'll change this

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Tue Mar 15 03:05:19 MSK 2016
+#Mon Jul 18 17:19:35 PDT 2016
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.12-bin.zip
```
benjchristensen: Why is this changed?

reactivesocket-tck-drivers/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Thu Jul 21 10:57:24 PDT 2016
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.13-bin.zip
```
benjchristensen: Why is this file needed here?

reactivesocket-tck-drivers/run
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
benjchristensen: I suggest appending the .sh suffix to the file name.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
benjchristensen: The Facebook one should be on all these files. Copyrights are whoever submits/edits/touches a file, so there can be multiple at the top of the file.

If someone doesn't care about their copyright, then they can "assign" it to the default Netflix, but we use the Facebook copyright. 

We are however adopting the Apache 2 license since that's what this project uses.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/EchoSubscription.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.reactivesocket.Payload;
+import io.reactivesocket.util.PayloadImpl;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+public class EchoSubscription implements Subscription {
```
benjchristensen: It would help to have a javadoc explaining what this class is for.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/ParseThread.java
benjchristensen: I'm going to provide examples with Rx as ParseMarble is begging for the Observable.create(SyncOnSubscribe) generator. That's exactly what it's trying to do ... be a ReactiveStreams Publisher.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/TestSubscriber.java
```diff
@@ -0,0 +1,1115 @@
...
+            if (s != null) {
+                long mr = missedRequested.getAndSet(0L);
+                if (mr != 0L) {
+                    s.request(mr);
+                }
+            }
+        }
+    }
+
+    public final void setEcho(EchoSubscription echosub) {
```
benjchristensen: This seems odd to need a one-off concrete implementation here. Why does TestSubscriber need to know about this?

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/main/Main.java
benjchristensen: e.printStackTrace() would be preferable to that line.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/ParseMarble.java
```diff
@@ -0,0 +1,165 @@
...
+import io.reactivesocket.util.PayloadImpl;
+import org.reactivestreams.Subscriber;
+
+import java.util.*;
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * This class is rather straight forward for most interactions except for channel.
+ */
+public class ParseMarble {
```
benjchristensen: This can probably be turned into an `Iterable` that parses the file. Then you can just do this:

```java
Publisher p = RxReactiveStreams.toPublisher(Observable.from(marbleIterable));
```

Or if something more custom is needed:

```java
    Observable
        .<String> create(SyncOnSubscribe.<Iterator<String>, String> createStateful(
            () -> {
              // parse state ... (don't do this if just an Iterable ... use the above simpler example
              ArrayList<String> list = new ArrayList<String>();
              list.add("one");
              list.add("two");
              list.add("three");
              return list.iterator();
            } ,
            (iter, obs) -> {
              // state == position in iterable

             // this gets called for each onNext to be done so custom parsing could occur

              if (iter.hasNext()) {
                obs.onNext(iter.next());
              } else {
                obs.onCompleted();
              }
              return iter;
            }))
```

gradle/wrapper/gradle-wrapper.properties
xytosis: I'm not too sure... it might have changed when I complied the project as I never opened this file before.

reactivesocket-tck-drivers/gradle/wrapper/gradle-wrapper.properties
xytosis: I'll remove it.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/EchoSubscription.java
xytosis: This was for setting up echo tests, but I might be able to get rid of this after changing parseMarble

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/TestSubscriber.java
xytosis: You're right, I probably can get rid of this to after changing ParseMarble

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
xytosis: It's static so I can't do that, but I'll do something cleaner.

NiteshKant:
@xytosis IMHO if we do not see the TCK driver code in the main Java repo once the Java repo is stabilized then probably it shouldn't belong here now. 
From stability and overhead of catching up with backward incompatible changes point of view all repos have that issue. Protocol is changing and so are the the implementations.

benjchristensen:
I don't understand your comment Nitesh. The point is that the TCK repo should not have all language implementations of TCK drivers. That would make that project nutty with all those languages. The TCK repo contains the TCK test script generator. Then each RS impl also needs the TCK driver in its language. This one is for Java. The C++ repo will end up with its own driver impl. Same with JS. Etc.

xytosis:
@NiteshKant I'm not sure I understand what you mean either. I understand that the protocol may change, but it would make sense to have reactivesocket-java specific stuff in this repo, and it would be better and more useful if it were updated here along with the rest of the code instead of sitting in some other place and becomes obsolete because it's forgotten about.

NiteshKant:
Nevermind I misunderstood the intent of the TCK driver. Apologies for the confusion.

gradle/wrapper/gradle-wrapper.properties
yschimke: Can you revert this unless you know why?  I think I've seen Intellij make this change before.

yschimke:
I'm in favour of landing this and iterating with smaller more understandable diffs.  Let's give @NiteshKant and @stevegury today to object, otherwise land and iterate.  Worst case we can temporarily disable the module if it causes build issues.

xytosis:
@yschimke, apologies for the late reply, I was out of the office yesterday. However, it doesn't seem like there were any objections, so I will do a quick pass through the code for some last minute changes/fixes and then land it.

gradle/wrapper/gradle-wrapper.properties
xytosis: Ok, I can manually revert the change.

stevegury: ditto

reactivesocket-tck-drivers/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+        configurations.compile.collect { it.isDirectory() ? it : zipTree(it) }
+    }
+}
+
+dependencies {
+
+    compile project(':reactivesocket-core')
+    compile project(':reactivesocket-client')
+    compile project(':reactivesocket-transport-tcp')
+    compile project(':reactivesocket-core')
```
stevegury: duplicate

reactivesocket-tck-drivers/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+    compile 'io.netty:netty-all:4.1.3.Final'
```
stevegury: Remove this and rely on the transitive dependency from reactivesocket-transport-tcp

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
```diff
@@ -0,0 +1,554 @@
...
+                    break;
+            }
+            line = reader.readLine();
+        }
+        tests.add(test);
+        tests = tests.subList(1, tests.size()); // remove the first list, which is empty
+        for (List<String> t : tests) {
+            TestThread thread = new TestThread(t);
+            thread.start();
+            thread.join();
```
stevegury: I'm not sure to understand why you use threads here?
(since you start and immediately wait)

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
```diff
@@ -0,0 +1,554 @@
...
+        // this check each of the subscribers to see that they all passed their assertions
+        if (id.size() > 0) {
+            boolean hasPassed = true;
+            for (String str : id) {
+                if (payloadSubscribers.get(str) != null) hasPassed = hasPassed && payloadSubscribers.get(str).hasPassed();
+                else hasPassed = hasPassed && fnfSubscribers.get(str).hasPassed();
+            }
+            return Optional.of(hasPassed);
+        }
+        else return Optional.empty();
```
stevegury: Maybe a better return type would be `boolean`, and use an exception for when `id.size() == 0`

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
```diff
@@ -0,0 +1,554 @@
...
+            @Override
+            public void subscribe(Subscriber<? super Payload> s) {
+                ParseMarble pm = new ParseMarble(s);
+                TestSubscription ts = new TestSubscription(pm, initialPayload, s);
+                s.onSubscribe(ts);
+                ParseChannel pc = new ParseChannel(commands, testsub, pm, name);
+                ParseChannelThread pct = new ParseChannelThread(pc);
+                pct.start();
+                mypct.set(pct);
+                c.countDown();
```
stevegury: It seems you don't wait for the thread to finish?
Also, why using a Thread, could you use synchronous code here?

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
```diff
@@ -0,0 +1,554 @@
...
+        } catch (InterruptedException e) {
+            System.out.println("interrupted");
+        }
+        mypct.get().join();
+    }
+
+    /**
+     * Trivial class that wraps around the ParseChannelThread
+     */
+    private class PCTWrapper {
```
stevegury: You can use AtommicReference instead

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
xytosis: My idea is that if there are any exceptions, they would be contained within the thread and would allow subsequent tests to run.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
xytosis: That makes sense, I'll change it to that

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
xytosis: I tried both waiting for the thread inside the publisher as well as using synchronous code, but both caused the tests to hang because (my assumption is) that waiting for the `parse()` method to finish inside the publisher delays the actual creation of the publisher until the parse method is finished, which means that while the server is trying to send the client data, the client has not even created it's own publisher and used it to subscribe.

Having the thread run asynchronously from the driver fixes the problem, and I wait for the thread to finish outside of the construction of the publisher.

stevegury: Ok, I understand what you're doing.

yschimke:
OK, merging now.  since its fairly isolated the cost of a blocking issue is small and cheap to resolve.  But its a lot easier to deal with smaller incremental diffs.

PR #154, opened 2016-07-21
Make RequestHandler an interface

rdegnan:
Making RequestHandler an interface rather than an abstract class for consistency with ReactiveSocket, makes it more flexible to extend.


stevegury:
👍

PR #155, opened 2016-07-26
Fix SETUP flags according to the spec

stevegury:
***Problem***
The flags used in the SETUP frame are different from the one defined in the specification.
They are defined as a byte (8bits) vs. a short (16bits)
(See https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md#setup-frame)

***Solution***
Update the constant to the right value.


yschimke:
Since this was the src/main part of this PR https://github.com/ReactiveSocket/reactivesocket-java/pull/104/files, am I ok to land the test part now?

stevegury:
@yschimke yes, this is ok, this part of the change actually didn't break anything since we were not enabling the lease.

PR #161, opened 2016-07-31
Updating TCK slightly

xytosis:
Gets rid of most non-deterministic errors. If any happen, rerun server and client and should disappear. The exception is the NPE in UnicastSubject. Not sure why the subscriber inside would be null and I don't think it's a bug on my end since it's happening in a reactivesocket.internal class, but currently unable to reproduce the error without using TCK driver.


reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
```diff
@@ -84,6 +84,13 @@ public void runTests() throws IOException {
             }
             line = reader.readLine();
         }
+        // "priming' server, seems to reduce the frequency of random errors
```
yschimke: "priming' probably doesn't need to air quotes...

Also for a TCK, I'd generally argue against fixes that "reduce the frequency of random errors. I would argue you want either 0 random errors or easily reproducible errors.

reactivesocket-tck-drivers/run.sh
```diff
@@ -1,3 +1,7 @@
 #!/bin/bash
 
-java -cp build/libs/reactivesocket-tck-drivers-0.2.2-SNAPSHOT.jar io.reactivesocket.tckdrivers.main.Main "$@"
+LATEST_VERSION=$(ls build/libs/ | sort -r | head -1)
```
yschimke: This could be perfectly safe, but looks brittle given the libs directory isn't just one file for a clean build.  I would suggest either tightening up the ls command to generally only find one file, or look at a command like this

http://stackoverflow.com/questions/24936781/gradle-plugin-project-version-number

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
xytosis: I was actually wrong about this line of code, but one of my other changes did get rid of random errors so they are virtually nonexistent now (before I was doing things in an order where the chance of a data race was extremely high). IMO because it's a network protocol there is a > 0% chance of something failing somewhere, but you're right, I phrased this incorrectly, and will change it.

reactivesocket-tck-drivers/run.sh
xytosis: That makes sense, I'll change this.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaClientDriver.java
yschimke: You have a 5 second timeout in there, so if the network glitches are meaning that we are seeing failures, then we want to see those problems more, not less :)

reactivesocket-tck-drivers/src/test/resources/client$.txt
```diff
@@ -1,262 +1,6 @@
...
-name%%requestResponsePass
-pass
-subscribe%%rr%%0e65a66f-0322-41e1-925c-b9e55b0db7f4%%a%%b
-request%%1%%0e65a66f-0322-41e1-925c-b9e55b0db7f4
-await%%atLeast%%0e65a66f-0322-41e1-925c-b9e55b0db7f4%%1%%100
-assert%%no_error%%0e65a66f-0322-41e1-925c-b9e55b0db7f4
-assert%%completed%%0e65a66f-0322-41e1-925c-b9e55b0db7f4
-assert%%received%%0e65a66f-0322-41e1-925c-b9e55b0db7f4%%a,a
+subscribe%%fnf%%746757d6-8e87-4673-b000-34e7fb305c33%%a%%b
+request%%1%%746757d6-8e87-4673-b000-34e7fb305c33
```
yschimke: is this change deliberate?  seems to be a lot smaller.

xytosis: Oops sorry, I was just deleting some parts of the script so I could run a single test. I can look into modifying the driver so that you can choose what test to run

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/main/Main.java
```diff
@@ -31,14 +33,18 @@
             "correct files")
```
yschimke: typo: missing space "thecorrect"

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/main/Main.java
```diff
@@ -31,14 +33,18 @@
...
     public static String file;
 
+    @Option(name = "--tests", description = "For the client only, optional argument to list out the tests you" +
+            " want run, should be comma separated names")
```
yschimke: "want *to* run"

yschimke:
Looks ok, submit after a few fixes.

xytosis:
Thanks Yuri, since I don't have write access to the repo, could you merge for me?

xytosis:
I also added some basic benchmarking, which I can expand on later after merging this PR

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/server/JavaServerDriver.java
```diff
@@ -125,22 +131,23 @@ public RequestHandler parse() {
...
             try {
-                System.out.println("Channel");
                 TestSubscriber<Payload> sub = new TestSubscriber<>();
                 payloadPublisher.subscribe(sub);
                 // want to get equivalent of "initial payload"
                 //sub.request(1); // first request of server is implicit, so don't need to call request(1) here
                 sub.awaitAtLeast(1);
                 Tuple<String, String> initpayload = new Tuple<>(sub.getElement(0).getK(), sub.getElement(0).getV());
-                System.out.println(initpayload.getK() + " " + initpayload.getV());
+                System.out.println(ANSI_CYAN + "Received Channel" + initpayload.getK()
```
yschimke: In a later change could you pull this out into a utility that expresses the intent e.g. `ConsoleUtils.info("Received Channel" + initpayload.getK())`

xytosis: That makes sense, so it'll basically be some sort of logger but for this specifically

PR #163, opened 2016-08-02
more app errors please

yschimke:
For discussion:

I'd argue this is the right code because its the behaviour you get with

        requestHandlerBuilder.withRequestResponse(payload ->
                RxReactiveStreams.toPublisher(Observable.error(new Exception("server failure")))).build();

We can't know all errors ahead of time, and I don't think we should ask users to map things to ApplicationError.


stevegury:
I agree, this is the right behavio~~u~~r.

PR #164, opened 2016-08-04
Benchmarking

xytosis:


xytosis:
My bad, clicked the wrong button

yschimke:
Could you work out why you have this ever growing list of commits above?  I suspect you should delete/move your local git master, and take a fresh copy from ReactiveSocket:master.  But that is a guess.  e.g. 

```
(master)$ git branch -m old_master
(old_master)$ git co -t origin/master
Branch master set up to track remote branch master from origin.
Switched to a new branch 'master'
```

xytosis:
I was thinking that wasn't normal either. I'll look into what you suggested and get rid of it.

PR #167, opened 2016-08-08
Updating TCK to be run in travis

xytosis:
Added a new TestMain to run both the client and the server, and also added ability for client to tell the server to shut down, so that when a test is run, the server doesn't hang forever trying to await shutdown.


xytosis:
@yschimke I'm not too familiar with gradle and travis. My understanding is that since the build script for travis just calls `gradlew build`, that script will only return 1 if the something in the build crashes. My tests are made not to crash the build though, so is there someway I can "return 1" to the gradle task that runs my test, and then have that gradle task tell the build process there was an error?

yschimke:
@xytosis 

```
public class ExitDemo {
//                System.exit(0); // passes
//                System.exit(1); // fails
                throw new RuntimeException("ex"); // fails and logs exception
    }
}
```

reactivesocket-tck-drivers/build.gradle
```diff
@@ -19,10 +18,18 @@ dependencies {
...
     compile 'io.airlift:airline:0.7'
 }
 
+task (runTests, dependsOn: jar, type: JavaExec) {
+    classpath(sourceSets.main.runtimeClasspath, sourceSets.main.compileClasspath)
+    main = 'io.reactivesocket.tckdrivers.main.TestMain'
+    args '--port',  '4567', '--serverfile',  'src/test/resources/server$.txt',  '--clientfile',  'src/test/resources/client$.txt'
+}
+
+build.finalizedBy(runTests)
```
yschimke: I don't think this is an appropriate use of finalizedBy, it looks like its meant for cleanup e.g. stopping a SQL db you launched and used multiple times in a test run.

e.g.

http://buransky.com/programming/integration-testing-with-gradle/
http://selimober.com/blog/2014/01/24/separate-unit-and-integration-tests-using-gradle/

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/server/JavaTCPServer.java
```diff
@@ -20,22 +20,29 @@
...
             file = realfile;
         }
 
+        TcpReactiveSocketServer server = TcpReactiveSocketServer.create(port);
+
         JavaServerDriver jsd =
-                new JavaServerDriver(file);
+                new JavaServerDriver(file, server);
+
+        /*server.start((setupPayload, reactiveSocket) -> {
```
yschimke: is this dead code now?

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/main/TestMain.java
```diff
@@ -0,0 +1,50 @@
...
+    public static String tests;
+
+    public static void main(String[] args) {
+        SingleCommand<TestMain> cmd = SingleCommand.singleCommand(TestMain.class);
+        cmd.parse(args);
+        ServerThread st = new ServerThread(port, serverfile);
+        st.start();
+
+        try {
+            Thread.sleep(2000); // wait for server to start up, 2 seconds to be safe
```
yschimke: If this is because ServerThread is running something in another thread and needs to get to some point before we continue, I'd prefer to add a hook into ServerThread to e.g. st.waitUntilReady()

These magic values are either noisy e.g. they fail in CI 2% of the time, or they are slow and cause local builds to take longer.

reactivesocket-tck-drivers/build.gradle
xytosis: That makes sense, I can run these tests the other ways.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/server/JavaTCPServer.java
xytosis: Yes, I will remove this line

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/main/TestMain.java
xytosis: That makes sense, the waiting was to ensure that the server had launched, but I agree adding in a hook would be a cleaner and safer way to do things.

reactivesocket-tck-drivers/build.gradle
```diff
@@ -26,10 +26,12 @@ dependencies {
...
+task runTests(type: JavaExec) {
     classpath(sourceSets.main.runtimeClasspath, sourceSets.main.compileClasspath)
     main = 'io.reactivesocket.tckdrivers.main.TestMain'
     args '--port',  '4567', '--serverfile',  'src/test/resources/server$.txt',  '--clientfile',  'src/test/resources/client$.txt'
 }
 
-build.finalizedBy(runTests)
+tasks.build.doLast() {
+    runTests.execute()
+}
```
xytosis: @yschimke I'm not sure the links you sent about integration tests could execute the behavior I wanted. In those cases, it seemed like the tests had to be explicitly called from the command line with `gradle <testname>`. However, I wanted it to behave like the other unit tests that run automatically during the travis build process, and this seemed like the cleanest way to do so.

reactivesocket-tck-drivers/build.gradle
```diff
@@ -19,10 +18,20 @@ dependencies {
...
     compile 'io.airlift:airline:0.7'
 }
 
+task runTests(type: JavaExec) {
+    classpath(sourceSets.main.runtimeClasspath, sourceSets.main.compileClasspath)
+    main = 'io.reactivesocket.tckdrivers.main.TestMain'
+    args '--port',  '4567', '--serverfile',  'src/test/resources/server$.txt',  '--clientfile',  'src/test/resources/client$.txt'
+}
+
+tasks.build.doLast() {
```
yschimke: Isn't this just "build.dependsOn runTests"? or does it do something surprising/

xytosis: You're right, the behavior is the same. I tried it earlier but I must have done something incorrectly.

xytosis:
I don't know why the interleave test is suddenly failing for no reason on travis. It works perfectly on my own machine...

yschimke:
@xytosis if tests are flaky in Travis, I think we owe it to the other developers to spend a bit of time investigating in case there is some inherent problem that will cause pain of occasional broken builds for them.  Also it should translate into good experience investigating the output of broken builds.

xytosis:
@yschimke that makes sense, if the functionality occasionally fails it should definitely be patched up or else the same behavior could happen when it's actually being used.

xytosis:
Do you think this diff is ready to be landed?

yschimke:
I would rather understand why some of the tests are failing before landing it.

xytosis:
It definitely is some data race, I documented what I found in https://github.com/ReactiveSocket/reactivesocket-java/issues/162

yschimke:
Thanks for pinning down the cause of the broken test to the bug you raised.  If you drop that test for now and it passes travis ci. Then we can land this.  We can disable with a one line change if it looks to be noisy at all in future.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/server/JavaTCPServer.java
```diff
@@ -15,29 +15,43 @@
...
-    public static void run(String realfile, int port) {
+    private CountDownLatch mutex;
+
+    public JavaTCPServer() {
+        mutex = new CountDownLatch(1);
+    }
+
+    public void run(String realfile, int port) {
 
         String file = "reactivesocket-tck-drivers/src/main/test/resources/servertest$.txt";
```
yschimke: this path is wrong -main

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
```diff
@@ -34,7 +34,7 @@
     private static URI uri;
     private static boolean debug;
 
-    public static void run(String realfile, String host, int port, boolean debug2, List<String> tests)
+    public void run(String realfile, String host, int port, boolean debug2, List<String> tests)
             throws MalformedURLException, URISyntaxException {
         debug = debug2;
         // we pass in our reactive socket here to the test suite
```
yschimke: path below is wrong -main

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/server/JavaTCPServer.java
xytosis: Thanks for noticing, those were the old tests, I'll update this.

xytosis:
@yschimke sounds good, I'll push some changes and land, although my hope is that this is used in all builds in the future. Perhaps if some of the tests are too noisy we can reduce the number of them to testing just core functionality.

xytosis:
Can we land this diff?

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/server/JavaTCPServer.java
```diff
@@ -15,29 +15,43 @@
...
+    private CountDownLatch mutex;
+
+    public JavaTCPServer() {
+        mutex = new CountDownLatch(1);
+    }
 
-        String file = "reactivesocket-tck-drivers/src/main/test/resources/servertest$.txt";
+    public void run(String realfile, int port) {
+
+        String file = "reactivesocket-tck-drivers/src/main/test/resources/server$.txt";
```
yschimke: It looks like this path is still wrong.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
```diff
@@ -34,11 +34,11 @@
...
-        String file = "reactivesocket-tck-drivers/src/main/test/resources/clienttest$.txt";
+        String file = "reactivesocket-tck-drivers/src/main/test/resources/client$.txt";
```
yschimke: It looks like this path is still wrong.

xytosis:
My bad, the path should now be correct.

PR #170, opened 2016-08-09
Need to eventually add back these tests

xytosis:
Once this bug has been fixed:
https://github.com/ReactiveSocket/reactivesocket-java/issues/162
merge this branch so that the tests that are failing will pass.


yschimke:
You still have old commits showing up.  Can you look into why this keeps happening for you?

send me a fbpaste of first page of "git log master" and "git log pr_bug_fix"

yschimke:
I don't understand this branch.  I assume it #167 but without the tests at the end commented out?

If so, can we fix the tests first before integrating this into the build.

xytosis:
@yschimke sorry, I know why old commit are showing up, it's because I need to rebase from upstream. But in this case I didn't rebase because those commits that show up haven't been accepted yet, but thinking about it now, I probably should in order to avoid merge conflicts.

As for the purpose of the branch, it should just be the failing tests commented out. Let me rebase and do another request.

yschimke:
The commits looked very old. I did the same thing to @lehecka with overlapping PRs. So I share your pain. But can you check the earliest commits in this PR

xytosis:
@yschimke I don't think my local master is the problem, because I've tried that solution. It's the origin/master that has all the commits, and the only way to get rid of those is to rebase the fork on upstream/master or I guess delete the fork and refork. I just tried rebasing, and it seems like I got rid of the old commit.

xytosis:
Also, I don't think https://github.com/ReactiveSocket/reactivesocket-java/issues/162 has been marked as a bug yet. Would you be able to do that?

PR #173, opened 2016-08-17
updated so that server fails on unknown initial payload

xytosis:
Server notifies user if it receives an payload that it is not set up to handle

### DIFF
--- a/reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
+++ b/reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
@@ -38,7 +38,7 @@ public class JavaTCPClient {
             throws MalformedURLException, URISyntaxException {
         debug = debug2;
         // we pass in our reactive socket here to the test suite
-        String file = "reactivesocket-tck-drivers/src/test/resources/client$.txt";
+        String file = "reactivesocket-tck-drivers/src/test/resources/client.txt";
         if (realfile != null) file = realfile;
         try {
             setURI(new URI("tcp://" + host + ":" + port + "/rs"));
--- a/reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
+++ b/reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
@@ -38,7 +38,7 @@ public class JavaTCPClient {
             throws MalformedURLException, URISyntaxException {
         debug = debug2;
         // we pass in our reactive socket here to the test suite
-        String file = "reactivesocket-tck-drivers/src/test/resources/client.txt";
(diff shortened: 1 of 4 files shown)


xytosis:
@yschimke can this diff be landed? It would greatly help Ondrej with the CPP stuff, we've already found a few bugs so far, and this can possibly help debug initial payload issues.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
```diff
@@ -38,7 +38,7 @@ public void run(String realfile, String host, int port, boolean debug2, List<Str
             throws MalformedURLException, URISyntaxException {
         debug = debug2;
         // we pass in our reactive socket here to the test suite
-        String file = "reactivesocket-tck-drivers/src/test/resources/client$.txt";
+        String file = "reactivesocket-tck-drivers/src/test/resources/client.txt";
```
yschimke: This path looks wrong 

https://github.com/ReactiveSocket/reactivesocket-java/tree/master/reactivesocket-tck-drivers/src/test/resources

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/server/JavaTCPServer.java
```diff
@@ -30,7 +30,7 @@ public JavaTCPServer() {
 
     public void run(String realfile, int port) {
 
-        String file = "reactivesocket-tck-drivers/src/main/resources/server$.txt";
+        String file = "reactivesocket-tck-drivers/src/main/resources/server.txt";
```
yschimke: as above

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/server/JavaServerDriver.java
```diff
@@ -145,7 +145,10 @@ public RequestHandler parse() {
                 ParseMarble pm = new ParseMarble(marble, s);
                 s.onSubscribe(new TestSubscription(pm));
                 new ParseThread(pm).start();
-            }
+            } else {
+                ConsoleUtils.failure("Request response payload " + initialPayload.getK() + " " + initialPayload.getV()
```
yschimke: It isn't very obvious that this will cause the test to fail.  The javadoc says "Logs a failure event".  But more importantly it's a hidden boolean.

This isn't a new issue but I missed it until now.

xytosis: My bad, I should have been more clear in the documentation.

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/client/JavaTCPClient.java
xytosis: Sorry, I had changed the TCK to remove the dollar sign, but I will keep things consistent with what we have here.

PR #178, opened 2016-10-24
[0.5.x] Setup payload

NiteshKant:
#### Problem

`SetupProvider` currently does not provide a way to specify data and metadata for the setup frame.

#### Modification

Add a method `setupPayload(Payload)` to support optionally specifying a payload for the setup frame. Default is empty still metadata and data.

#### Result

Now data or metadata can be specified for a setup from the client.


NiteshKant:

PR #179, opened 2016-10-25
Client is not receiving leases

NiteshKant:
#### Problem

`ClientServerInputMultiplexer` is not sending any stream 0 frames to `ClientReactiveSocket`. Since, leases come on stream 0, clients were not receiving the leases.

#### Modification

Modified `ClientServerInputMultiplexer` to send stream 0 frames to both sockets (since both ends receive frames on stream 0) and let them decide which ones are useful for them.

Also modified `FairLeaseDistributor` to subscribe to ticks after the first socket is registered. This provides predictable behavior w.r.t the first lease distribution i.e. if the lease ticks start on subscription, generally the first lease is missed by the first socket.

#### Result

Clients now receive leases.


NiteshKant:
\cc @robertroeser

NiteshKant:

PR #180, opened 2016-10-26
UnitTest for local transport explicit connection close.

NiteshKant:
#### Problem

No tests to verify whether close works for local transport.

#### Modification

Added a test to close connection.

#### Result

More test coverage.


NiteshKant:

PR #183, opened 2016-10-31
ReactiveSocket frame logger (tcp)

NiteshKant:
#### Problem

Tracking ReactiveSocket frames is not easy with wire logging as it logs the bytes written/read.

#### Modification

Added a netty handler to be added as the first handler in the pipeline and logs frame objects as-is written and read on the channel.
Following netty's logging handler design, this handler can be configured to log at a particular log level, which can be changed by the user at runtime.

#### Result

Better tracking of ReactiveSocket frames on the channel.


ReactiveSocketAdmin:
👍

stevegury:
👍

PR #184, opened 2016-11-02
Microbenchmarks for tcp and local

NiteshKant:
#### Problem

No benchmarks for transports.

#### Modification

Added a benchmark to compare request-response across multiple transports. Added local and tcp for now.

#### Result

Benchmark to give some idea about how different changes impact the perf.


stevegury:
👍

PR #185, opened 2016-11-02
`RemoteReceiver` handle events before subscription.

NiteshKant:
#### Problem

`RemoteReceiver` is used for receiving a stream over a `ReactiveSocket`, eg: `requestChannel()` request or `requestStream()` response.
For request stream, it may so happen that the receiver does not subscribe to the request stream. This is OK for `onNext`s but terminal events can flow in without explicit `requestN`s. `RemoteReceiver` does not handle such terminal events today.

#### Modification

Added `missedCompletion` and `missedError` to `RemoteReceiver` that correctly buffers the terminal events if there is no subscription and sends these events if and when a subscriber arrives.

#### Result

No errors when terminal events flow without subscription.

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/internal/RemoteReceiver.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/internal/RemoteReceiver.java
@@ -53,7 +53,6 @@ import org.reactivestreams.Subscription;
  */
 public final class RemoteReceiver implements Processor<Frame, Payload> {
 
-    private final Publisher<Frame> transportSource;
     private final DuplexConnection connection;
     private final int streamId;
     private final Runnable cleanup;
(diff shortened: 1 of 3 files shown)


reactivesocket-core/src/main/java/io/reactivesocket/internal/RemoteReceiver.java
```diff
@@ -176,6 +186,16 @@ public void cancel() {
         onError(new CancelException("Remote subscription cancelled."));
     }
 
+    private boolean isMissed() {
```
stevegury: why not just returning `subscription == null` ?
I don't see the need for the synchronized block.

NiteshKant: Hmm, it's actually a bug introduced by my refactoring of `isMissed()` logic into a method.

There is a race between concurrent `subscribe()` and `on*()` setting `missed*` variables. These operations have to be atomic i.e. `subscription == null` and `missed*` variable initialization as `subscribe()` depends on it for properly invoking `subscriber`:

```java
synchronized(this) {
...
_missed = missedComplete || null != missedError;
...
}
if (_missed) {
...
``` 

Thanks!

NiteshKant:
@stevegury Addressed your comment, can you review this again?

PR #186, opened 2016-11-02
Fix for `requestChannel` and `requestStream`

NiteshKant:
__This contains #185, so do not merge this first, it is just a headsup for review__

#### Problem
In the current implementation stream and channel methods are broken.

#### Modification
Fixed stream and channel methods to work end to end. Also added examples for the same.
They are now implemented using the existing `RemoteReceiver` and `RemoteSender` classes.
One major problem with using `ConnectableUnicastProcessor` was that it does not discriminate between subscriber of request and sender of `requestN` and `cancel` frames. So, if the request source is completed, sending `requestN` and `cancel` frames result in them being rejected by the transport. This is one of the reasons `RemoteReceiver` and `RemoteSender` classes were created.

 __request-response implementation is not changed in this change as that needs to be optimized for a single item request-response__

#### Result
Channel and stream work now.


NiteshKant:
\cc @robertroeser @stevegury

PR #189, opened 2016-11-05
[0.5.x-events] Events interface

NiteshKant:
__Problem__

Insights into the internals of `ReactiveSocket` is not available for users. eg: Load balancing internals, leases sent/received, keep-alives sent/received, etc.

__Modification__

This change is first of a series of changes intended to address this problem. It only contains the core interfaces to support event publications from `ReactiveSocket` internals.

The model is adopted from the [tried and tested model in `RxNetty`](https://github.com/ReactiveX/RxNetty/issues/98) to provide an event publishing infrastructure that can be used to implement metrics of event firehose. One of the primary guideline of this model is to reduce the object allocations that comes along with an event stream based on emitting an event as an object. Instead, this approach embraces distinct callbacks that any listener can choose to implement if interested.

Primary classes are `EventListener` and `EventSource`. There are a few extensions of `EventListener` to provide events specific to a server, client or load balancer.

__Result__

Better insight into `ReactiveSocket` internals.

### DIFF
--- a/reactivesocket-client/src/main/java/io/reactivesocket/client/events/LoadBalancingClientListener.java
+++ b/reactivesocket-client/src/main/java/io/reactivesocket/client/events/LoadBalancingClientListener.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ * <p>
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ *  the License. You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


NiteshKant:
\cc @stevegury @rdegnan @robertroeser @yschimke @ktoso 

These changes are targetted to a temporary branch [0.5.x-events](https://github.com/ReactiveSocket/reactivesocket-java/tree/0.5.x-events) , feedback is appreciated

PR #195, opened 2016-11-16
JMH test for larger payloads

NiteshKant:
#### Problem

Current JMH tests only benchmark smaller payloads "hello". It will be good to know the numbers with larger payloads.

#### Modifications

- Refactored `RequestResponsePerf` and extracted a base class that can be used in all benchmarks.
- Added another benchmark class to test different payload sizes.

#### Result

More insight into perf.

### DIFF
--- a/reactivesocket-examples/src/jmh/java/io/reactivesocket/perf/RequestResponseLargePayloadPerf.java
+++ b/reactivesocket-examples/src/jmh/java/io/reactivesocket/perf/RequestResponseLargePayloadPerf.java
@@ -14,7 +14,6 @@
 package io.reactivesocket.perf;
 
 import io.reactivesocket.ReactiveSocket;
-import io.reactivesocket.perf.util.ClientServerHolder;
 import io.reactivesocket.util.PayloadImpl;
 import org.openjdk.jmh.annotations.Benchmark;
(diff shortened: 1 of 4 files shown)


reactivesocket-examples/src/jmh/java/io/reactivesocket/perf/RequestResponseLargePayloadPerf.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+@State(Scope.Benchmark)
+public class RequestResponseLargePayloadPerf extends AbstractReactiveSocketPerf {
+
+    @Param({"16", "1024"})
```
stevegury: You could also add the case of very small messages, i.e. a few bytes.

NiteshKant: Existing `RequestResponsePerf ` benchmark uses a small message `hello`.

PR #197, opened 2016-11-18
Additional flags in `Frame.toString()`

NiteshKant:
#### Problem

Currently `Frame.toString()` does not print the various flags associated with different frame types. eg: `requestN` value in `RequestN` frame. This limits the usefulness of frame logging.

#### Modifications

Added `additionalFlags` string which is generated per frame type.

#### Result

More insight for frame logging.

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/Frame.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/Frame.java
@@ -541,9 +541,11 @@ public class Frame implements Payload {
         FrameType type = FrameType.UNDEFINED;
         StringBuilder payload = new StringBuilder();
         long streamId = -1;
+        String additionalFlags = "";
 
         try {
             type = FrameHeaderFlyweight.frameType(directBuffer, 0);
+
             ByteBuffer byteBuffer;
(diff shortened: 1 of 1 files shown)


reactivesocket-core/src/main/java/io/reactivesocket/Frame.java
```diff
@@ -562,9 +564,37 @@ public String toString() {
             }
 
             streamId = FrameHeaderFlyweight.streamId(directBuffer, 0);
+
+            switch (type) {
+            case LEASE:
+                additionalFlags = " Permits: " + Lease.numberOfRequests(this) + ", TTL: " + Lease.ttl(this);
```
yschimke: This is a bit inconsistent as you don't use , as a separator in the normal toString values

NiteshKant: Aah I didn't notice that. I will fix it in the next change I do

NiteshKant: FYI #198 contains this formatting change.

PR #199, opened 2016-11-18
Fix to publish Snapshots to jfrog

NiteshKant:
#### Problem

Currently travis is publishing 0.2.3-SNAPSHOT artifacts for 0.5.x branch.

#### Modifications

I am not 100% sure if this is the correct way to make travis publish the correct version, but it is a way!
Modified `buildViaTravis.sh` to override the version, just for snapshot builds.

#### Result

Correct version number for snapshot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ repositories {
 }
 
 dependencies {
-    compile 'io.reactivesocket:reactivesocket:0.0.1-SNAPSHOT'
+    compile 'io.reactivesocket:reactivesocket:0.5.0-SNAPSHOT'
 }
 ```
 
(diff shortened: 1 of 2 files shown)


NiteshKant:
FYI @2m @ktoso @drewhk

2m:
So this fixes the version number. But does this also mean that snapshots are deployed after every commit to JFrog?

NiteshKant:
@2m the snapshots are built and published via travis today. The only issue is that it is publishing them with the wrong version. This change (hopefully) fixes the version.

NiteshKant:
FYI: @2m the snapshots are published to JFrog now with 0.5.0-SNAPSHOT name.

PR #200, opened 2016-12-05
added sliding window histogram

robertroeser:
Problem
Current histogram doesn't resets on every get on every dimension. This creates choppy metrics.

Modifications
Modified the timer to have n histrograms in a queue and rotate of over them to smooth out the histogram over a sliding window. Switched the dimensions of the histrogram to a label called value. Switched the reseting of the histrogram to a closure that is called by the timer instead so it will rotate the histrogram once per window.

Result
Correct metrics over a rolling window.

### DIFF
--- a/reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/HdrHistogramGauge.java
+++ b/reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/HdrHistogramGauge.java
@@ -21,37 +21,27 @@ import com.netflix.servo.monitor.MonitorConfig;
 import com.netflix.servo.monitor.NumberGauge;
 import org.HdrHistogram.Histogram;
 
-import java.util.concurrent.TimeUnit;
-
 /**
  * Gauge that wraps a {@link Histogram} and when it's polled returns a particular percentage
  */
 public class HdrHistogramGauge extends NumberGauge {
-    private static final long TIMEOUT = TimeUnit.MINUTES.toMillis(1);
-    private final Histogram histogram;
+    private final SlidingWindowHistogram histogram;
     private final double percentile;
-    private volatile long lastCleared = System.currentTimeMillis();
+    private final Runnable slide;
 
-    public HdrHistogramGauge(MonitorConfig monitorConfig, Histogram histogram, double percentile) {
+    public HdrHistogramGauge(MonitorConfig monitorConfig, SlidingWindowHistogram histogram, double percentile, Runnable slide) {
         super(monitorConfig);
         this.histogram = histogram;
         this.percentile = percentile;
+        this.slide = slide;
 
         DefaultMonitorRegistry.getInstance().register(this);
     }
 
     @Override
     public Long getValue() {
-        long value = histogram.getValueAtPercentile(percentile);
-        if (System.currentTimeMillis() - lastCleared > TIMEOUT) {
-            synchronized (histogram) {
-                if (System.currentTimeMillis() - lastCleared > TIMEOUT) {
-                    histogram.reset();
-                    lastCleared = System.currentTimeMillis();
-                }
-            }
-        }
-
+        long value = histogram.aggregateHistogram().getValueAtPercentile(percentile);
+        slide.run();
         return value;
     }
 }
--- a/reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/HdrHistogramServoTimer.java
+++ b/reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/HdrHistogramServoTimer.java
@@ -17,8 +17,6 @@ package io.reactivesocket.loadbalancer.servo.internal;
 
 import com.netflix.servo.monitor.MonitorConfig;
 import com.netflix.servo.tag.Tag;
-import org.HdrHistogram.ConcurrentHistogram;
-import org.HdrHistogram.Histogram;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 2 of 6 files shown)


reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/HdrHistogramServoTimer.java
```diff
@@ -46,31 +48,26 @@
...
-
-        p50 = new HdrHistogramGauge(MonitorConfig.builder(label + "_p50").withTags(tags).build(), histogram, 50);
-        p90 = new HdrHistogramGauge(MonitorConfig.builder(label + "_p90").withTags(tags).build(), histogram, 90);
-        p99 = new HdrHistogramGauge(MonitorConfig.builder(label + "_p99").withTags(tags).build(), histogram, 99);
-        p99_9 = new HdrHistogramGauge(MonitorConfig.builder(label + "_p99_9").withTags(tags).build(), histogram, 99.9);
-        p99_99 = new HdrHistogramGauge(MonitorConfig.builder(label + "_p99_99").withTags(tags).build(), histogram, 99.99);
+        min = new HdrHistogramMinGauge(MonitorConfig.builder(label).withTag("value", "min").withTags(tags).build(), histogram);
+        max = new HdrHistogramMaxGauge(MonitorConfig.builder(label).withTag("value", "min").withTags(tags).build(), histogram);
+
+        p50 = new HdrHistogramGauge(MonitorConfig.builder(label).withTag("value", "p50").withTags(tags).build(), histogram, 50, this::slide);
```
NiteshKant: This will rotate the `histogram` for each slide of `min`, `max`, etc. Is that what you intend?

reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/HdrHistogramGauge.java
```diff
@@ -21,37 +21,27 @@
...
-            synchronized (histogram) {
-                if (System.currentTimeMillis() - lastCleared > TIMEOUT) {
-                    histogram.reset();
-                    lastCleared = System.currentTimeMillis();
-                }
-            }
-        }
-
+        long value = histogram.aggregateHistogram().getValueAtPercentile(percentile);
+        slide.run();
```
NiteshKant: This could lead to unintentional sliding of the window. eg: If a user action fetches the value (eg: from a JMX console), it will reset the histogram.

reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/SlidingWindowHistogram.java
```diff
@@ -0,0 +1,107 @@
...
+        histogramAtomicReference.get().recordValue(value);
+    }
+
+    /**
+     * Slides the Histogram window. Pops a Histogram off a queue, resets it, and places the old
+     * on in the queue.
+     */
+    public void rotateHistogram() {
+        histogramAtomicReference
+            .updateAndGet(old -> {
```
NiteshKant: Since, the passed function to `updateAndGet` should be side-effect free (as per the javadoc), I don't think this is a good usage of `updateAndGet`. 
Maybe this should be synchronized to disallow concurrent modification and simply rotate the histograms atomically.

reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/SlidingWindowHistogram.java
```diff
@@ -0,0 +1,107 @@
...
+
+            });
+    }
+
+    /**
+     * Aggregates the Histograms into a single histogramAtomicReference and returns it.
+     *
+     * @return Aggregated histogramAtomicReference
+     */
+    public Histogram aggregateHistogram() {
```
NiteshKant: Since gets will most likely not be concurrent, done by the poller, do we need threadlocal isolation of aggregation? Synchronized aggregation may be simpler and efficient? Also, you are already synchronizing additions to the aggregate.

reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/HdrHistogramServoTimer.java
robertroeser: get min and max don't reset the histogram

reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/SlidingWindowHistogram.java
robertroeser: No it's be bad because sync would block the aggregations. One line 86 it adds the live one, and then in the sync block it adds the ones in the queue. 
I can replace the thread local with a new object.

robertroeser: No it's be bad because sync would block the aggregations. One line 86 it adds the live one, and then in the sync block it adds the ones in the queue.
I can replace the thread local with a new object.

reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/SlidingWindowHistogram.java
robertroeser: yeah i can do that

reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/HdrHistogramGauge.java
NiteshKant: Yes but do you intend to slide (and in the other class, today it rotates the histogram) on every `get` ?

reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/SlidingWindowHistogram.java
NiteshKant: Yes the sync will block aggregations but one histogram will be used by one metric which will have one poller and hence there would only be a single thread calling `aggregate` isn't?

reactivesocket-stats-servo/src/main/java/io/reactivesocket/loadbalancer/servo/internal/HdrHistogramGauge.java
robertroeser: yes

NiteshKant: My bad, didn't see the timeout condition, thanks!

PR #201, opened 2016-12-06
memory leak

robertroeser:
Problem
There is a race conditon that causes a SocketAdder to not be removed from the of available sockets if it is removed from the list of available servers. This has the side-effect of creating a memory leak because it will continually try to connect, but fail. 

Modifications
After a SocketAdder errors out five times it will no long add itself to the list of active factories.

Result
No more memory leak, and after 5 errors the SocketAdder is removed from the list.


stevegury:
I think this is a work around the memory leak, it should be ok to infinitely try to connect on an invalid ip address.

NiteshKant:
Probably a slightly better way is to check for factory availability. If it isn't `> 0.0` then do not add it back to active factories.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -632,7 +634,11 @@ public void onError(Throwable t) {
             logger.warn("Exception while subscribing to the ReactiveSocket source", t);
             synchronized (LoadBalancer.this) {
                 pendingSockets -= 1;
-                activeFactories.add(factory);
+                if (++errors < 5) {
+                    activeFactories.add(factory);
+                } else {
+                    logger.warn("Exception count greater than 5, not re-adding factory {}", factory.toString());
```
yschimke: nit: greater than 4?

stevegury:
@NiteshKant the loadbalancer (most of the time) won't select the factory if the availability is 0, it will only do it if there are no other available factories.

The bug is that there's a period (while we're connecting to a host), when the factory is not in `activeFactories` and not yet in `activeSockets`. If the server is removed from discovery at that time, the factory will never be removed. This may lead to unbounded attempts to connect to that host.

I agree that this is a bug, but disagree that it is the cause of the memory leak.

NiteshKant:
@stevegury thanks for the explanation of the bug!

I was just wondering whether that is a better condition than checking for 5 magical error number :)
Since, one would decorate the factory with `ReactiveSocketClients.detectFailure()` hence changing availability, it also makes it configurable.

Anyway, if this change looks good to you, please approve!

robertroeser:
@NiteshKant I think we should merge this, and add an issue to address this properly.

NiteshKant:
@robertroeser Agreed.

stevegury:
I think the work around is acceptable for now.
👍

PR #202, opened 2016-12-07
Empty stacktrace of `TimeoutException`

NiteshKant:
#### Problem

`TimeoutPublisher` creates a new `TimeoutException` instance for every timeout. This is costly when the system is under load and has a lot of timers that timeout, which is typical when handling latent outbound services.

#### Modification

Used a cached `TimeoutException` instance and empty the stacktrace of the exception.

#### Result

Less overhead of timeouts.


reactivesocket-publishers/src/main/java/io/reactivesocket/reactivestreams/extensions/internal/publishers/TimeoutPublisher.java
```diff
@@ -28,6 +28,15 @@
 
 public final class TimeoutPublisher<T> implements Px<T> {
 
+    @SuppressWarnings("ThrowableInstanceNeverThrown")
+    private static final TimeoutException timeoutException = new TimeoutException();
+
+    private static final StackTraceElement[] EMPTY_STACK = new StackTraceElement[0];
```
stevegury: Rather than setting the stack trace manually, it would be better to override the `fillInStacktrace` method in all reactivesocket exceptions.

```@Override
public synchronized Throwable fillInStackTrace() {
    return this;
}
```

NiteshKant: Makes sense. I will update.
BTW, this code is in publishers package and `TimeoutException` is in core, so this code here uses `java.util.concurrent.TimeoutException`

PR #204, opened 2016-12-08
Refactored `StressTest` for easier configuration

NiteshKant:
#### Problem

`StressTest` is pretty useful to do blackbox testing of ReactiveSocket but it is hard to add behavior like leases, keep alive, timeouts, etc.

#### Modification

Refactored and did the following changes:

- Introduced `TestConfig` that is all that needs to be altered to do behavior changes like lease, timeouts, etc.
- Changed `StressTest` to use `TestConfig` and only contain the core logic of the test.
- Introduced a `StressTestDriver` that contains the creation of config and invocation of `StressTest`

#### Result

More powerful `StressTest` which can be used for a variety of situations.


stevegury:
👍

PR #209, opened 2016-12-15
Reusable `Payload`

NiteshKant:
__Problem__

`PayloadImpl` caches `data` and `metadata` buffers. This means once those buffers are read, they can not be reused as the pointer of `ByteBuffer` has moved to the end of buffer.
This restricts the usage of `PayloadImpl` instance with something like:

```java
socket.requestResponse(new PayloadImpl("Hello")).retry(2);
```

For any retry in the above code, data read from the payload instance will be empty.

__Modification__

- store buffer `position()` on creation and reset them on every `get()`.
- Additional constructor to override this behavior and create a payload for single use.
- `PayloadImpl` defaults to reusable payload.

__Result__

Possible to retry a request without having a custom `Payload` implementation.

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/util/PayloadImpl.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/util/PayloadImpl.java
@@ -23,13 +23,23 @@ import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 
 /**
- * An implementation of {@link Payload}
+ * An implementation of {@link Payload}. This implementation is <b>not</b> thread-safe, and hence any method can not be
+ * invoked concurrently.
+ *
+ * <h2>Reusability</h2>
+ *
+ * By default, an instance is reusable, i.e. everytime {@link #getData()} or {@link #getMetadata()} is invoked, the
+ * position of the returned buffer is reset to the start of data in the buffer. For creating an instance for single-use,
+ * {@link #PayloadImpl(ByteBuffer, ByteBuffer, boolean)} must be used.
  */
 public class PayloadImpl implements Payload {
 
     public static final Payload EMPTY = new PayloadImpl(Frame.NULL_BYTEBUFFER, Frame.NULL_BYTEBUFFER);
 
     private final ByteBuffer data;
+    private final int dataStartPosition;
+    private final int metadataStartPosition;
+    private final boolean reusable;
     private final ByteBuffer metadata;
 
     public PayloadImpl(String data) {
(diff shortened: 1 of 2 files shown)


reactivesocket-core/src/main/java/io/reactivesocket/util/PayloadImpl.java
```diff
@@ -61,17 +71,39 @@ public PayloadImpl(ByteBuffer data) {
...
+     * New instance where every invocation of {@link #getMetadata()} and {@link #getData()} will reset the position of
+     * the buffer to the position when it is created, if and only if, {@code singleUse} is set to {@code false}.
+     *
+     * @param data Buffer for data.
+     * @param metadata Buffer for metadata.
+     * @param singleUse {@code true} if the buffer position is to be reset on every invocation of {@link #getData()} and
+     * {@link #getMetadata()}.
+     */
+    public PayloadImpl(ByteBuffer data, ByteBuffer metadata, boolean singleUse) {
         this.data = data;
```
NiteshKant: @robertroeser `ByteBuffer.slice()` always uses the current position i.e. there is no such method as `ByteBuffer.slice(position)`. So, if we were to slice, we still have to reset the position first.
Do you see any other value in using `slice()` even though we have to reset position first?

reactivesocket-core/src/main/java/io/reactivesocket/util/PayloadImpl.java
```diff
@@ -61,17 +71,39 @@ public PayloadImpl(ByteBuffer data) {
...
+    /**
+     * New instance where every invocation of {@link #getMetadata()} and {@link #getData()} will reset the position of
+     * the buffer to the position when it is created, if and only if, {@code singleUse} is set to {@code false}.
+     *
+     * @param data Buffer for data.
+     * @param metadata Buffer for metadata.
+     * @param singleUse {@code true} if the buffer position is to be reset on every invocation of {@link #getData()} and
+     * {@link #getMetadata()}.
+     */
+    public PayloadImpl(ByteBuffer data, ByteBuffer metadata, boolean singleUse) {
```
stevegury: nitpick: but renaming/reversing `singleUse` to `reusable` may be easier to understand (especially if you don't read the doc).

reactivesocket-core/src/main/java/io/reactivesocket/util/PayloadImpl.java
NiteshKant: ping @robertroeser

ktoso:
Ping, any chance to get this in soon? :)

NiteshKant:
@ktoso you got it 😄

reactivesocket-core/src/main/java/io/reactivesocket/util/PayloadImpl.java
NiteshKant: @stevegury done, renamed to `reusable` and changed implementation!

ktoso:
Excellent, thanks! Very nice to see much movement here in the new year 👍

Issue #212, opened 2016-12-28
[0.5.x] Questions about StreamIdSupplier


yschimke:
1. StreamIdSupplier.clientSupplier().nextStreamId() -> 3

Why not start with 1?

2. isValid seems flipped

3. Should isValid consider odds/even? clientSupplier and serverSupplier may be vastly out of sync.

```
    public static void main(String[] args) {
        StreamIdSupplier s = StreamIdSupplier.clientSupplier();

        System.out.println("1 " + s.isValid(1));

        System.out.println("next " + s.nextStreamId());
        System.out.println("next " + s.nextStreamId());

        System.out.println("3 " + s.isValid(3));
        System.out.println("4 " + s.isValid(4));
        System.out.println("5 " + s.isValid(5));
        System.out.println("6 " + s.isValid(6));
        System.out.println("7 " + s.isValid(7));
    }
```

Output

```
1 false
next 3
next 5
3 false
4 false
5 false
6 true
7 true
```

yschimke:
Used here, which implies that isValid returns whether the stream id was previously issued

```
    private void handleMissingResponseProcessor(int streamId, FrameType type, Frame frame) {
        if (!streamIdSupplier.isValid(streamId)) {
            if (type == FrameType.ERROR) {
                // message for stream that has never existed, we have a problem with
                // the overall connection and must tear down
                String errorMessage = getByteBufferAsString(frame.getData());

                throw new IllegalStateException("Client received error for non-existent stream: "
                    + streamId + " Message: " + errorMessage);
            } else {
                throw new IllegalStateException("Client received message for non-existent stream: " + streamId +
                                                ", frame type: " + type);
            }
        }
        // receiving a frame after a given stream has been cancelled/completed,
        // so ignore (cancellation is async so there is a race condition)
    }
```

yschimke:
https://github.com/ReactiveSocket/reactivesocket-java/pull/213

NiteshKant:
@yschimke yeah the intent of `isValid` is to be used in the code you pointed to. Perhaps it is badly named method.

>>> Should isValid consider odds/even? clientSupplier and serverSupplier may be vastly out of sync.

Since this is checking for validity only for received frames and the even/odd filtering is done beforehand by `ClientServerInputMultiplexer`, I do not think it is required to check for even/odd here. This class should also be pkg-private.

NiteshKant:
@yschimke can I close this now as #213 is merged?

PR #213, opened 2016-12-28
StreamIdSupplier returns 1,3,5... or 2,4,6...

yschimke:


NiteshKant:
@yschimke I replied with my thoughts on issue #212 

I think semantically we should stick to `isValid()` doing what it does today but initializing the supplier with `-1` and `0` makes sense. Can you revert the `isValid()` change?

Also, I think this class should be pkg-private and `isValid()` should perhaps be named `isBeforeCurrent()` (or some better name)? If you want you can make those changes too 😄

yschimke:
So I don't think I should fully revert the isValid changes, since I still think its flipped logically.  For an unknown stream id, I still think you only want to log when the stream id is negative or greater than the last issued.

Also the class is used outside the package so I can't make the other change without refactoring.

NiteshKant:
>> Also the class is used outside the package so I can't make the other change without refactoring.

Duh, I didn't notice the outside package usages, sorry about that.

Thanks @yschimke !

PR #216, opened 2017-01-04
Duplex interactions

NiteshKant:
#### Problem

Duplex interactions (server sending requests to the client) are broken.

#### Modification

- `ReactiveSocketServer` wasn't using `ClientServerInputMultiplexer` to split input into client and server sockets. This made the responses sent from the client to not reach `ClientReactiveSocket`. Modified `ReactiveSocketServer` to use `ClientServerInputMultiplexer`.
- `ClientReactiveSocket` didn't handle a request-stream/channel request before transport subscription arrived. Now using a buffering subscription to buffer request-n and cancel till transport subscription arrives.
- Added `DuplexClient` example to demonstrate duplex interactions.

#### Result

Duplex interactions works!


NiteshKant:
\cc @benjchristensen

reactivesocket-core/src/main/java/io/reactivesocket/server/ReactiveSocketServer.java
```diff
@@ -42,24 +45,28 @@
...
                   .switchTo(setupFrame -> {
                       if (setupFrame.getType() == FrameType.SETUP) {
+                          ClientServerInputMultiplexer multiplexer = new ClientServerInputMultiplexer(connection);
                           ConnectionSetupPayload setupPayload = ConnectionSetupPayload.create(setupFrame);
-                          ClientReactiveSocket sender = new ClientReactiveSocket(duplexConnection,
+                          ClientReactiveSocket sender = new ClientReactiveSocket(multiplexer.asServerConnection(),
                                                                                  Throwable::printStackTrace,
                                                                                  StreamIdSupplier.serverSupplier(),
                                                                                  KeepAliveProvider.never());
+                          LeaseHonoringSocket lhs = new DefaultLeaseHonoringSocket(sender);
```
NiteshKant: The oddity here is that there is no way today in the protocol where a server can tell whether it will honor leases from the client. Probably this should piggy back on the client will honor lease in the setup saying if a client honors lease, server will too.
I am interested in hearing what others think about this.

benjchristensen: @tmontgomery Your input would be useful on this I imagine.

benjchristensen: Does it even make sense for a client responder to use leases? Either it can take requests or it can't (reject/fail). For example, a server-mobile relationship doesn't make sense for leases, as once a mobile client is connected, it only has that one connection to send requests over. Leases really only make sense in server-server situations where a requesting server has multiple connections open and can send requests to different servers depending on their leases.

stevegury: I think it's an oversight.
We could decide that the server will match the configuration of the client.

stevegury:
👍

PR #217, opened 2017-01-04
Server and Client event publishing

NiteshKant:
__Problem__

No events are published for `ReactiveSocketClient` and `ReactiveSocketServer`

__Modification__

Added event publishing for `ReactiveSocketClient` and `ReactiveSocketServer`

__Result__

Events for clients and servers.

PS: This does not yet add events for `LoadBalancer`. I will send a separate PR for that.

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/client/AbstractReactiveSocketClient.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/client/AbstractReactiveSocketClient.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ * <p>
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ *  the License. You may obtain a copy of the License at
+ *  <p>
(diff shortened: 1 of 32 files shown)


NiteshKant:
FYI: Since the change was small for the server, I added that to this PR too.

reactivesocket-core/src/main/java/io/reactivesocket/client/AbstractReactiveSocketClient.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2016 Netflix, Inc.
```
stevegury: 2017 ;-)

NiteshKant: Happy new year :)

Looks like a copy-paste otherwise my template has year parameterized.

NiteshKant:
>>> But I would also like to experiment with the tracing pattern which is less dynamic than your approach but more lightweight, this would allow us to instrument even performance critical piece of code and only pay the price of this when we listen to those specific events.

@stevegury yeah I would be interested in seeing this done differently!

PR #218, opened 2017-01-09
Event publishing for `LoadBalancer`

NiteshKant:
__Problem__

No events are published for `LoadBalancer`

__Modification__

Publishing events for `LoadBalancer`

__Result__

More events, more insight!

### DIFF
--- a/reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
+++ b/reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
@@ -15,11 +15,17 @@
  */
 package io.reactivesocket.client;
 
+import io.reactivesocket.Availability;
 import io.reactivesocket.Payload;
 import io.reactivesocket.ReactiveSocket;
+import io.reactivesocket.client.events.LoadBalancingClientListener;
+import io.reactivesocket.events.ClientEventListener;
+import io.reactivesocket.events.EventSource;
 import io.reactivesocket.exceptions.NoAvailableReactiveSocketException;
 import io.reactivesocket.exceptions.TimeoutException;
 import io.reactivesocket.exceptions.TransportException;
+import io.reactivesocket.internal.DisabledEventPublisher;
+import io.reactivesocket.internal.EventPublisher;
 import io.reactivesocket.reactivestreams.extensions.Px;
 import io.reactivesocket.reactivestreams.extensions.internal.EmptySubject;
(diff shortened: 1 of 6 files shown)


reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -313,38 +335,44 @@ private void updateAperture(int newValue, long now) {
...
+            long prev = refreshPeriod;
+            refreshPeriod = (long) Math.min(refreshPeriod * 1.5, maxRefreshPeriod);
+            logger.debug("Bumping refresh period, {}->{}", prev / 1000, refreshPeriod / 1000);
+            if (prev != refreshPeriod && eventListener != null) {
+                eventListener.socketRefreshPeriodChanged(prev, refreshPeriod, Clock.unit());
+            }
+            lastRefresh = now;
+            addSockets(1);
+            if (eventListener != null) {
+                eventListener.socketsRefreshCompleted(Clock.elapsedSince(now), Clock.unit());
```
stevegury: I think this would be better moved inside `addSocket`, since the `addSocket` action is asynchronous.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
```diff
@@ -204,17 +223,17 @@ public LoadBalancer(Publisher<? extends Collection<ReactiveSocketClient>> factor
 
     private synchronized void addSockets(int numberOfNewSocket) {
         int n = numberOfNewSocket;
-        if (n > activeFactories.size()) {
-            n = activeFactories.size();
+        if (n > activeFactories.holder.size()) {
```
stevegury: Why not adding a `size` method in ActiveList?

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
NiteshKant: (Discussed offline)
It is more involved to emit this event after the `addSocket` is finished. This event is more useful to indicate that socket refresh is happening, latency of connect can be useful in determining latency of refresh.

reactivesocket-client/src/main/java/io/reactivesocket/client/LoadBalancer.java
NiteshKant: Done.

NiteshKant:
@sgury I have addressed your comments.

PR #223, opened 2017-01-10
fails on requestSubscription calls

yschimke:
We haven't yet killed off requestSubscription and the Frame type constants still include it, so I'd expect this to work for now.

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/events/EventListener.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/events/EventListener.java
@@ -33,6 +33,7 @@ public interface EventListener {
     enum RequestType {
         RequestResponse,
         RequestStream,
+        RequestSubscription,
         RequestChannel,
         MetadataPush,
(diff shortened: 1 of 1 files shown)


NiteshKant:
@yschimke removing request-subscription is next on my plate after events are done. Since, removing it is really short term, I did not want to add and remove.

yschimke:
@NiteshKant can we temporarily land this and then remove subscription completely? reactivesocket-java in an awkward middle state where the one main operation we use is mostly supported (api + plumbing) but fails because of event logging.

NiteshKant:
@yschimke didn't know that you are using request-subscription.
Merging now, does this fail even if there are no listeners attached?
Most likely, events will be broken for request-subscription (or may be emitted as request-stream, not sure), is that ok?

yschimke:
If I get issues, I'll submit small patches.  Anything bigger is probably good indication to bail and give up for now.

n.b. this is mainly for integration testing with reactivesocket-cpp.

PR #231, opened 2017-02-17
Update SETUP frame to 1.0 spec

lexs:
Moves streamId to front
Makes frame type 6 bits and flags 10 bits
Make all used flag 10 bits

Unit test pass, also added a minimal test for the frame type/flags


lexs:
Turns out I missed updating to the new frame length size. Closing for now until I fix that.

PR #233, opened 2017-02-20
Update frame header to 1.0 spec

lexs:
Makes frame and metadata length 24 bits.
Metadata length also excludes the length field itself
Moves streamId to front
Makes frame type 6 bits and flags 10 bits
Make all used flag 10 bits

Some new unit tests to verify all this.

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/frame/FrameHeaderFlyweight.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/frame/FrameHeaderFlyweight.java
@@ -187,8 +187,7 @@ public class FrameHeaderFlyweight {
         FrameType result = FrameType.from(typeAndFlags >> FRAME_TYPE_SHIFT);
 
         if (FrameType.RESPONSE == result) {
-            // FIXME
-            final int flags = flags(directBuffer, offset);
+            final int flags = typeAndFlags & FRAME_FLAGS_MASK;
 
             boolean complete = FLAGS_RESPONSE_C == (flags & FLAGS_RESPONSE_C);
             if (complete) {
--- a/reactivesocket-core/src/main/java/io/reactivesocket/frame/FrameHeaderFlyweight.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/frame/FrameHeaderFlyweight.java
@@ -43,33 +43,39 @@ public class FrameHeaderFlyweight {
 
     private static final boolean INCLUDE_FRAME_LENGTH = true;
 
+    private static final int FRAME_TYPE_BITS = 6;
+    private static final int FRAME_TYPE_SHIFT = 16 - FRAME_TYPE_BITS;
+    private static final int FRAME_FLAGS_MASK = 0b0000_0011_1111_1111;
+
+    public static final int FRAME_LENGTH_SIZE = 3;
+    public static final int FRAME_LENGTH_MASK = 0xFFFFFF;
+
     private static final int FRAME_LENGTH_FIELD_OFFSET;
-    private static final int TYPE_FIELD_OFFSET;
-    private static final int FLAGS_FIELD_OFFSET;
+    private static final int FRAME_TYPE_AND_FLAGS_FIELD_OFFSET;
     private static final int STREAM_ID_FIELD_OFFSET;
     private static final int PAYLOAD_OFFSET;
 
-    public static final int FLAGS_I = 0b1000_0000_0000_0000;
-    public static final int FLAGS_M = 0b0100_0000_0000_0000;
+    public static final int FLAGS_I = 0b10_0000_0000;
+    public static final int FLAGS_M = 0b01_0000_0000;
 
-    public static final int FLAGS_KEEPALIVE_R = 0b0010_0000_0000_0000;
+    // TODO(lexs): These are frame specific and should not live here
+    public static final int FLAGS_KEEPALIVE_R = 0b00_1000_0000;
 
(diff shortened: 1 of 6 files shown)


reactivesocket-core/src/main/java/io/reactivesocket/frame/FrameHeaderFlyweight.java
```diff
@@ -79,39 +85,40 @@ public static int computeFrameHeaderLength(final FrameType frameType, int metada
...
-        final FrameType frameType,
-        final int streamId
+            final MutableDirectBuffer mutableDirectBuffer,
+            final int offset,
+            final int frameLength,
+            final int flags,
+            final FrameType frameType,
+            final int streamId
     ) {
         if (INCLUDE_FRAME_LENGTH) {
```
yschimke: should INCLUDE_FRAME_LENGTH be a param as well? or leaving for later deliberately?

reactivesocket-core/src/main/java/io/reactivesocket/frame/FrameHeaderFlyweight.java
```diff
@@ -171,13 +178,16 @@ public static int encode(
...
+        return typeAndFlags & FRAME_FLAGS_MASK;
     }
 
     public static FrameType frameType(final DirectBuffer directBuffer, final int offset) {
-        FrameType result = FrameType.from(directBuffer.getShort(offset + TYPE_FIELD_OFFSET, ByteOrder.BIG_ENDIAN));
+        int typeAndFlags = directBuffer.getShort(offset + FRAME_TYPE_AND_FLAGS_FIELD_OFFSET, ByteOrder.BIG_ENDIAN);
+        FrameType result = FrameType.from(typeAndFlags >> FRAME_TYPE_SHIFT);
 
         if (FrameType.RESPONSE == result) {
+            // FIXME
```
yschimke: ?

reactivesocket-core/src/main/java/io/reactivesocket/frame/FrameHeaderFlyweight.java
```diff
@@ -219,40 +229,77 @@ public static ByteBuffer sliceFrameMetadata(final DirectBuffer directBuffer, fin
...
+        return length == 0 ? 0 : length + FRAME_LENGTH_SIZE;
+    }
+
+    private static void encodeLength(
+            final MutableDirectBuffer mutableDirectBuffer,
+            final int offset,
+            final int length
+    ) {
+        if ((length & ~FRAME_LENGTH_MASK) != 0) {
+            throw new IllegalArgumentException("Length is larger than 24 bits");
```
yschimke: or negative?

reactivesocket-core/src/main/java/io/reactivesocket/frame/FrameHeaderFlyweight.java
lexs: Leaving for later, I don't plan to make this better. It seems like a larger task to move this out of framing altogether.

reactivesocket-core/src/test/java/io/reactivesocket/frame/SetupFrameFlyweightTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import io.reactivesocket.FrameType;
+import org.agrona.concurrent.UnsafeBuffer;
+import org.junit.Test;
+
+import java.nio.ByteBuffer;
+
+import static org.junit.Assert.*;
+
+public class SetupFrameFlyweightTest {
```
yschimke: TIOLI: I'd find it really helpful to have at least one hexstring example in the test.

reactivesocket-core/src/main/java/io/reactivesocket/frame/FrameHeaderFlyweight.java
benjchristensen: It should be a signed 24-bit integer, which means 0-16,777,215 as the valid range.

benjchristensen: Yes, the error message could say it's either larger, or negative, since a negative number would fail this test (as it should).

reactivesocket-core/src/test/java/io/reactivesocket/frame/FrameHeaderFlyweightTest.java
```diff
@@ -0,0 +1,102 @@
...
+import io.reactivesocket.FrameType;
+import org.agrona.concurrent.UnsafeBuffer;
+import org.junit.Test;
+
+import java.nio.ByteBuffer;
+
+import static io.reactivesocket.frame.FrameHeaderFlyweight.NULL_BYTEBUFFER;
+import static org.junit.Assert.*;
+
+public class FrameHeaderFlyweightTest {
```
benjchristensen: 👍 for adding these unit tests

benjchristensen:
All looks good to me (as well as I can in my head do all the bit stuff). Unit tests seem valid, and all other unit tests are passing. So merging ...

reactivesocket-core/src/main/java/io/reactivesocket/frame/FrameHeaderFlyweight.java
lexs: Wait, this is a bit unclear, currently this is validating the full 24 bit range, effectively treating it as an unsigned 24 integer.

16,777,215 is the max of a imaginary 24 bit unsigned int.

PR #235, opened 2017-02-21
Properly handle SETUP frames with RESUME_ENABLED

lexs:
Properly parse SETUP frames with resumption but no support internally besides that.


lexs:
I tried writing a integration test to verify that we can reject the SETUP frame but that whole functionality seems to be broken, see https://github.com/ReactiveSocket/reactivesocket-java/issues/234

PR #238, opened 2017-02-22
Make request N non-optional for REQUEST_CHANNEL

lexs:


reactivesocket-core/src/main/java/io/reactivesocket/frame/RequestFrameFlyweight.java
```diff
@@ -77,6 +74,9 @@ public static int encode(
         final ByteBuffer metadata,
         final ByteBuffer data
     ) {
+        if (type == FrameType.REQUEST_CHANNEL) {
+            throw new AssertionError("REQUEST_CHANNEL most have initial request N");
```
yschimke: typo: must

reactivesocket-core/src/main/java/io/reactivesocket/frame/RequestFrameFlyweight.java
```diff
@@ -77,6 +74,9 @@ public static int encode(
         final ByteBuffer metadata,
         final ByteBuffer data
     ) {
+        if (type == FrameType.REQUEST_CHANNEL) {
```
yschimke: Should this only throw based on the flags?

PR #245, opened 2017-02-24
Add test to verify wire format

lexs:
This test writes the expected wire format and then verifies that we encode a frame like that.
I've verified by hand that the generated format is correct:
000000000000000000001001 // size (24 bits) = 9
00000000000000000000000000000101 // stream id (32 bits) = 5
001010 // frame type (6 bits) = 0x0A
0001000000 // flags (6 bits) == COMPLETE


reactivesocket-core/src/test/java/io/reactivesocket/frame/FrameHeaderFlyweightTest.java
```diff
@@ -99,4 +103,40 @@ public void typeAndFlagTruncated() {
...
+        FrameHeaderFlyweight.encode(directBuffer, 0, 5, flags, frameType, NULL_BYTEBUFFER, NULL_BYTEBUFFER);
+
+        ByteBuffer expected = ByteBufferUtil.preservingSlice(expectedMutable.byteBuffer(), 0, currentIndex);
+        ByteBuffer actual = ByteBufferUtil.preservingSlice(directBuffer.byteBuffer(), 0, FRAME_HEADER_LENGTH);
+
+        assertEquals(bytesToHex(expected), bytesToHex(actual));
+
+    }
+
+    private static final char[] hexArray = "0123456789ABCDEF".toCharArray();
```
yschimke: Can you pull this out into a useful test util class?

lexs: done

PR #247, opened 2017-03-01
Set version to 1.0

lexs:


reactivesocket-core/src/test/java/io/reactivesocket/frame/VersionFlyweightTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class VersionFlyweightTest {
+    @Test
+    public void simple() {
+        int version = VersionFlyweight.encode(1, 0);
+        assertEquals(1, VersionFlyweight.major(version));
```
yschimke: Can you add a test that doesn't rely on the main code to both cases?

assertEquals(0x0100, version);

lexs: done

Issue #244, opened 2017-02-23
Bring back tck driver implementation for reactivesocket-java


somasun:
The old tck implementation was removed in https://github.com/ReactiveSocket/reactivesocket-java/commit/a912a5f450146c6a717b7fd5a29eebb10a05ee75.  It would be nice to have it back to ensure parity across languages.

I will be first looking at cpp-tck driver implementation.  After that / time-permitting I plan to get back to java-tck implementation.  Others can feel free to give it a shot before that.

PR #249, opened 2017-03-02
Resurrect old tck-driver code and get it running

somasun:
This PR is to address #244.

I have resurrected the code from https://github.com/ReactiveSocket/reactivesocket-java/commit/a912a5f450146c6a717b7fd5a29eebb10a05ee75 and did a lot of changes to get it running.

It is a very basic effort to get the code running and have some tests passing (Stream/FireAndForget/RequestResponse).  The code is not yet in a great shape.  But want to get this out soon.  I can refine the code based on my available time-cycles.


reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/AddThread.java
```diff
@@ -0,0 +1,53 @@
...
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ *  specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivesocket.tckdrivers.common;
+
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * A thread that is created to wait to be able to add a marble string. We wait for the previous thread to have finished
```
benjchristensen: I don't understand this ... can you elaborate on what a "marble string" is, and why we need a new thread for this?

How many of these threads are running and blocking at any point in time? Do they block the underlying IO?

Extra threads probably doesn't matter here since it's just testing and not prod, but I'd like to confirm.

settings.gradle
```diff
@@ -25,4 +25,5 @@ include 'reactivesocket-spectator'
 include 'reactivesocket-test'
 include 'reactivesocket-transport-aeron'
 include 'reactivesocket-transport-local'
-include 'reactivesocket-transport-tcp'
\ No newline at end of file
+include 'reactivesocket-transport-tcp'
+include 'reactivesocket-tck-drivers'
```
benjchristensen: This should just include for compilation, which I'm okay with. But I don't want TCK tests running as part of the blocking continuous build yet until we prove we're happy with them manually. 

I don't see anything in this diff to cause them to run, but just confirming.

benjchristensen:
I think I'm good with this. 

@robertroeser @rdegnan are you okay with us merging this as we start to bring back the TCK and continue iterating forward?

reactivesocket-tck-drivers/src/main/java/io/reactivesocket/tckdrivers/common/AddThread.java
somasun: This part is copy-pasta from old code.  "marble string" refers to a marble representation (like - rr%%a%%b%%------------x--------------|&&{"x":{"hello":"goodbye"}}).  This class is used for channel tests, which is not complete as part of this PR.  But in general, the java tck-driver creates a thread per test (which seems to follow the recommendation from https://github.com/ReactiveSocket/reactivesocket-tck/blob/master/README.md).

settings.gradle
somasun: You are right. No tests should be running.  I have removed the tests tasks from the old gradle file.  https://github.com/ReactiveSocket/reactivesocket-java/pull/249/files#diff-516721e9b05fb6098ef05dbb7722adc3

somasun:
Rebased code to work with latest changes from 1.0.x branch

PR #250, opened 2017-03-03
Reactor

rdegnan:
Problem
Internal ReactiveStreams implementation was buggy, and required additional support. 

Modification
Switched to using Reactor-core, and Reactor-Netty for the netty tcp implementation.

Result
Less code, and abstractions using a well maintained reactive streams implementation.


benjchristensen:
This is a hell of a lot of files to review :-) I agree with the premise of the change so will scan through and see if anything stands out.

reactivesocket-core/build.gradle
```diff
@@ -14,30 +14,24 @@
...
     zip64 = true
     warmupBatchSize = 10
     iterations = 500
-
+    duplicateClassesStrategy = 'warn'
 }
 
 dependencies {
-    compile project(':reactivesocket-publishers')
+    compile 'io.projectreactor:reactor-core:3.0.5.RELEASE'
```
benjchristensen: Should we shade this and only expose the ReactiveStreams interfaces? Or are we okay with this public dependency?

The problem would come if Reactor makes breaking changes, when it should just be an implementation detail.

reactivesocket-core/build.gradle
```diff
@@ -14,30 +14,24 @@
...
+    compile 'org.agrona:Agrona:0.5.4'
```
benjchristensen: I think this should be shaded.

PR #253, opened 2017-03-07
Implement Payload NEXT bit

lexs:
The existing COMPLETE frame now generates a real PAYLOAD frame with the NEXT bit unset but the COMPLETE bit set. NEXT_COMPLETE sets both while NEXT only sets the NEXT bit.

Added tests to verify this behavior and also used reactivesocket-cli to verify frames sent/received.

Currently throwing an exception if we see a PAYLOAD frame with neither NEXT nor COMPLETE set pending https://github.com/ReactiveSocket/reactivesocket/issues/226.


benjchristensen:
Seems correct, though I may have misread something as there are many little details in this. Approving and trusting your unit test.

PR #257, opened 2017-03-09
Remove metadata length field from CANCEL, METADATA_PUSH and LEASE

lexs:


lexs:
As agreed here: https://github.com/ReactiveSocket/reactivesocket/issues/228

reactivesocket-core/src/test/java/io/reactivesocket/frame/LeaseFrameFlyweightTest.java
```diff
@@ -0,0 +1,19 @@
+package io.reactivesocket.frame;
+
+import org.agrona.concurrent.UnsafeBuffer;
+import org.junit.Test;
+
+import java.nio.ByteBuffer;
+
+import static org.junit.Assert.*;
+
+public class LeaseFrameFlyweightTest {
```
yschimke: Can you add 2 tests 

1) that passes with 2 ^ 24 - 1 - (9 + 4 * 2) bytes worth of metadata
and fails with
2) one byte more

yschimke: For context: the visible change should be that our metadata can be 3 bytes longer in a single frame than previously possible.  But I don't think we cover that with a test, and it isn't clear where it will fail currently when we break that limit? e.g. will it be delayed until the transport? Will transports without frame length encoding silently allow these invalid frames?

lexs: There is a test in FrameHeaderFlyweightTest that test encoding a frame at MAX_SIZE and then that it fails at MAX_SIZE + 1. I think that's the best you're going to get :)

PR #259, opened 2017-03-11
Fix frame size field serialization

lehecka:
According to the spec the field size should not be included in the frame size value.


lexs:
LGTM, cc @yschimke

Issue #264, opened 2017-03-23
request n is split and off by one


yschimke:
```
      Flux<Payload> stream = client.requestStream(payload);

    stream.subscribe(new PayloadSubscriber());

  private class PayloadSubscriber implements Subscriber<Payload> {
    @Override public void onSubscribe(Subscription s) {
      Shell.this.subscription = s;

        s.request(5);
    }
  ...
  }
```

15:51:15.628	sending -> Frame[0] => Stream ID: 1 Type: REQUEST_STREAM Initial Request N: 1 15:51:15.637	sending -> Frame[0] => Stream ID: 1 Type: REQUEST_N RequestN: 5 Payload:

PR #265, opened 2017-03-27
fixes client stream request n

robertroeser:
Problem
client wasn't sending correct initial request n (#264)

Modification
In LimitableRequestPublisher the added a call to request n after subscription, and in the ClientReactiveSocket moved sending the frames into a doOnRequest() method so that it can get the initial request n. 

Result
client sends initial request n

### DIFF
--- a/reactivesocket-core/src/main/java/io/reactivesocket/ClientReactiveSocket.java
+++ b/reactivesocket-core/src/main/java/io/reactivesocket/ClientReactiveSocket.java
@@ -77,7 +77,7 @@ public class ClientReactiveSocket implements ReactiveSocket {
     public Mono<Void> fireAndForget(Payload payload) {
         Mono<Void> defer = Mono.defer(() -> {
             final int streamId = streamIdSupplier.nextStreamId();
-            final Frame requestFrame = Frame.Request.from(streamId, FrameType.FIRE_AND_FORGET, payload, 0);
+            final Frame requestFrame = Frame.Request.from(streamId, FrameType.FIRE_AND_FORGET, payload, 1);
             return connection.sendOne(requestFrame);
         });
 
@@ -101,7 +101,7 @@ public class ClientReactiveSocket implements ReactiveSocket {
 
     @Override
     public Mono<Void> metadataPush(Payload payload) {
-        final Frame requestFrame = Frame.Request.from(0, FrameType.METADATA_PUSH, payload, 0);
+        final Frame requestFrame = Frame.Request.from(0, FrameType.METADATA_PUSH, payload, 1);
         return connection.sendOne(requestFrame);
     }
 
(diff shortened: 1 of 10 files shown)


reactivesocket-core/src/main/java/io/reactivesocket/Frame.java
```diff
@@ -397,6 +397,11 @@ public static int requestN(final Frame frame) {
     public static class Request {
         private Request() {}
 
+        public static Frame from(int streamId, FrameType type, Payload payload, long initialRequestN) {
+            int v = initialRequestN > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) initialRequestN;
```
yschimke: Can we reduce initialRequestN to int now?  Also this seems unsafe for input like -1, as that will get past this check.

reactivesocket-core/src/main/java/io/reactivesocket/ClientReactiveSocket.java
```diff
@@ -179,62 +181,98 @@ public ClientReactiveSocket start(Consumer<Lease> leaseConsumer) {
...
+            volatile MonoProcessor<Void> subscribedRequests;
+            boolean firstRequest = true;
+
+            boolean isValidToSendFrame() {
+                return contains(streamId) && connection.availability() > 0.0 && !receiver.isTerminated();
+            }
 
-            UnicastProcessor<Payload> receiver = UnicastProcessor.create();
+            void sendOneFrame(Frame frame) {
+                if (isValidToSendFrame()) {
```
yschimke: It seems like this silently drops the message, is that intended?  What are the semantics for DuplexConnection availability?

yschimke:
Generally LGTM - I'm learning for the review more than offering a full review.

yschimke:
BTW confirmed the fix locally

reactivesocket-core/src/main/java/io/reactivesocket/ClientReactiveSocket.java
robertroeser: Generally it means the duplex connection isn't healthy and there would be an error sending it  - i.e. its probably closed.

reactivesocket-core/src/main/java/io/reactivesocket/Frame.java
robertroeser: I added this as convenance method for the doOnRequest method. ReactiveStreams sends back-pressure signals as a Long. doOnRequest will only return positive numbers. I do the same thing in the Frame.RequestN.from method.

yschimke: The negative case still seems problematic.  Should it error on 0 or negative?

robertroeser: Yeah- I have it throw an exception if it's 0 or negative

yschimke: ok, I'll take your word for it.

reactivesocket-core/src/main/java/io/reactivesocket/ClientReactiveSocket.java
yschimke: Should this method throw if it can't send?

robertroeser: It will if there is something wrong with the connection, but there are times when it shouldn't send the frame

PR #267, opened 2017-03-30
Use ByteBuf instead of DirectBuffer

rdegnan:
Replaces the dependency on Agrona with netty-buffer, which provides similar functionality but allows us to avoid copying frame data for netty based transports and properly implement frame pooling.

New vs old allocations in TcpPing, for example, shows the effect this has on allocation pressure:

screen shot 2017-03-29 at 11 37 06 pm
screen shot 2017-03-29 at 11 37 23 pm


reactivesocket-core/src/main/java/io/reactivesocket/ServerReactiveSocket.java
```diff
@@ -143,71 +147,75 @@ public ServerReactiveSocket start() {
...
+                        case PAYLOAD:
+                            // TODO: Hook in receiving socket.
+                            return Mono.empty();
+                        case METADATA_PUSH:
+                            return metadataPush(new PayloadImpl(frame));
+                        case LEASE:
+                            // Lease must not be received here as this is the server end of the socket which sends leases.
+                            return Mono.empty();
+                        case NEXT:
+                            synchronized (ServerReactiveSocket.this) {
```
yschimke: I'd suggest pulling this out to a synchronized method to avoid the number of places that enforce this synchronization block

reactivesocket-core/src/main/java/io/reactivesocket/ServerReactiveSocket.java
```diff
@@ -321,33 +330,31 @@ private synchronized void cleanUpReceivingSubscription(Subscription subscription
...
-                .doOnRequest(l -> {
-                    if (connection.availability() > 0.0) {
-                        connection.sendOne(Frame.RequestN.from(streamId, l)).subscribe(null, errorConsumer::accept);
-                    }
-                })
-                .cast(Payload.class);
+        UnicastProcessor<Frame> frames = UnicastProcessor.create();
+        Flux<Payload> payloads = frames
+            .doOnCancel(() -> {
+                if (connection.availability() > 0.0) {
```
yschimke: tangential to this review.  Wouldn't this check be better handled within the connection?  it could always change between the check and the call anyway.

reactivesocket-core/src/main/java/io/reactivesocket/exceptions/Exceptions.java
```diff
@@ -28,31 +28,29 @@ private Exceptions() {}
...
 
         RuntimeException ex;
         switch (errorCode) {
             case APPLICATION_ERROR:
-                ex = new ApplicationException(frame);
+                ex = new ApplicationException(new PayloadImpl(frame));
                 break;
             case CONNECTION_ERROR:
-                ex = new ConnectionException(message);
+                ex = new ConnectionException(StandardCharsets.UTF_8.decode(frame.getData()).toString());
```
yschimke: Be good to have a utility method for this

reactivesocket-core/src/main/java/io/reactivesocket/Frame.java
```diff
@@ -15,234 +15,267 @@
...
-     * NOTE: always allocates. Used for pooling.
+     * If no data is present, the ByteBuffer will have 0 capacity.
      *
-     * @param directBuffer to wrap
-     * @return new {@link Frame}
+     * @return ByteBuffer containing the data
      */
-    public static Frame allocate(final MutableDirectBuffer directBuffer) {
-        return new Frame(directBuffer);
+    public ByteBuffer getData() {
```
yschimke: What are the pros/con for keeping this as ByteBuffer?  The design seems very netty centric anyway.  Is there value in making this ByteBuf and then putting additional work into the Aeron specific connectors?  I haven't considered whether that is possible to do efficiently.  Just asking the question.

reactivesocket-core/src/main/java/io/reactivesocket/util/PayloadImpl.java
```diff
@@ -25,37 +25,35 @@
...
+    }
 
     public PayloadImpl(String data) {
-        this(data, (String) null);
+        this(data, Charset.defaultCharset());
     }
 
     public PayloadImpl(String data, String metadata) {
-        this(fromString(data), fromString(metadata));
+        this(data, Charset.defaultCharset(), metadata, Charset.defaultCharset());
```
yschimke: Does defaultCharset() make sense for a network centric application? Should we just be more UTF_* opinionated instead?

yschimke:
LGTM cc @lexs

reactivesocket-core/src/main/java/io/reactivesocket/Frame.java
robertroeser: The code for the Aeron version will basically be the same because we need to copy the data. We need the bytes outside the fragment handler so you'd have to copy them. We were trying to keep the same signature as before for payload.  I thought people might care negatively if we exposed bytebuf in the signature of the payload.

yschimke: I kind of think Netty is good enough/ubiquitous enough that its ok.  It's a subtle change from reactivesocket-java is a network library that includes netty implementations, to reactivesocket-java is based on netty but also supports non-netty transports.

rdegnan: The main drawback of exposing ByteBuf in Payload from my point of view is that consumers of the library will have to worry about reference counting -- it would be nice to be zero copy all the way through, but I don't think it's worth the impact to usability.

reactivesocket-core/src/main/java/io/reactivesocket/exceptions/Exceptions.java
```diff
@@ -28,31 +28,29 @@ private Exceptions() {}
...
     public static RuntimeException from(Frame frame) {
         final int errorCode = Frame.Error.errorCode(frame);
-        ByteBuffer dataBuffer = frame.getData();
-        String message = dataBuffer.remaining() == 0 ? "<empty message>" : ByteBufferUtil.toUtf8String(dataBuffer);
 
         RuntimeException ex;
         switch (errorCode) {
             case APPLICATION_ERROR:
-                ex = new ApplicationException(frame);
+                ex = new ApplicationException(new PayloadImpl(frame));
```
rdegnan: You're right, it would probably be better to have a payload() method on the frame.

Issue #62, opened 2015-09-23
Scheduling of Writes (Head of Line Blocking)


benjchristensen:
Ensure that a single stream can not block other streams. This is a scheduling decision of the implementation.

NiteshKant:
Should this be a `DuplexConnection` implementation concern?

eg: RxNetty tries to be fair to all concurrent writers on the same connection (See [this](https://github.com/ReactiveX/RxNetty/pull/515) for more information on how). 

There will always be pathological cases where even 1 items saturates the channel buffer, in which case, we will have to buffer at the transport layer.

Issue #61, opened 2015-09-23
Fragmentation & Reassembly


benjchristensen:
Need to support fragmentation and reassembly.

benjchristensen:
We need to decide on fragmentation of requests: https://github.com/ReactiveSocket/reactivesocket/issues/77

Right now the protocol only allows fragmenting responses.

tmontgomery:
REQUEST_CHANNEL can be fragmented. At least it is specified in the frame that it has an F flag. Also, the fragmenter impl should support it. Off hand, not sure about the reassembler, but it should be easy to account for.

benjchristensen:
Right, REQUEST_CHANNEL does, so it's the other 4 REQUEST types.

tmontgomery:
Ah ok. I will add that to the spec in the next couple days.

benjchristensen:
Thanks Todd

benjchristensen:
A possible implementation of response fragmentation and reassembly is in pull request https://github.com/ReactiveSocket/reactivesocket-java/pull/63

PR #269, opened 2017-04-06
Adding fragmentation

robertroeser:
Adds a duplex connection that fragments frames based on an mtu. Uses ByteBuf to allow zero copy re-assembly. 

Addresses #62 and #61


reactivesocket-core/src/main/java/io/reactivesocket/Plugins.java
```diff
@@ -24,4 +25,30 @@
 
     private Plugins() {}
 
+    static {
+        if (Boolean.getBoolean("io.reactivesocket.fragmentation.enable")) {
+            int mtu = Integer.getInteger("io.reactivesocket.fragmentation.mtu", 1024);
+
+            if (Plugins.DUPLEX_CONNECTION_INTERCEPTOR == null) {
```
yschimke: What is the reasoning behind the system property plugin mechanism?  Shouldn't this be driven by individual applications wiring up connections as they see fit?  Could this be expressed non-statically via a helper that wraps a DuplexConnection with sensible defaults?

reactivesocket-core/src/main/java/io/reactivesocket/fragmentation/FragmentionDuplexConnection.java
```diff
@@ -0,0 +1,98 @@
...
+            return source.sendOne(frame);
+        }
+    }
+
+    @Override
+    public Flux<Frame> receive() {
+        return source
+            .receive()
+            .concatMap(frame -> {
+                if (FrameHeaderFlyweight.FLAGS_F == (frame.flags() & FrameHeaderFlyweight.FLAGS_F)) {
```
yschimke: can this be a method on frame? frame.isFlagSet?

reactivesocket-core/src/main/java/io/reactivesocket/fragmentation/FrameFragmenter.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class FrameFragmenter {
+    private final int mtu;
+
+    public FrameFragmenter(int mtu) {
+        this.mtu = mtu;
+    }
+
+    public boolean shouldFragment(Frame frame) {
+        return isFragmentableFrame(frame.getType()) && FrameHeaderFlyweight.payloadLength(frame.content()) > mtu;
```
yschimke: should the mtu be expressed in entire frame length instead of payload length? would you normally set mtu to avoid fragmentation at lower levels? TCP? HTTP/2?

reactivesocket-core/src/main/java/io/reactivesocket/fragmentation/FrameFragmenter.java
```diff
@@ -0,0 +1,72 @@
...
+        return Flux.generate(sink -> {
+            final int metadataLength = metadata.readableBytes();
+            final int dataLength = data.readableBytes();
+
+            if (metadataLength > mtu) {
+                sink.next(Frame.PayloadFrame.from(streamId, frameType, metadata.readSlice(mtu), Unpooled.EMPTY_BUFFER,
+                    flags | FrameHeaderFlyweight.FLAGS_M | FrameHeaderFlyweight.FLAGS_F));
+            } else if (metadataLength > 0) {
+                if (dataLength > 0) {
+                    sink.next(Frame.PayloadFrame.from(streamId, frameType, metadata.readSlice(metadataLength), Unpooled.EMPTY_BUFFER,
```
yschimke: The protocol spec says to include as much data as possible in this frame also

https://github.com/rsocket/rsocket/blob/master/Protocol.md#fragmentation-and-reassembly

reactivesocket-core/src/main/java/io/reactivesocket/fragmentation/FrameReassembler.java
```diff
@@ -0,0 +1,48 @@
...
+        this.frameType = frame.getType();
+        this.streamId = frame.getStreamId();
+        this.flags = frame.flags();
+        dataBuffer = PooledByteBufAllocator.DEFAULT.compositeBuffer();
+        metadataBuffer = PooledByteBufAllocator.DEFAULT.compositeBuffer();
+    }
+
+    public synchronized void append(Frame frame) {
+        if (FrameHeaderFlyweight.FLAGS_M == (FrameHeaderFlyweight.FLAGS_M & frame.flags())) {
+            final ByteBuf buffer = FrameHeaderFlyweight.sliceFrameMetadataRetained(frame.content());
```
yschimke: should also read data in this case

reactivesocket-core/src/main/java/io/reactivesocket/fragmentation/FragmentionDuplexConnection.java
```diff
@@ -0,0 +1,98 @@
...
+import io.reactivesocket.frame.FrameHeaderFlyweight;
+import io.reactivesocket.internal.Int2ObjectHashMap;
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+/**
+ * Fragments and Re-assembles frames. MTU is number of bytes per fragment. The default is 1024
+ */
+public class FragmentionDuplexConnection implements DuplexConnection {
```
yschimke: Fragmention is a strange word, is it the right one?

reactivesocket-core/src/main/java/io/reactivesocket/fragmentation/FragmentionDuplexConnection.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    @Override
+    public Flux<Frame> receive() {
+        return source
+            .receive()
+            .concatMap(frame -> {
+                if (FrameHeaderFlyweight.FLAGS_F == (frame.flags() & FrameHeaderFlyweight.FLAGS_F)) {
+                    FrameReassembler frameReassembler = getFrameReassmbler(frame);
```
yschimke: typo: Reassmbler

reactivesocket-core/src/main/java/io/reactivesocket/fragmentation/FrameReassembler.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public FrameReassembler(Frame frame) {
+        this.frameType = frame.getType();
+        this.streamId = frame.getStreamId();
+        this.flags = frame.flags();
+        dataBuffer = PooledByteBufAllocator.DEFAULT.compositeBuffer();
+        metadataBuffer = PooledByteBufAllocator.DEFAULT.compositeBuffer();
+    }
+
+    public synchronized void append(Frame frame) {
```
yschimke: We should check all following frames are of type Payload

reactivesocket-core/src/main/java/io/reactivesocket/fragmentation/FrameFragmenter.java
```diff
@@ -0,0 +1,72 @@
...
+            case NEXT_COMPLETE:
+            case METADATA_PUSH:
+                return true;
+            default:
+                return false;
+        }
+    }
+
+    public Flux<Frame> fragment(Frame frame) {
+        final FrameType frameType = frame.getType();
```
yschimke: frame type for all additional frames should be payload

reactivesocket-core/src/test/java/io/reactivesocket/fragmentation/FragmentionDuplexConnectionTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+                return Mono.empty();
+            });
+        when(mockConnection.sendOne(any(Frame.class))).thenReturn(Mono.empty());
+
+        ByteBuffer data = createRandomBytes(16);
+        ByteBuffer metadata = createRandomBytes(16);
+
+        Frame frame1 = Frame.Request.from(1, FrameType.REQUEST_RESPONSE, new PayloadImpl(data, metadata), 1);
+        Frame frame2 = Frame.Request.from(2, FrameType.REQUEST_RESPONSE, new PayloadImpl(data, metadata), 1);
```
yschimke: FrameType.PAYLOAD?

reactivesocket-core/src/main/java/io/reactivesocket/fragmentation/FrameFragmenter.java
rdegnan: Good catch

reactivesocket-core/src/main/java/io/reactivesocket/Plugins.java
robertroeser: Its static because are going to refactor setup and this was more for testing at moment

reactivesocket-core/src/main/java/io/reactivesocket/fragmentation/FrameFragmenter.java
robertroeser: You could, but this is also to prevent head of the line blocking. You send 32kb message most the time and everyone once and in a while a 128 meg one. That is my understanding.

Issue #271, opened 2017-04-26
RequestResponse Handler sends a CANCEL frame after the underlying connection has closed, resulting in crash.


somasun:
The sequence is as follows

RQ -> RS: REQUEST_RESPONSE
RS -> RQ: ERROR
RS -> RQ: CANCEL

This deviates from protocol definition [1] and could also result in java server crashing [2].

This could be reproduced by running a java server/client and making the java-server call onError right after receiving RequestResponse.
Example: 
client -> io.reactivesocket.tckdrivers.main.Main --client --host 127.0.0.1 --port 9898 --tests requestResponseTest3 
server -> io.reactivesocket.tckdrivers.main.Main --server --port 9898

[1] https://github.com/rsocket/rsocket/blob/master/Protocol.md
[2] Java server can crash if RQ closes the TCP connection right after receiving ERROR.  RQ will send a TCP_RST for the subsequent CANCEL frame.  Java server will crash with "java.io.IOException: Connection reset by peer"

PR #272, opened 2017-04-26
dont send CANCEL from RequestResponseHandler

somasun:
Addresses #271 

A RequestResponseHandler will never have to send CANCEL for response (according to spec)


yschimke:
Should handleStream be updated as well?

PR #274, opened 2017-05-03
Handle and swallow errors on the Mono<Payload> to stop cancelling Flux<Frame>

yschimke:


yschimke:
Good catch, it worked for the error case, but failed on 3 successes

reactivesocket-core/src/main/java/io/reactivesocket/ServerReactiveSocket.java
```diff
@@ -288,16 +284,8 @@ private synchronized void cleanUpChannelProcessors() {
...
-                    if (connection.availability() > 0.0) {
-                        connection.sendOne(Frame.Error.from(streamId, t)).subscribe(null, errorConsumer);
-                    }
-                })
-                .doOnComplete(() -> {
-                    if (connection.availability() > 0.0) {
-                        connection.sendOne(Frame.PayloadFrame.from(streamId, FrameType.COMPLETE)).subscribe(null, errorConsumer);
-                    }
-                })
+                .doOnComplete(() -> Mono.just(Frame.PayloadFrame.from(streamId, FrameType.COMPLETE)))
```
yschimke: Fixing this, needs to be part of the stream

yschimke:
Landed because its a high priority bug fix for us.  Will address any comments in follow up reviews.

yschimke:
cc @rdegnan @robertroeser 

n.b. I removed the connection availability check in some places, happy to discuss if still needed.

PR #276, opened 2017-05-04
Allow errorConsumer to be set

yschimke:
This change just make it possible for client and server implementors to provide an alternative errorConsumer


yschimke:
@rdegnan @robertroeser 

Should the start method swallow exceptions since it has already passed to errorConsumer? Would termination the stream via onErrorContinueWith(empty()) be preferable?

robertroeser:
@yschimke probably not, but it's ok for now. planning on refactoring the setup stuff and clearing it/simplifying it.

yschimke:
@robertroeser 

my happy place would be

- all application/network errors handled by reactivesocket hooks or fallback to errorConsumer
- only a serious rs-java bug hits Netty pipeline loggers or Thread uncaught exception handler

PR #278, opened 2017-05-05
Fix and simplify keepalives

yschimke:
1) KeepAliveProvider.from(durationInMillis)
2) KeepAliveProvider::ticks returns Flux\<Object\> instead or Flux\<Long\>
3) correct the starting logic for keepalives
4) Exception message io.reactivesocket.exceptions.ConnectionException: Missed 4 keepalive from the peer, last keepalive 1493974801675 last ack 1493974793366


rsocket-core/src/main/java/io/rsocket/client/KeepAliveProvider.java
```diff
@@ -29,20 +30,20 @@
  */
 public final class KeepAliveProvider {
 
-    private volatile boolean ackThresholdBreached;
+    private volatile ConnectionException ackThresholdBreached;
     private volatile long lastKeepAliveMillis;
     private volatile long lastAckMillis;
-    private final Flux<Long> ticks;
+    private final Flux<Object> ticks;
```
lexs: Why does `Object` make sense here? Perhaps `Number` in that case?

yschimke: The numbers are only used in tests, nowhere else.  The wallclock spacing between ticks is the interesting part, not the value emitted each time it ticks.  I'll make this clearer in the documentation.

PR #280, opened 2017-05-05
tcp concurrent test

yschimke:


robertroeser:
A DirectProcessor doesn't actually buffer any events if you aren't subscribed so the test is just dropping them. I image the reason it works when with onComplete the processor is complete so you don't block anything. I think this test does what you want:

```
@Test(timeout = 2_000L)
    public void testTwoConcurrentStreams() throws InterruptedException {
        // Two concurrent streams
        // Stream 1 is a single response
        // Stream 2 is a single response, then completed by the server
        // Stream 1 is then cancelled by the client
        
        ConcurrentHashMap<String, UnicastProcessor<Payload>> map = new ConcurrentHashMap<>();
        UnicastProcessor<Payload> processor1 = UnicastProcessor.create();
        map.put("REQUEST1", processor1);
        UnicastProcessor<Payload> processor2 = UnicastProcessor.create();
        map.put("REQUEST2", processor2);
        
        handler = new AbstractReactiveSocket() {
            @Override
            public Flux<Payload> requestStream(Payload payload) {
                ByteBuffer data = payload.getData();
                byte[] b = new byte[data.remaining()];
                data.get(b);
                String s = new String(b);
                return map.get(s);
            }
        };
        
        ReactiveSocket client = buildClient();
        
        Flux<Payload> response1 = client.requestStream(new PayloadImpl("REQUEST1"));
        Flux<Payload> response2 = client.requestStream(new PayloadImpl("REQUEST2"));
        
        processor1.onNext(new PayloadImpl("RESPONSE1A"));

        processor2.onNext(new PayloadImpl("RESPONSE2A"));
        processor2.onComplete();
        
        CountDownLatch latch = new CountDownLatch(3);
        
        response2
            .subscribeOn(Schedulers.newSingle("2"))
            .subscribe(c -> latch.countDown(), t -> {
            }, latch::countDown);
        
        response1
            .subscribeOn(Schedulers.newSingle("1"))
            .subscribe(c -> latch.countDown());
        
        latch.await();
        
    }
```

yschimke:
@robertroeser Thanks, that's working now.  I've updated the test slightly to show the next happens for both streams before the complete.  Hopefully it's just the KeepAlive bug that was causing our problems.

PR #282, opened 2017-05-05
add tck tests on travis push

ragansa:
Causes TCK tests to run when a new pull request is made


yschimke:
LGTM - I'd like an ACK before merging from @rdegnan @robertroeser that they are bought into keeping these working

robertroeser:
I would prefer to make this not use RXJava 2 TestSubscriber first, and make it a gradle task - not a separate main method -  before it is run all the time.

rdegnan:
Agreed, I can take a stab at refactoring to add a gradle task/remove the
use of TestSubscriber.

On May 10, 2017, at 3:57 PM, Robert Roeser <notifications@github.com> wrote:

I would prefer to make this not use RXJava 2 TestSubscriber first, and make
it a gradle task - not a separate main method - before it is run all the
time.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/ReactiveSocket/reactivesocket-java/pull/282#issuecomment-300635864>,
or mute the thread

yschimke:
@rdegnan @robertroeser can we land this and refactor the main method later?

robertroeser:
@yschimke yeah that's fine.

robertroeser:
@yschimke the server start blocks - this probably needs to be refactor to emit a Mono that the RSocketFactory produces. When the Mono completes you can start the test. That probably needs to happen for this to work.

yschimke:
This code has been refactored to use @Test instead

PR #283, opened 2017-05-07
echo keepalive data

yschimke:
As specified https://github.com/rsocket/rsocket/blob/master/Protocol.md#keepalive-frame-0x03 echo the keepalive data payload in the response


yschimke:
@lexs it's encoded inline, so there is no need to retain the buffer based on the current call to Frame.Keepalive.from

https://github.com/ReactiveSocket/reactivesocket-java/blob/965d3d6f70a0389948b457a0354a46b984091a70/rsocket-core/src/main/java/io/rsocket/frame/KeepaliveFrameFlyweight.java#L46

PR #287, opened 2017-05-19
Update buildViaTravis.sh

yschimke:
Disabling publishing of snapshots and releases until the permissions for Sonatype are sorted


yschimke:
cc @benjchristensen @stevegury

yschimke:
Shipping as a broken build doesn't help anyone

PR #290, opened 2017-05-21
RFC: send error frame on bad version

yschimke:
Need some guidance on how to cleanly implement this.

Open questions

- Should all state changes be done inside a Mono block, or can cancellation actions be done inline?
- Current behaviour is exception (ClosedChannelException) if error received before sending, or timeout if received afterwards.  How to cleanly handle this?

TODO

- Test by sending a raw frame (bad version) to a duplex connection


robertroeser:
@yschimke I think the way you have it is good. I don't think you need to wrap this in a mono because it happens inside an chain of publishers, and won't start until you subscribe. 

_Current behaviour is exception (ClosedChannelException) if error received before sending, or timeout if received afterwards. How to cleanly handle this?_
Can you explain this a bit more - you mean in the context of when it returns an error during setup?

yschimke:
Probably best if I add two integration tests to demo. It doesn't work end to end like I'd expect.  But I see the right frames in the wire logging.

yschimke:
I'll add a test based on the plugin support, so landing this.

PR #291, opened 2017-05-25
Single TCP connection for TCK driver client

junaidkhalid:
Modified the javaClientDriver to open a single TCP connection, instead of opening a new TCP connection for every new request/stream.

### DIFF
--- a/rsocket-tck-drivers/src/main/java/io/rsocket/tckdrivers/client/JavaClientDriver.java
+++ b/rsocket-tck-drivers/src/main/java/io/rsocket/tckdrivers/client/JavaClientDriver.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
 
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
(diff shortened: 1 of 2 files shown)


rsocket-tck-drivers/src/main/java/io/rsocket/tckdrivers/client/JavaClientDriver.java
```diff
@@ -519,7 +519,7 @@ private void handleCancel(String[] args) {
 
     private void handleEOF() {
         MySubscriber<Void> fnfsub = new MySubscriber<>(0L, AGENT);
-        RSocket fnfclient = createClient.get();
+        RSocket fnfclient = createClient;
```
yschimke: nit: inline these

rsocket-tck-drivers/src/main/java/io/rsocket/tckdrivers/client/JavaClientDriver.java
```diff
@@ -53,12 +53,12 @@
     private final Map<String, MySubscriber<Payload>> payloadSubscribers;
     private final Map<String, MySubscriber<Void>> fnfSubscribers;
     private final Map<String, String> idToType;
-    private final Supplier<RSocket> createClient;
+    private final RSocket createClient;
```
yschimke: Can you call this just client.

yschimke:
Failing on the TcpIntegrationTest I added conflicting ports, I'll fix that separately.

yschimke:
I don't know how important it is to remove the Supplier, you could do this as a 1 line change by making your supplier just return a single instance.

junaidkhalid:
I removed Supplier because I thought java.util.function.Supplier does not have memorization option. How can I make Supplier to return a single instance ?

yschimke:
@junaidkhalid Supplier is a functional interface, so it can do anything.  I meant just

Client c = createClient();
Supplier<Client> sc = () -> c;

yschimke:
@junaidkhalid I don't think I'm explaining myself well.  Sorry.

My point was because it is a supplier, you didn't need any changes in JavaClientDriver.java to support what you wanted.  You could make the supplier in JavaTCPClient.java return a constant value.

PR #292, opened 2017-05-26
Add Statistics related structures to RSocket to observe runtime behavior

phoad:
<Work in Progress - to get early feedback>

Add some hooks into RSocket server/client to make it more observable. With the help of RSocketStats, we will be able to understand number of active connections per time period, number of bytes sent, errors per time period, etc.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
+++ b/rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
@@ -113,13 +113,8 @@ public interface RSocketFactory {
 
         private @Nullable RSocketStats stats;
 
-        public ClientRSocketFactory stats(RSocketStats stats) {
+        public ClientRSocketFactory stats(@Nullable RSocketStats stats) {
             this.stats = stats;
-
-            if (stats != null) {
-                stats.socketCreated();
-            }
-
             return this;
         }
 
@@ -225,11 +220,13 @@ public interface RSocketFactory {
                     .connect()
                     .then(connection -> {
                         if (stats != null) {
+                            stats.socketCreated();
                             stats.duplexConnectionCreated(connection);
                             connection
                                 .onClose()
                                 .doFinally(signalType -> {
                                     stats.duplexConnectionClosed(connection);
+                                    stats.socketClosed(signalType);
                                 })
                                 .subscribe();
                         }
@@ -294,11 +291,8 @@ public interface RSocketFactory {
             return new ServerTransport();
         }
 
-        public ServerRSocketFactory stats(RSocketStats stats) {
+        public ServerRSocketFactory stats(@Nullable RSocketStats stats) {
             this.stats = stats;
-            if (stats != null) {
-                stats.socketCreated();
-            }
             return this;
         }
 
@@ -329,11 +323,13 @@ public interface RSocketFactory {
                     .get()
                     .start(connection -> {
                         if (stats != null) {
+                            stats.socketCreated();
                             stats.duplexConnectionCreated(connection);
                             connection
                                 .onClose()
                                 .doFinally( signalType -> {
                                     stats.duplexConnectionClosed(connection);
+                                    stats.socketClosed(signalType);
                                 })
                                 .subscribe();
                         }
@@ -344,8 +340,7 @@ public interface RSocketFactory {
                                 new FragmentationDuplexConnection(connection, mtu),
                                 stats);
                         } else {
-                            multiplexer = new ClientServerInputMultiplexer(connection,
-                                stats);
+                            multiplexer = new ClientServerInputMultiplexer(connection, stats);
                         }
 
                         return multiplexer
--- a/rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
+++ b/rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
@@ -3,11 +3,13 @@ package io.rsocket;
 import io.rsocket.fragmentation.FragmentationDuplexConnection;
 import io.rsocket.frame.SetupFrameFlyweight;
 import io.rsocket.internal.ClientServerInputMultiplexer;
+import io.rsocket.stat.RSocketStats;
 import io.rsocket.transport.ClientTransport;
 import io.rsocket.transport.ServerTransport;
 import io.rsocket.util.PayloadImpl;
 import reactor.core.publisher.Mono;
 
+import javax.annotation.Nullable;
 import java.time.Duration;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
@@ -109,6 +111,18 @@ public interface RSocketFactory {
         private String dataMineType = "application/binary";
         private String metadataMimeType = "application/binary";
 
+        private @Nullable RSocketStats stats;
+
+        public ClientRSocketFactory stats(RSocketStats stats) {
+            this.stats = stats;
+
+            if (stats != null) {
+                stats.socketCreated();
+            }
+
+            return this;
+        }
+
         @Override
         public ClientRSocketFactory keepAlive() {
             tickPeriod = Duration.ofSeconds(20);
@@ -169,7 +183,7 @@ public interface RSocketFactory {
             @Override
             public Start<RSocket> transport(Supplier<io.rsocket.transport.ClientTransport> transportClient) {
                 ClientRSocketFactory.this.transportClient = transportClient;
-                return new StartClient();
+                return new StartClient(ClientRSocketFactory.this.stats);
             }
         }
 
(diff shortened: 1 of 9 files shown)


rsocket-core/build.gradle
```diff
@@ -28,6 +28,7 @@ jmh {
 dependencies {
     compile 'io.projectreactor:reactor-core:3.0.7.RELEASE'
     compile 'io.netty:netty-buffer:4.1.9.Final'
+    compile project(':rsocket-transport-netty')
```
yschimke: I don't think this can work.  core should be core.

rsocket-core/src/jmh/java/io/rsocket/RSocketPerf.java
```diff
@@ -90,8 +92,9 @@ public void fireAndForgetHello(Input input) {
         static final TestDuplexConnection clientConnection = new TestDuplexConnection(serverReceive, clientReceive);
         static final TestDuplexConnection serverConnection = new TestDuplexConnection(clientReceive, serverReceive);
 
+        static final RSocketStats stats = RSocketStatsFactory.create();
         static final Object server = RSocketFactory
-            .receive()
+            .receive(stats)
```
yschimke: make stats a builder method instead of a new parameter

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
```diff
@@ -109,6 +110,12 @@ static ServerRSocketFactory receive() {
         private String dataMineType = "application/binary";
         private String metadataMimeType = "application/binary";
 
+        private RSocketStats stats;
```
yschimke: @Nullable

rsocket-core/src/main/java/io/rsocket/stat/NoopStats.java
```diff
@@ -0,0 +1,65 @@
+package io.rsocket.stat;
+
+import io.rsocket.DuplexConnection;
+import io.rsocket.FrameType;
+
+/**
+ * Created by geleri on 5/25/17.
+ */
+public class NoopStats implements RSocketStats{
```
yschimke: This is a useful base class, maybe call this BaseStats and allow people to extend this

rsocket-core/src/main/java/io/rsocket/stat/RSocketStatsFactory.java
```diff
@@ -0,0 +1,21 @@
+package io.rsocket.stat;
+
+/**
+ * Either use the default NoopStats, or assign another implementation of
+ * RSocketStatsFactory to generate other Stats instances.
+ */
+public class RSocketStatsFactory {
```
yschimke: Are you sure we need factories for these?

rsocket-core/src/test/java/io/rsocket/RSocketStatsTest.java
```diff
@@ -0,0 +1,94 @@
...
+        assertThat("Number of disconnected connections should be 0", testStats.disconnectedSocketCount, is(0));
+    }
+
+    private static class SocketAcceptorImpl implements SocketAcceptor {
+        @Override
+        public Mono<RSocket> accept(ConnectionSetupPayload setupPayload, RSocket reactiveSocket) {
+            return Mono.just(new AbstractRSocket() {
+                @Override
+                public Flux<Payload> requestStream(Payload payload) {
+                    return Flux.interval(Duration.ofMillis(100))
```
yschimke: why interval here? This adds 1 second for the test to run, correct?

rsocket-transport-netty/src/test/java/io/rsocket/transport/netty/WebsocketPongServer.java
```diff
@@ -16,14 +16,17 @@
...
 import io.rsocket.test.PingHandler;
 import io.rsocket.transport.netty.server.WebsocketServerTransport;
 
 public final class WebsocketPongServer {
 
     public static void main(String... args) throws Exception {
+        RSocketStats stats = RSocketStatsFactory.create();
         RSocketFactory
-            .receive()
+            .receive(stats)
```
yschimke: I think a good sign of this change being seamless, is if you don't need to change these examples which don't care about stats.

phoad: Good suggestion.

rsocket-core/src/main/java/io/rsocket/stat/RSocketStatsFactory.java
phoad: Yes. I got your point.

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
```diff
@@ -109,6 +111,18 @@ static ServerRSocketFactory receive() {
         private String dataMineType = "application/binary";
         private String metadataMimeType = "application/binary";
 
+        private @Nullable RSocketStats stats;
+
+        public ClientRSocketFactory stats(RSocketStats stats) {
```
yschimke: @Nullable

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
```diff
@@ -109,6 +111,18 @@ static ServerRSocketFactory receive() {
...
+            this.stats = stats;
+
+            if (stats != null) {
+                stats.socketCreated();
```
yschimke: Is this possibly too early? e.g. can the builder be left in a dry state, and triggered later?

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
```diff
@@ -257,15 +286,22 @@ public ClientRSocketFactory setupPayload(Payload payload) {
...
         @Override
         public Transport<io.rsocket.transport.ServerTransport, Closeable> acceptor(Supplier<SocketAcceptor> acceptor) {
             this.acceptor = acceptor;
             return new ServerTransport();
         }
 
+        public ServerRSocketFactory stats(RSocketStats stats) {
+            this.stats = stats;
+            if (stats != null) {
+                stats.socketCreated();
```
yschimke: This seems like its per server, which can have at any time 0..n connections.  Should this be deferred to ServerStart::start where it receives a DuplexConnection?

yschimke:
Adding non FB reviewers, and it's big enough change that I think getting broad agreement is warranted.

robertroeser:
I added the plugin so we can wrap RSocket, and Duplex connections so we could add instrumentation like this without having to include it in RSocket-core. I think you should be able to do everything you're doing here using that. If something missing we should add it there to make your 
stuff work, and use composition.

For instance you should be able to figure out the number of bytes, and if the connection is new or not using the DuplexConnectionInterceptor.

The spector package uses this to add metrics.
https://github.com/rsocket/rsocket-java/tree/1.0.x/rsocket-spectator

tmontgomery:
It might be worth considering to use byte weaving instead to add in stats gathering. Makes the code a lot cleaner, faster, and opens up new methods of doing things.

yschimke:
@robertroeser damn. If i saw the metrics package before we could probably have started with that. Is it easy to split out stats when there are multiple servers and clients?

yschimke:
@tmontgomery if it can be done externally then I'm all for it being supported. But Having gone through the aspectj wars, probably like you, I lost too many friends in that battle to sign up for another tour of duty.

tmontgomery:
@yschimke indeed. However, just like anything.... judicious usage can be good. AOP is a tool. Sometimes it fits and sometimes it doesn't. ByteBuddy is quite nice in that it is simple to use and avoids some of the nastier edges.

To be a little more succinct. Always on stats are best handled, IMO, via writing them to shared memory and having another async process/thread reading them. For logging, bytecode weaving can be a very useful means to hook without cluttering code with hooks, special cases, and running afoul of the optimizer..... but still logging needs to be async and done in binary.

yschimke:
@tmontgomery yep. I don't want to introduce a dependency on our stats library (which does what you suggest). So I'd like an interface that we can implement and register at a per client socket or per server instance.

yschimke:
@phoad is doing the work, so he'll drive the design.

That said, the pattern in rsocket-spectator looks like a great approach to adding in stats (com.netflix.spectator.api.Counter, fb303, finagle util-stats) or tracing (zipkin etc).

There is zero cost (it seems) when not enabled, and can be added in as a plugin.

My one concern is that I think we should change the Plugins from being a JVM singleton registry, to per client or per server.  I don't think the current model works well, if you are a client to two different reactivesocket services, or you are both a client and a server simultaneously.

phoad:
I am pulling back the change. I will be expecting @yschimke 's changes.

PR #293, opened 2017-05-26
ephemeral port

yschimke:
Allow TCP server to use an ephemeral port

### DIFF
--- a/rsocket-examples/src/test/java/io/rsocket/integration/TcpIntegrationTest.java
+++ b/rsocket-examples/src/test/java/io/rsocket/integration/TcpIntegrationTest.java
@@ -6,6 +6,7 @@ import io.rsocket.Payload;
 import io.rsocket.RSocket;
 import io.rsocket.RSocketFactory;
 import io.rsocket.transport.netty.client.TcpClientTransport;
+import io.rsocket.transport.netty.server.NettyContextClosable;
 import io.rsocket.transport.netty.server.TcpServerTransport;
 import io.rsocket.util.PayloadImpl;
 import io.rsocket.util.RSocketProxy;
@@ -17,7 +18,6 @@ import reactor.core.publisher.Mono;
 import reactor.core.publisher.UnicastProcessor;
 import reactor.core.scheduler.Schedulers;
 
-import java.net.InetSocketAddress;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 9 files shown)


rsocket-examples/src/test/java/io/rsocket/integration/TcpIntegrationTest.java
```diff
@@ -28,21 +29,25 @@
...
             .block();
+
+        address = serverTransport.address();
     }
 
     private RSocket buildClient() {
         return RSocketFactory
             .connect()
-            .transport(TcpClientTransport.create("localhost", 8000))
+            .transport(TcpClientTransport.create(address.getHostString(), address.getPort()))
```
yschimke: Considering adding TcpClientTransport.create(InetSocketAddress) here

yschimke:
cc @phoad

rsocket-examples/src/test/java/io/rsocket/integration/TcpIntegrationTest.java
```diff
@@ -28,21 +29,25 @@
...
+    private InetSocketAddress address;
 
     @Before
     public void startup() {
+        TcpServerTransport serverTransport = TcpServerTransport.create("localhost", 0);
         server = RSocketFactory
             .receive()
             .acceptor((setup, sendingSocket) -> Mono.just(new RSocketProxy(handler)))
-            .transport(TcpServerTransport.create("localhost", 8000))
+            .transport(serverTransport)
```
yschimke: Ideally server would be a better type than Closeable, then I could get address off that.

rsocket-examples/src/test/java/io/rsocket/integration/TcpIntegrationTest.java
```diff
@@ -77,6 +82,9 @@ public void testSingleStream() throws InterruptedException {
 
         RSocket client = buildClient();
 
+        // TODO without this failure is always a timeout
+        Thread.sleep(1000);
```
phoad: nit

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/TcpServerTransport.java
```diff
@@ -47,15 +54,21 @@ public static TcpServerTransport create(TcpServer server) {
...
+        Mono<? extends NettyContext> serverMono = server
+                .newHandler((in, out) -> {
+                    in.context().addHandler("server-length-codec", new RSocketLengthCodec());
+                    NettyDuplexConnection connection = new NettyDuplexConnection(in, out, in.context());
+                    acceptor.apply(connection).subscribe();
+
+                    return out.neverComplete();
+                });
+
+        serverMono.map(s -> s.address()).subscribe(address);
```
phoad: If we provide port value of "0", each time we call start(), it generates a new NettyContext, which has a different "address" - "port" value. So if the caller first calls start(..) and then asks for the "address" value, while another caller also calls start(..) and asks for the address value, then there will be a race condition. Other than that, when we call start(..) at the inside of a single thread, the address value of the initial call will be lost.

It seems we only need the "address" if we pass port value of "0", otherwise address can be created by merging "bindAddress" and "port"? If so, we might have a overload method which also provides the address value back? If caller really needs the address back, the caller can call this overloaded one. So we will not need any class member and it will still be ok for multithreading?

yschimke: I hadn't considered that you would reuse a TcpServerTransport.  It seems problematic to do so as the code currently stands as you either

1) use a static port and clients are responsible for remembering the state e.g. is there an open server port?  The start call will fail unless a client has called closed.
2) use a free port (0) and then multiple calls to start work but with the race condition and address() is on the wrong class.

Overall this makes me think my original comment in the PR is correct "Alternatively we can make sure that server is a better type than Closeable."  So this start method should return NettyDuplexConnection instead of Closeable.

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/TcpServerTransport.java
```diff
@@ -46,16 +45,14 @@ public static TcpServerTransport create(TcpServer server) {
...
-            .newHandler((in, out) -> {
-                in.context().addHandler("server-length-codec", new RSocketLengthCodec());
-                NettyDuplexConnection connection = new NettyDuplexConnection(in, out, in.context());
-                acceptor.apply(connection).subscribe();
-
-                return out.neverComplete();
-            })
-            .map(NettyContextClosable::new);
-    }
+                .newHandler((in, out) -> {
```
phoad: nit - indentation change.

PR #296, opened 2017-05-31
Request Response, Request Stream and fireandforget tests

junaidkhalid:
Tests are generated from
ReactiveSocket/reactivesocket-tck/blob/master/src/main/scala/io/reactivesocket/tck/ReactiveSocketCoreTests.scala

Modified requestResponseRequestAfterCancel() and requestStreamAfterCancel() to
remove "(pass = fail)" because it is not supported anymore by tckdrivers.


yschimke:
@junaidkhalid did these tests pass for you? I had trouble running TestMain.

https://github.com/rsocket/rsocket-java/issues/297

junaidkhalid:
I ran all these tests with io.rsocket.tckdrivers.main.Main, and they pass. 
I will try to run them with TestMain.

junaidkhalid:
@yschimke 
In case of TestMain, waitStart.countDown() is not getting called. It was removed in https://github.com/robertroeser/reactivesocket-java/commit/4e66dd074cf6d14f189a7253c9fd4fb2bd0d96bb#diff-beb06ddd53c49a3fd2474f3a0abd9e22

I did not face this issue, when I was running the "Main" because it does not wait(it does not call st.awaitStart())

PR #300, opened 2017-06-04
Exposed server rsocket to client acceptor

gregwhitaker:
Currently, it is not possible for the client side acceptor to call the
server rsocket.

This change exposes the server rsocket to the client acceptor.


yschimke:
Could you revert the formatting changes in this diff?  I'm not actually clear whether its right or wrong, but we should do that in a separate diff even if its desired.

robertroeser:
looks like it just added some spacing, but I see what this does - fixes duplexing on the client side.  can we merge this?

PR #306, opened 2017-06-11
Plugins refactor

yschimke:
RFC, still hacking on this.

- Allow per client/server plugins
- move fragmentation to the existing MTU on factories
- default to existing global plugins


rsocket-core/src/main/java/io/rsocket/plugins/PluginRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+  public PluginRegistry() {}
+
+  public PluginRegistry(PluginRegistry defaults) {
+    this.connections.addAll(defaults.connections);
+    this.clients.addAll(defaults.clients);
+    this.servers.addAll(defaults.servers);
+  }
+
+  public void addConnectionPlugin(DuplexConnectionInterceptor interceptor) {
+    connections.add(interceptor);
```
phoad: It doesn't look like differentiating between different server/clients but applies the same interceptor list to all server and clients..

rsocket-core/src/main/java/io/rsocket/plugins/Plugins.java
```diff
@@ -0,0 +1,24 @@
+package io.rsocket.plugins;
+
+/** JVM wide plugins for RSocket */
```
phoad: Do you have a TODO to don't make this JVM wide?

rsocket-examples/src/test/java/io/rsocket/integration/IntegrationTest.java
```diff
@@ -90,6 +93,8 @@ public void startup() {
 
     server =
```
phoad: What about creating two+ servers and two+ clients and being able to use different interceptors for them?

yschimke: I'll test that via more specific tests, e.g. a unit test of the builders that shows the right interceptors are applied.  I think Integration tests are better left for testing end to end scenarios.

rsocket-core/src/main/java/io/rsocket/plugins/Plugins.java
yschimke: It's part of this PR, you can add interceptors to the RSocketFactory.  But I got the impression that it may be a useful feature to apply some plugins at app startup for all clients.  I'm not against that being used judiciously.

rsocket-core/src/main/java/io/rsocket/plugins/PluginRegistry.java
yschimke: If you have 1 client, 1 server, then you have 3 plugin registries.

- JVM defaults via Plugins
- RSocketFactory.receive() holds one
- RSocketFactory.connect() holds one

So I think that counts.

rsocket-core/src/main/java/io/rsocket/Plugins.java
```diff
@@ -1,49 +0,0 @@
...
-  public static volatile DuplexConnectionInterceptor DUPLEX_CONNECTION_INTERCEPTOR =
-      NOOP_DUPLEX_CONNECTION_INTERCEPTOR;
-  public static volatile RSocketInterceptor CLIENT_REACTIVE_SOCKET_INTERCEPTOR = NOOP_INTERCEPTOR;
-  public static volatile RSocketInterceptor SERVER_REACTIVE_SOCKET_INTERCEPTOR = NOOP_INTERCEPTOR;
-
-  private Plugins() {}
-
-  static {
-    if (Boolean.getBoolean("io.rsocket.fragmentation.enable")) {
-      int mtu = Integer.getInteger("io.rsocket.fragmentation.mtu", 1024);
```
yschimke: This is removed as its already supported in the Builder, so I can move this system property logic to become the default MTU.

If I leave it here, you get fragmentation twice if user overrides.

yschimke:
@phoad do you want to see if this works for you now? Let me know and if so I'll merge in.

phoad:
I have re-reviewed the update and it seems good to me.

PR #307, opened 2017-06-12
URI Registry

yschimke:
Designed to allow for things like CLI or load test clients that could specify a target purely by a URI supported by a provider on the classpath


yschimke:
cc @tmontgomery

yschimke:
TODO still support aeron + local + websocket + tcp (server)

tmontgomery:
Looks good to me. Each transport has their own URI scheme. So, should be straight forward.

yschimke:
@tmontgomery I'll complete and then get a final shipit.  Thanks

yschimke:
Merging - based on earlier review.  Feel free to add and review comments and I'll address.

Issue #311, opened 2017-06-14
Setup failures are seen by client as connection closed 


yschimke:
Client receives the error frame with a useful reason, but it doesn't bubble up to callers.

yschimke:
Test that shows the problem

https://github.com/yschimke/rsocket-java/pull/1

yschimke:
10:33:47,747 DEBUG [reactor-tcp-nio-7] (ClientServerInputMultiplexer.java) - receiving -> Frame => Stream ID: 0 Type: SETUP Version: 99.34 keep-alive interval: 30000 max lifetime: 90000 metadata mime type: application/binary data mime type: application/binary Payload: 
10:33:47,749  INFO [Time-limited test] (Loggers.java) - onSubscribe(FluxConcatArray.ConcatArraySubscriber)
10:33:47,753  INFO [Time-limited test] (Loggers.java) - request(unbounded)
10:33:47,755 DEBUG [reactor-tcp-nio-7] (ClientServerInputMultiplexer.java) - sending -> Frame => Stream ID: 0 Type: ERROR Error code: 1 Payload: data: "Unsupported version 99.34" 
10:33:47,760 DEBUG [reactor-tcp-nio-6] (ClientServerInputMultiplexer.java) - receiving -> Frame => Stream ID: 0 Type: ERROR Error code: 1 Payload: data: "Unsupported version 99.34" 
io.rsocket.exceptions.InvalidSetupException: Unsupported version 99.34
	at io.rsocket.exceptions.Exceptions.from(Exceptions.java:43)
	at io.rsocket.RSocketClient.handleStreamZero(RSocketClient.java:387)
	at io.rsocket.RSocketClient.handleIncomingFrames(RSocketClient.java:375)
	at reactor.core.publisher.FluxPeek$PeekSubscriber.onNext(FluxPeek.java:165)
	at reactor.core.publisher.FluxPeek$PeekSubscriber.onNext(FluxPeek.java:173)
	at reactor.core.publisher.MonoFlatMap$FlatMapInner.onNext(MonoFlatMap.java:235)
	at reactor.core.publisher.FluxPeekFuseable$PeekFuseableSubscriber.onNext(FluxPeekFuseable.java:188)
	at reactor.core.publisher.FluxGroupBy$UnicastGroupedFlux.drainRegular(FluxGroupBy.java:566)
	at reactor.core.publisher.FluxGroupBy$UnicastGroupedFlux.drain(FluxGroupBy.java:642)
[...]

PR #317, opened 2017-06-16
Setup failure should close all streams

yschimke:


yschimke:
Needs a thorough review of both the test behaviour I expect, and the way I fixed it in RSocketClient.

yschimke:
Attempted fix for https://github.com/rsocket/rsocket-java/issues/311

PR #321, opened 2017-06-19
Version bumps

yschimke:


yschimke:
I'll land in 24 hours, unless there is an objection.

phoad:
LGTM

PR #323, opened 2017-06-19
Reactor 3.1.0.M2

yschimke:
Reactor 3.1.x is not compatible with Reactor 3.0.x, upgrading since there is no point targetting the older version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 buildscript {
-    repositories { jcenter() }
+    repositories {
+        jcenter()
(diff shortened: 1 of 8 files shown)


yschimke:
@robertroeser @rdegnan @tmontgomery Given we are currently blocked on published releases, and reactor-core is about to release in next couple of weeks.  I'd like to land after a thorough review so I'm working on the latest codebase.

Any objections?

cc @tsachev

yschimke:
@phoad it is available in snapshot repos so you can republish those

PR #334, opened 2017-06-24
Terrible hack to fix websocket frame length handling

yschimke:


yschimke:
I tried fixing this two other ways

1) passing through a boolean into the frame/header encoding code and it quickly became really messy
2) dropping length from the frame and adding it back in for NettyDuplexConnection.  This made the code reliant on ByteBuf.readableBytes() which is currently frequently wrong.

Putting this up as the "mcdonalds suggestion" for now as it is currently out of spec.

cc @ragansa

ragansa:
👍👍

yschimke:
Main reason to land this is correctness trumps elegance.  What we currently have is not an implementation of the spec.

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/WebsocketDuplexConnection.java
```diff
@@ -43,12 +47,20 @@ public WebsocketDuplexConnection(NettyInbound in, NettyOutbound out, NettyContex
 
   @Override
   public Mono<Void> sendOne(Frame frame) {
-    return out.sendObject(new BinaryWebSocketFrame(frame.content())).then();
+    ByteBuf content = frame.content().skipBytes(3);
```
lexs: Please use `FrameHeaderFlyweight .FRAME_LENGTH_SIZE` instead of hardcoding

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/WebsocketDuplexConnection.java
```diff
@@ -43,12 +47,20 @@ public WebsocketDuplexConnection(NettyInbound in, NettyOutbound out, NettyContex
...
+    return out.sendObject(new BinaryWebSocketFrame(content)).then();
   }
 
   @Override
   public Flux<Frame> receive() {
-    return in.receive().map(buf -> Frame.from(buf.retain()));
+    return in.receive().map(buf -> {
+      buf.retain();
```
lexs: What does this return?

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/WebsocketDuplexConnection.java
```diff
@@ -43,12 +47,20 @@ public WebsocketDuplexConnection(NettyInbound in, NettyOutbound out, NettyContex
...
+      CompositeByteBuf composite = context.channel().alloc().compositeBuffer();
+      ByteBuf length = Unpooled.wrappedBuffer(new byte[3]);
+      FrameHeaderFlyweight.encodeLength(length, 0, buf.readableBytes());
+      composite.addComponents(true, length, buf);
```
lexs: Seems like the `buf.retain()` is better put here.

phoad: If we would like to keep the 'buf' alive, it seems better to call 'retain' prior to giving it to 'addComponents' function. 

https://netty.io/4.0/api/io/netty/buffer/CompositeByteBuf.html
"buffer - the ByteBuf to add. ReferenceCounted.release() ownership is transfered to this CompositeByteBuf."

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/WebsocketDuplexConnection.java
```diff
@@ -43,12 +55,18 @@ public WebsocketDuplexConnection(NettyInbound in, NettyOutbound out, NettyContex
 
   @Override
   public Mono<Void> sendOne(Frame frame) {
-    return out.sendObject(new BinaryWebSocketFrame(frame.content())).then();
+    return out.sendObject(new BinaryWebSocketFrame(frame.content().skipBytes(3))).then();
```
phoad: FRAME_LENGTH_SIZE instead of 3

Issue #336, opened 2017-06-29
Bug in setting MimeType


phoad:
RSocketFactory.java has this code

```
Frame setupFrame =
    Frame.Setup.from(
        flags,
        (int) ackTimeout.toMillis(),
        (int) ackTimeout.toMillis() * missedAcks,
        dataMineType,     <--
        metadataMimeType,  <--
        setupPayload);
```

though Frame.java has this factory method

```
public static Frame from(
    int flags,
    int keepaliveInterval,
    int maxLifetime,
    String metadataMimeType,   <--
    String dataMimeType,   <--
    Payload payload) {
```

the order of the parameters are in reverse.

PR #337, opened 2017-06-29
fix mimetypes

yschimke:


yschimke:
relates to https://github.com/rsocket/rsocket-java/issues/336

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
```diff
@@ -99,9 +100,9 @@ static ServerRSocketFactory receive() {
   }
 
   interface MimeType<T> {
-    T mimeType(String dataMineType, String metadataMimeType);
+    T mimeType(String metadataMimeType, String dataMimeType);
```
phoad: Even though it's good that switching the order of these two should be fixing the current users, as the users will not notice anything they can still assume that the first parameter is dataMimeType. 
So their code will still be 
```
mimeType("datamimetype", "metadatamimetype")
```
but in reality the first parameter is assigned to metadataMimeType now.

I mean, now we force them to have a bug without they notice. So I am in favor of not changing the order of these two.

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
```diff
@@ -182,15 +183,15 @@ public ClientRSocketFactory keepAliveMissedAcks(int missedAcks) {
     }
 
     @Override
-    public ClientRSocketFactory mimeType(String dataMineType, String metadataMimeType) {
-      this.dataMineType = dataMineType;
+    public ClientRSocketFactory mimeType(String dataMimeType, String metadataMimeType) {
```
phoad: Now the order again doesn't match to the declaration in the interface.

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
```diff
@@ -35,7 +35,8 @@
 import reactor.core.publisher.Mono;
 
 /** Factory for creating RSocket clients and servers. */
-public interface RSocketFactory {
+public interface
```
phoad: nit: is it in purpose that this line became two lines?

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
phoad: As all other places use this new order, I also believe it should be ..(metadataMimeType, dataMimeType)

yschimke: Thanks for spotting it twice!

PR #341, opened 2017-07-03
Resume support classes

yschimke:


yschimke:
This is the milestone 1/3 for resume support

1. support classes
2. Resume support without resume support  e.g. client and server track frames correctly, but don't reconnect, server send resume rejected.
3. Implement nice API for client and server resumption

rsocket-core/src/main/java/io/rsocket/resume/BufferStrategy.java
```diff
@@ -0,0 +1,15 @@
+package io.rsocket.resume;
+
+import io.rsocket.Frame;
+
+public interface BufferStrategy {
```
lexs: This name isn't great. Can this be called something like `ResumePositionCounter`?. It's not really something that does buffering.

lexs: Also a doc-comment or two.

rsocket-core/src/main/java/io/rsocket/resume/ResumeCache.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import reactor.core.publisher.Flux;
+
+public class ResumeCache {
+  private final BufferStrategy strategy;
+  private final int maxBufferSize;
+
+  private LinkedHashMap<Integer, Frame> frames = new LinkedHashMap<>();
```
lexs: `private final`?

rsocket-core/src/main/java/io/rsocket/resume/ResumeCache.java
```diff
@@ -0,0 +1,119 @@
...
+
+  public void sent(Frame frame) {
+    if (ResumeUtil.isTracked(frame)) {
+      frames.put(currentPosition, frame.retain());
+      bufferSize += strategy.cost(frame);
+
+      currentPosition += ResumeUtil.offset(frame);
+
+      if (frames.size() > maxBufferSize) {
+        Frame f = frames.remove(first(frames));
```
lexs: for the perf `frames.keySet().iterator().remove()` is likely a tiny bit faster but less readable so no need to change

rsocket-core/src/main/java/io/rsocket/resume/ResumeCache.java
```diff
@@ -0,0 +1,119 @@
...
+      Map.Entry<Integer, Frame> cachePosition = positions.next();
+
+      if (remotePosition < cachePosition.getKey()) {
+        resend.add(cachePosition.getValue().retain());
+      }
+
+      // TODO error handling
+    }
+
+    return Flux.fromIterable(resend);
```
lexs: If nobody observes this you'll leak the frames, probably not a problem of course

rsocket-core/src/main/java/io/rsocket/resume/ResumeToken.java
```diff
@@ -0,0 +1,50 @@
+package io.rsocket.resume;
+
+import io.netty.buffer.ByteBufUtil;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.UUID;
+
+public final class ResumeToken {
+  private final byte[] resumeToken;
```
lexs: Should this be stored as a bytebuffer? Seems like you'll save some allocations then.

rsocket-core/src/main/java/io/rsocket/resume/ResumeCache.java
yschimke: missing a .next(), but I take your point.  But don't think it helps enough to change it.

rsocket-core/src/main/java/io/rsocket/resume/ResumeCache.java
yschimke: good point, I'll think about it...

yschimke: Going to copy into what I believe is an unpooled subset copy.  Avoiding stealing from netty's io pooled buffers.

rsocket-core/src/main/java/io/rsocket/resume/ResumeToken.java
yschimke: Not sure it will, will reconsider as I test this.

PR #345, opened 2017-07-08
API support tweaks for external transports

yschimke:


yschimke:
Useful for non netty transports

rsocket-core/src/main/java/io/rsocket/transport/TransportHeaderAware.java
```diff
@@ -0,0 +1,12 @@
+package io.rsocket.transport;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+/**
+ * Extension interface to support Transports with headers at the transport layer,
+ * e.g. Websockets, Http2.
+ */
+public interface TransportHeaderAware {
```
yschimke: I'd love some input on ways to do this more cleanly.  My issue is that my client builds the transport using a string url, and then if it supports headers at the transport layer I want to apply those before connecting.

@robertroeser @rdegnan

yschimke: wondering whether this should return a copy of the Transport with these headers applied?

Issue #297, opened 2017-06-01
Refactor TCK test runner


yschimke:
Implement a junit test runner that generates tests at runtime such that right click in Intellij "Run All Tests" will run the TCK tests in IDE, and "./gradlew test" will run tests also.

Junit 4 https://stackoverflow.com/a/7404732/1542667
Junit 5 https://dzone.com/articles/junit-5-dynamic-tests-generate-tests-at-run-time

yschimke:
We should consider refactoring to a junit TestSuite

yschimke:
The goal here is to have them run as part of the normal development flow whether a developer uses an IDE or the command line.

yschimke:
This is an example parameterized test

```
import org.junit.Test;
import org.junit.runner.RunWith;
import org.junit.runners.Parameterized;

import java.io.File;
import java.util.ArrayList;
import java.util.Collection;

import static org.junit.Assert.assertEquals;

@RunWith(Parameterized.class)
public class TestX {
    private File file;

    public TestX(File file) {
        this.file = file;
    }

    @Test
    public void test1() throws Exception {
        assertEquals("path1", this.file.getName());
    }

    @Parameterized.Parameters
    public static Collection<Object[]> data() {
        // load the files as you want
        Object[] fileArg1 = new Object[]{new File("path1")};
        Object[] fileArg2 = new Object[]{new File("path2")};

        Collection<Object[]> data = new ArrayList<Object[]>();
        data.add(fileArg1);
        data.add(fileArg2);
        return data;
    }
}
```

yschimke:
https://github.com/rsocket/rsocket-java/pull/347

PR #347, opened 2017-07-10
Refactor TCK test runner #297

junaidkhalid:
Replaced TestMain.java and Main.java with TckTest.java. Unlike TestMain.java, TckTest.java is based on juint and does not require main method. 

The assumption is that all the tests files will reside in rsocket-java/rsocket-tck-drivers/src/test/resources/ . These files will be in pair (client<NAME>.txt and server<NAME>.txt). 

What is missing:
1) Currently the assert check is on the output of the test ( assertEquals(result, JavaClientDriver.TestResult.PASS)). I will update the ConsoleUtils checks with assert statements in the later commit. (FIXED)
2) make sure "./gradlew test" works (FIXED)
3) remove un-used code (FIXED)


yschimke:
That is fantastic, thanks for fixing that.  n.b. I think you need to move the TckTest.java from src/main/java to src/test/java to get it to run automatically in gradle.  But it's running perfectly for me in an IDE.

rsocket-tck-drivers/src/main/java/io/rsocket/tckdrivers/test/TckTest.java
```diff
@@ -0,0 +1,149 @@
...
+    for (int i = 0; i < listOfFiles.length; i++) {
+      File file = listOfFiles[i];
+      if (file.isFile() && file.getName().startsWith(clientPrefix)) {
+        String serverFileName = serverPrefix + file.getName().replaceFirst(clientPrefix, "");
+
+        File f = new File(path + serverFileName);
+        if (f.exists() && !f.isDirectory()) {
+
+          //starting a server
+          ServerThread st = new ServerThread(currentPort, path + serverFileName);
```
yschimke: I think more of this logic could be moved behind the Test lifecycle.  As it is all threads start when you build the list of tests, this happens even if you have selected only a single test to run.

rsocket-tck-drivers/src/main/java/io/rsocket/tckdrivers/test/TckTest.java
```diff
@@ -0,0 +1,149 @@
...
+          st.start();
+          st.awaitStart();
+
+          try {
+            RSocket client = createClient(new URI("tcp://" + hostname + ":" + currentPort + "/rs"));
+            JavaClientDriver jd = new JavaClientDriver(() -> client);
+
+            BufferedReader reader = new BufferedReader(new FileReader(file));
```
yschimke: Not essential for this PR, but this method is quite big and could be improved by refactoring, e.g. extract methods, introduce a TckIndividualTest instead of String name, List<String> test

rsocket-tck-drivers/src/main/java/io/rsocket/tckdrivers/test/TckTest.java
```diff
@@ -0,0 +1,149 @@
...
+  private List<String> test; // test instructions/commands
+
+  public TckTest(String name, JavaClientDriver jd, List<String> test) {
+    this.name = name;
+    this.jd = jd;
+    this.test = test;
+  }
+
+  /** Runs the test. */
+  @Test
```
yschimke: add a timeout here, when I made a test fail it caused the whole test suite to hang.

image

rsocket-tck-drivers/src/main/java/io/rsocket/tckdrivers/test/TckTest.java
junaidkhalid: I have fixed all the other issues. I will generate a PR for this

junaidkhalid:
also added filename to the test name (filename does not include "server/client" prefix)
<INDEX>: <TESTNAME> (<FILENAME>)

PR #351, opened 2017-07-11
Refactor TckTest(added TckIndividualTest and ExtractTests)

junaidkhalid:
Refactor TckTest:
1) Introduced TckIndividualTest 
2) Added ExtractTests method

### DIFF
--- a/rsocket-tck-drivers/src/test/java/io/rsocket/tckdrivers/TckTest.java
+++ b/rsocket-tck-drivers/src/test/java/io/rsocket/tckdrivers/TckTest.java
@@ -96,10 +96,10 @@ public class TckTest {
   /**
    * A function that parses the file and extract the individual tests
    *
-   * @param file: file to read as input.
+   * @param file The file to read as input.
    * @return a list of individual tests. Each individual test is also a list of String.
    */
-  private static List<List<String>> ExtractTests(File file) throws Exception {
+  private static List<List<String>> extractTests(File file) throws Exception {
 
     BufferedReader reader = new BufferedReader(new FileReader(file));
     List<List<String>> tests = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


rsocket-tck-drivers/src/test/java/io/rsocket/tckdrivers/TckTest.java
```diff
@@ -79,19 +87,50 @@ public void TckTestRunner() {
...
+      jd.runTest(this.tckTest.test.subList(1, this.tckTest.test.size()), this.tckTest.name);
     } catch (Exception e) {
       e.printStackTrace();
     }
   }
 
   /**
+   * A function that parses the file and extract the individual tests
+   *
+   * @param file: file to read as input.
```
yschimke: remove colon

`@param file The file to read as input.`

rsocket-tck-drivers/src/test/java/io/rsocket/tckdrivers/TckTest.java
```diff
@@ -79,19 +87,50 @@ public void TckTestRunner() {
...
+   * @return a list of individual tests. Each individual test is also a list of String.
+   */
+  private static List<List<String>> ExtractTests(File file) throws Exception {
```
yschimke: lower case e for the method name, extractTests.

yschimke: If you change this method to return List<TckIndividualTest> then we are moving towards a model where the input could be a library and calling a function instead of assuming there is a file to read from.

rsocket-tck-drivers/src/test/java/io/rsocket/tckdrivers/TckTest.java
```diff
@@ -79,19 +87,50 @@ public void TckTestRunner() {
...
+
+    BufferedReader reader = new BufferedReader(new FileReader(file));
```
yschimke: Suggestion: why not try using Reactor library here?  As a *optional* programming exercise, what does this method look like if you refactor it to use methods that only use reactor types as the reponses e.g. Flux<TckIndividualTest>, Flux<File>, Flux<String>

PR #354, opened 2017-07-14
Allow frames without metadata

yschimke:
Currently >0 length metadata is present (flag in frame), while 0 length is missing.

This changes it to 
null: not present
0 length = ""
anything else is the content.


robertroeser:
I'd really prefer that we used an empty buffer instead of null, and made sure the length was set to zero, then using null.

yschimke:
@robertroeser but they are fundamentally 3 different values

null - not present
present and ""
present and non zero length

It seems against spec and generally wrong to munge null and "" together.  It didn't work in early databases :)

robertroeser:
I think we can all agree it's better to avoid null pointer exceptions. ReactiveStreams also doesn't allow null values, you're get an exception. Because of this the user will have to null check there code. I don't really see the difference between byte[0] vs null for a user other than one will cause null pointer exceptions.

yschimke:
@robertroeser I have it working with all 3 cases supported, this also fixes the bug when receiving the frames with empty metadata that rsocket-js was sending.

However before reviewing, I am going to address your concerns, I'll make the public APIs not return null, instead use the hasMetadata check so naive usage is exactly as before, but if you care about null vs "" you can check the flag on Frame or Payload.

robertroeser:
I like the idea of adding hasMetadata flag

yschimke:
@robertroeser the public APIs in the PR use hasMetadata, and getMetadata returns byteBuf(0).  But internally I'm using nulls.  So I think this is close.  I've added exceptions so that if you provide non empty metadata and don't indicate the flag then it fails.

PR #355, opened 2017-07-15
utf8 helpers on payload

yschimke:
Protocol already favours UTF-8, so  having these helpers saves a lot of boilerplate


yschimke:
I can inline this back to an external utility trivially, so landing.

PR #379, opened 2017-08-28
[WIP] Context exploration

yschimke:
This is a prototype (currently) of supporting Context from Reactor-Core.  This seems very similar to Finagle's Context which allows request metadata to be associated with a pipeline of operations in a somewhat similar manner to ThreadLocal in a synchronous program.

Typical uses
 - per request auth tokens
 - Zipkin tracing context
 - Naming overrides

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,6 +67,8 @@ subprojects {
         compile 'org.reactivestreams:reactive-streams:1.0.0'
         compile 'org.slf4j:slf4j-api:1.7.25'
         compile 'com.google.code.findbugs:jsr305:3.0.0'
+        compile 'com.fasterxml.jackson.core:jackson-core:2.8.8'
+        compile 'com.fasterxml.jackson.core:jackson-databind:2.8.8'
 
         testCompile 'junit:junit:4.12'
         testCompile 'org.mockito:mockito-core:1.10.19'
--- a/rsocket-core/src/main/java/io/rsocket/BaseContextEncoder.java
+++ b/rsocket-core/src/main/java/io/rsocket/BaseContextEncoder.java
@@ -4,9 +4,11 @@ import io.rsocket.util.PayloadImpl;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import reactor.util.context.Context;
+import reactor.util.function.Tuple2;
+import reactor.util.function.Tuples;
 
 public abstract class BaseContextEncoder implements ContextEncoder {
-  @Override public Payload apply(Payload payload, Context context) {
+  @Override public Payload encode(Payload payload, Context context) {
     try {
       if (!payload.hasMetadata()) {
         ByteBuffer contextMetadata = encode(context);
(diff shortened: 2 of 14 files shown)


yschimke:
cc @tsachev

build.gradle
```diff
@@ -67,6 +67,8 @@ subprojects {
         compile 'org.reactivestreams:reactive-streams:1.0.0'
         compile 'org.slf4j:slf4j-api:1.7.25'
         compile 'com.google.code.findbugs:jsr305:3.0.0'
+        compile 'com.fasterxml.jackson.core:jackson-core:2.8.8'
```
robertroeser: not thrilled about the idea of making jackson a dependency of rsocket

yschimke: Yep - I will pull that out and not provide any implementation apart from NONE in core.

yschimke:
@robertroeser We are bought into spring reactor-core pretty heavily.  This is also a well trodden path e.g. finagle has the same concept and its used a lot.

I will look if I can implement cleanly as an interceptor, in which case there will be zero overhead by default.  and implementation may be cleaner.   If I can't it probably suggests interceptors are not designed correctly either.

But after having implemented a few "real" apps, clients & servers with rsocket-java, I think this is something that should be trivial to setup and use.  The really awkward thing about metadata in our current design is that to use it currently you have to make a choice

1) Always treat things as Flux<Payload>
2) Use higher level types but always use Flux<Pair<MyMetadata, MyResponse>>
3) Drop the metadata after initial routing decisions

robertroeser:
@yschimke RSocket was never met to be used by everyone directly. The original intent was to put a layer on top of it to make it easier to use. We have internally experimented with a protobuf code generation that exposed a user payload and metadata. The layer we created did several things 1) It handled payload serialization, 2) It provided easy access to metadata, and 3) It would route your message to the correct method.  I think the best thing to do would be to start a new project and create a layer on top like analogous to GRPC or Thrift. We could make the serialization pluggable if we wanted. I don't think you could use protobuf at Facebook?

I can go ahead and start the project. I think it would be good to have this include the C++, and JS versions of RSocket.
I think something like <RPC Layer> -> RSocket -> <Transport>  would be good. 

Regarding Reactor-core- We're using reactor-core because it's an implementation of reactive-streams. The implementations we made was buggy, and the stuff is hard to get right. We were going to use RXJava2, but RXNetty was never ported over to use it.  Additionally, there was conscious decision to optimize RXJava2 for Android where as reactor-core (more less written by the same guy) was optimized for Java8. Combined with the fact they are actively working on reactor-netty it seemed a good choice. We talked about shadding reactor-core but decided against for now because of operator fusion - it wouldn't work with reactor-netty if we shaded it.

yschimke:
Yep, I have no issue with reactor-core.  It's been nice to use, and I'm really happy with java 8 as a base assumption.  The only complaint I have is the lack of pre Android O support. :)

In fact this PR is suggesting that in addition to being able to interoperate between different reactive libraries, that we should support (as optional) reactor-core features when useful.  That's what I'm suggesting for supporting Context.  Define this as our supported mechanism for propagating request/subscription context including over network boundaries.

I'll pull out any Jackson dependant code into a separate module, and make it an optional feature implemented cleanly by interceptors.  But I think this feature should be a part of rsocket-java.  Not purely a layer on top.

I agree with and am happy to hear about any plans to build databinding libraries on top.  If we had something like Retrofit, I'd be a happy camper.

PR #382, opened 2017-09-12
update build dependencies

qweek:
plugins: com.jfrog.artifactory 4.5.2
dependencies: jsr305 3.0.2, aeron 1.4.1, spectator 0.57.1, hdrHistogram 2.1.9, jackson 2.9.1, commons-lang3 3.6, airline 0.8, rxjava 2.1.3, mockito 2.10.0, jmh 1.19


qweek:
Please check #383 first

yschimke:
@qweek nice! Thanks for the general cleanup also.

robertroeser:
@qweek @yschimke Why are the version the version not inline with the includes? Can you put this back please.

yschimke:
@robertroeser I don't have a strong opinion on this, but this is pretty common for gradle projects.

To be clear you mean, you would prefer to see

compile "io.netty:netty-transport-native-epoll:4.1.15.Final"

robertroeser:
Yes because you don't have to scroll to find what version you are using.

PR #394, opened 2017-09-16
redundant condition

qweek:
Condition 'metadata != null' is always 'true' when reached


yschimke:
@qweek Thanks for the cleanup! As these are getting smaller, could you batch up any further ones into a single PR and I'll land after a week or so.

Keep 'em coming!

PR #402, opened 2017-10-03
Feature/gradle 4.2

qweek:
upgrade gradle to version 4.2
change distribution type to binary


yschimke:
Can you submit this PR unstacked.  And why bin over all? Intellij often prompts to use the src distribution.

qweek:
I don't see any difference: autocomplete for gradle works fine, sources decompiled
I can revert distribution type if you have any loss in functionality

yschimke:
Can you refactor without splitting up the gradle file?

PR #404, opened 2017-10-03
Bugfix/fix warnings

qweek:
warning: [cast] redundant cast to int
warning: [fallthrough] possible fall-through into case
warning: [serial] serializable class has no definition of serialVersionUID
warning: [try] auto-closeable resource has a member method close() that could throw InterruptedException


yschimke:
Can you refactor without splitting up the gradle file?

PR #409, opened 2017-10-05
Feature/fix lint warnings

qweek:
warning: [cast] redundant cast to int
warning: [fallthrough] possible fall-through into case
warning: [serial] serializable class has no definition of serialVersionUID
warning: [try] auto-closeable resource AeronChannel has a member method close() that could throw InterruptedException


qweek:
Refactor #404

PR #413, opened 2017-10-07
java 9 migration

qweek:
add jdk 9 to travis
compile with java 9

#### Fix dependency error [stackoverflow/java-9-how-to-get-access-to-javax-annotation-resource-at-run-time](https://stackoverflow.com/questions/46502001/java-9-how-to-get-access-to-javax-annotation-resource-at-run-time)
java.xml.ws.annotation module is deprecated for removal
```
error: cannot find symbol
@javax.annotation.Generated(
                 ^
  symbol:   class Generated
  location: package javax.annotation
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,13 +44,9 @@ subprojects {
     group = 'io.rsocket'
     version = mavenversion
 
-    compileJava {
-        sourceCompatibility = 1.8
-        targetCompatibility = 1.8
-        options.compilerArgs << '-Xlint:all,-overloads,-rawtypes,-unchecked'
-    }
-
     ext {
+        isJava9 = System.getenv('TRAVIS_JDK_VERSION') == 'oraclejdk9'
+        jdkVersion = isJava9 ? 1.9 : 1.8
         // common
         jsr305Version = '3.0.2'
         reactorVersion = '3.1.0.RELEASE'
@@ -59,6 +55,7 @@ subprojects {
         slf4jVersion = '1.7.25'
         // aeron
         aeronVersion = '1.4.1'
+        javaxAnnotationVersion = '1.3.1'
         // netty
         reactorNettyVersion = '0.7.0.M2'
         // spectator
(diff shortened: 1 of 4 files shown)


qweek:
#### Known issue [netty/netty/#7254](https://github.com/netty/netty/issues/7254)
Illegal reflective access warnings when running Netty with Java 9
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by io.netty.util.internal.ReflectionUtil (file:netty-common-4.1.16.Final.jar) to constructor java.nio.DirectByteBuffer(long,int)
WARNING: Please consider reporting this to the maintainers of io.netty.util.internal.ReflectionUtil
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

qweek:
#### Possible problems with modularization [netty/netty/#7266](https://github.com/netty/netty/issues/7266) [netty/netty/#7268](https://github.com/netty/netty/pull/7268)
Unable to use Netty native transports in modularized Java 9 projects

Automatic module name for the epoll transport would be: `netty.transport.native.epoll`
But `native` is reserved keyword in java

qweek:
Before merge, someone should check that `publish` step will be executed once

build.gradle
```diff
@@ -76,6 +73,12 @@ subprojects {
         jmhVersion = '1.19'
     }
 
+    compileJava {
+        sourceCompatibility = jdkVersion
+        targetCompatibility = jdkVersion
```
yschimke: Shouldn't this stay 1.8 in all cases?

qweek: jdkVersion 1.9 used only for oraclejdk9 image on Travis
Why you want to use 1.8 in this case?

yschimke: I don't have any objections to compiling and running with JDK 9, that's a useful test.

But I don't think we want JDK 9 features used in source, or JDK9 specific jars.  Was this the aim?

qweek: Got it, no it's just compatibility test.

yschimke:
If you want it as a compatibility test and that requires changing the source or target, then update the publish step as you suggest to only publish for 1.8 

https://github.com/rsocket/rsocket-java/blob/1.0.x/gradle/buildViaTravis.sh

robertroeser:
why are we moving to java 9 before we release 1.0? I'm closing this.

qweek:
We are not moving, just testing
Someone might want to use RSocket library in Java 9 project

PR #415, opened 2017-10-08
fix client unsubscribe

qweek:
'ConcurrentHashMap<AeronSocketAddress, Subscription>' may not contain objects of type 'Long'


qweek:
Please also check error handling ('throw' inside 'finally' block)
```
AeronClientChannelConnector:141
AeronChannelServer:220
```
Possible solution [stackoverflow/how-to-avoid-throw-clause-in-finally-block](https://stackoverflow.com/questions/28187304/how-to-avoid-throw-clause-in-finally-block)
```
private int poll() {
  int poll;
  Exception exception = null;
  try {
    poll = managementSubscription.poll(fragmentAssembler, 4096);
  } catch (Exception e) {
    exception = e;
    throw e;
  } finally {
    if (running) {
      boolean execute = eventLoop.execute(this::poll);
      if (!execute) {
        running = false;
        IllegalStateException suppressed =
            new IllegalStateException("unable to keep polling, eventLoop rejection");
        if (exception == null) {
          throw suppressed;
        }
        exception.addSuppressed(suppressed);
      }
    }
  }

  return poll;
}
```

PR #420, opened 2017-10-09
remove ignores

yschimke:


yschimke:
@robertroeser I'd like to discuss the availability aspect with you.  To my mind, an unavailable connection should result in visible exceptions through the normal requestResponse error handling, not ignored or sent to the connection error handler (only).

robertroeser:
@yschimke  we can try throwing errors. I think I did this before and it lead to some issues where the side effects were causing errors - or we were sending frames that weren't necessary without checking if the connection was still open. It might be better now.

yschimke:
@robertroeser any blocker to landing this?

PR #421, opened 2017-10-10
Lease support

mostroverkhov:
Integrated into RSocketFactory in adhoc manner
Missing end-to-end tests, but there are 2 runnable snippets in `rsocket-examples`
Connection setup flow with LEASE enabled implemented according to rsocket/rsocket#242


yschimke:
@rdegnan @robertroeser I think you know the leasing a lot better than I do.

I'll review in the next few days if it doesn't get a review sooner.

rsocket-core/src/main/java/io/rsocket/lease/LeaseImpl.java
```diff
@@ -76,4 +76,24 @@ public String toString() {
         + expiry
         + '}';
   }
+
+  @Override
+  public boolean equals(Object o) {
```
mostroverkhov: It was mistake to add  eq/hashcode for Lease. Going to remove this from PR

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
```diff
@@ -168,6 +179,7 @@ private synchronized void cleanUpChannelProcessors() {
           return metadataPush(new PayloadImpl(frame));
         case LEASE:
           // Lease must not be received here as this is the server end of the socket which sends leases.
```
mostroverkhov: Guess this comment is not valid as Connection.receive() is constrained as Unicast so has to be handled here anyways?

yschimke:
@mostroverkhov If I added support for the rsocket-cli to use leases, any suggestions for the client and server config to use?

--lease 100/60s

For my own edumacation - I'd like to review this in the context of runnable code as I haven't used RSocket leases yet.

rsocket-core/src/main/java/io/rsocket/LeaseSocketAcceptor.java
```diff
@@ -0,0 +1,8 @@
+package io.rsocket;
+
+import reactor.core.publisher.Mono;
+
+public interface LeaseSocketAcceptor {
```
robertroeser: Last time leasing was implemented it was leaky, and there was a bunch of different "leasing" acceptors and sockets. I don't want that to happen again. I don't want special sockets for leasing. The person using it shoudn't know that it's happening other than a flag in setup.

mostroverkhov:
@yschimke I haven't used cli yet (will give it a try tonight), but if cli acts only as Client Requester, for now it probably should:
* be able to request Lease support in Setup frame
* show notification when lease received from peer
* show notification when lease updated as request is sent

RSocketServer should be constructed with lease support enabled, and send leases when application thinks its appropriate - not related to cli itself. 

Things will be more clear after I address and discuss requested changes, later today.

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
```diff
@@ -168,6 +179,7 @@ private synchronized void cleanUpChannelProcessors() {
...
+          leaseConsumer.ifPresent(consumer -> consumer.accept(frame));
```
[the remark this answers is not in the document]
mostroverkhov: I agree with naming according to spec, doing so helped me a lot while making this pull. Probably comment above made more confusion that clarification, so I'll explain what motivated me to make this change. Each side of connection needs to listen for LEASE frames on 0 stream. Atm, `ClientServerMultiplexer` forwards both Client and Zero incoming frames (except SETUP) to `clientConnection()`, and demands there is only one subscriber for that connection. So either original `clientConnection()` is wrapped to listen for incoming LEASE frames as e.g. `conn.receive().doOnNext(frame)`, or pass listener directly to RSocketClient/Server (in current terminology), as sole subscribers to connection. I selected second option.

rsocket-core/src/main/java/io/rsocket/lease/LeaseGranter.java
```diff
@@ -0,0 +1,156 @@
...
+        LeaseManager requesterLeaseManager,
+        LeaseManager responderLeaseManager,
+        Consumer<Throwable> errorConsumer) {
+      super(senderConnection, requesterLeaseManager, responderLeaseManager, errorConsumer);
+    }
+
+    @Override
+    Consumer<Frame> grantedLeasesReceiver() {
+      return f -> {
+        synchronized (lock) {
```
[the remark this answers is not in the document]
mostroverkhov: Probably yes

rsocket-core/src/main/java/io/rsocket/LeaseSocketAcceptor.java
mostroverkhov: Users of requester RSocket would need to know when lease is granted/withdrawn, and how many credits are available. Otherwise users would blindly retry requests failed due to missing lease, which is effectively polling. Also, It can be holder for "simple" RSocket, and whatever factory wants to expose to user (Context). Or raw socket is enough?

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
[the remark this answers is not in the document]
robertroeser: Each side, the connecting side, and the receiving side has client and server. The server should not except frame.

rsocket-core/src/main/java/io/rsocket/LeaseSocketAcceptor.java
robertroeser: There is availability in the RSocket - you need to update that to 0 when there are no leases, and then it's up to the client to decided with to send the request or not. For instance the load balancer uses availability wither or not to send the request to that RSocket.

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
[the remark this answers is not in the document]
robertroeser: I suggest for this pull request you rename the RSocketClient and the RSocketServer to RSocketRequester and RSocketResponser to clarify things.

rsocket-core/src/main/java/io/rsocket/LeaseSocketAcceptor.java
robertroeser: Actually what we ended up doing was time as make availability a ratio of available_leases/total_leases  - servers with more leases should be more available

yschimke:
@mostroverkhov the CLI is pretty crude (at level of common curl type requests).  https://github.com/rsocket/rsocket-cli

It can run as a client (including passive) or server.  --debug should let you see the lease frames, and confirm you see the right behaviour over a few minutes etc.  You can run it straight out of a clean git directory.

rsocket-core/src/main/java/io/rsocket/lease/NoLeaseException.java
```diff
@@ -0,0 +1,18 @@
+package io.rsocket.lease;
+
+import io.rsocket.exceptions.RejectedException;
+import javax.annotation.Nonnull;
+
+public class NoLeaseException extends RejectedException {
```
yschimke: add serialVersionUID

rsocket-core/src/main/java/io/rsocket/lease/Lease.java
```diff
@@ -69,4 +69,9 @@ default boolean isExpired() {
   default boolean isExpired(long now) {
     return now > expiry();
   }
+
+  /** Checks if the lease has not expired and there are allowed requests available */
```
yschimke: nit: gradle complains about the missing @return

rsocket-core/src/main/java/io/rsocket/LeaseSocketAcceptor.java
yschimke: With this change, would we also collapse RSocketFactory.leaseAcceptor and acceptor?  This bit confused me when trying to use the API.

mostroverkhov: @yschimke I am reworking this atm based on Robert comments -  there will be no leaseAcceptor, and lease changes will be reflected in RSocket availability. From user point of view  `Client/ServerRSocketFactory` will be augmented with `enableLease(something)`. Hopefully will come up with runnable example by the end of weekend

rsocket-core/src/main/java/io/rsocket/lease/Lease.java
mostroverkhov: I will update javadocs with detailed explanations once apis and implementations are agreed

rsocket-core/src/main/java/io/rsocket/lease/LeaseRSocket.java
```diff
@@ -56,7 +56,7 @@ public double availability() {
         : 0.0;
```
robertroeser: Can lease.getStartingAllowedRequests() every equal zero? Might want to check for that.

mostroverkhov: lease.startingNumberOfRequests is set in constructor and never changes, so only case when it is 0 is when lease.numberOfRequests is 0, but then lease.isValid is false

rsocket-core/src/main/java/io/rsocket/lease/NoLeaseException.java
mostroverkhov: done

rsocket-core/src/main/java/io/rsocket/lease/LeaseRSocketRef.java
```diff
@@ -0,0 +1,106 @@
...
+
+  public double availability() {
+    return rsocket.availability();
+  }
+
+  public boolean isClosed() {
+    return isClosed;
+  }
+
+  public void grantLease(int numberOfRequests, long timeToLive, @Nullable ByteBuffer metadata) {
```
mostroverkhov: metadata on received lease is not exposed yet. Should it be ByteBuffer or whole Lease?

mostroverkhov:
@yschimke @robertroeser Thinking about this more I see that lease (and many other features relying on 0 connection) can be implemented on  DuplexConnection level with interceptors only. It will be more compact   and less invasive also. While investigating I prefer to close this PR.

PR #426, opened 2017-10-14
migration to junit 5 

qweek:


yschimke:
I've tried this on other projects and it's too early, let's wait for other projects to flush out issues first.  Generally works for one person with a particular IDE and specific plugin versions, fails for the next person.

yschimke:
Specifically good evidence would be examples of large, stable, projects adopting. More than one of Netty, reactor-core, Retrofit etc.

qweek:
No problems

robertroeser:
@qweek @yschimke I think before we switch JUnit we should switch the tests to use Reactor's Step Verifier.  The current tests are before we switched to reactor-core. Additionally, it would be nice if rsocket local was moved into rsocket core, and then all the various and assorted tests against RSocket were done with that in the core package. Right now tests are scattered about the various packages and that doesn't seem like a good idea.

yschimke:
@robertroeser a big +1 to moving the local transport into core.  It also clearly defines it isn't a fake mock transport, but should be the reference transport.

PR #430, opened 2017-10-17
Minor bug fixes

mostroverkhov:
fixed bugs in Lease frame and PayloadImpl

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/Frame.java
+++ b/rsocket-core/src/main/java/io/rsocket/Frame.java
@@ -406,7 +406,8 @@ public class Frame implements ByteBufHolder {
       final Frame frame = RECYCLER.get();
       frame.content =
           ByteBufAllocator.DEFAULT.buffer(
-              LeaseFrameFlyweight.computeFrameLength(metadata.readableBytes()));
+              LeaseFrameFlyweight.computeFrameLength(
+                  metadata != null ? metadata.readableBytes() : 0));
       frame.content.writerIndex(
           LeaseFrameFlyweight.encode(frame.content, ttl, numberOfRequests, metadata));
       return frame;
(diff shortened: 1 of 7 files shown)


rsocket-core/src/main/java/io/rsocket/Frame.java
```diff
@@ -406,7 +406,8 @@ public static Frame from(int ttl, int numberOfRequests, ByteBuf metadata) {
       final Frame frame = RECYCLER.get();
       frame.content =
           ByteBufAllocator.DEFAULT.buffer(
-              LeaseFrameFlyweight.computeFrameLength(metadata.readableBytes()));
+              LeaseFrameFlyweight.computeFrameLength(
```
yschimke: Lease frame must always have metadata, but it can be 0 length.  Metadata is quite awkward on other frames, when it is optional, or can be 0 length, etc.

Why make it optional here?

yschimke: And if we want this, can we keep rolling out @ Nullable annotations to make it clear.  But I think requiring it is safer here.

mostroverkhov: spec states otherwise, as i read it: Lease frame , `(M)etadata: Metadata present` implies it can be absent

yschimke: @mostroverkhov I think you are right, that flag is there, I missed that.  So maybe just add @Nullable and change LeaseFrameFlyweight.computeFrameLength to take @Nullable Integer instead of int.

robertroeser: don't really want to take an object when we could take a primitive do we?

mostroverkhov: It will be boxed eventually in `FrameHeaderFlyweight.computeMetadataLength` anyways. I can apply change proposed by Yuri here, or revert for now and open separate issue, thus will have separate PR per bug (should have done that from start i think)

yschimke: So metadata not present is different from empty - which is why the Integer.

Issue #286, opened 2017-05-18
Add infer @Nullable checks to CI build


yschimke:
Should we consider adopting some of the Square practices here?

https://medium.com/square-corner-blog/rolling-out-nullable-42dd823fbd89
https://medium.com/square-corner-blog/non-null-is-the-default-58ffc0bb9111

@Nullable, and @ParametersAreNonnullByDefault

lexs:
Strongly in favor. Internally we (Facebook) have our entire codebase be non-null by default using https://github.com/facebook/infer to force `@Nullable` where it's needed.

robertroeser:
@lexs Does infer work with the javax.annotation.Nullable? If it did that would be great because we wouldn't have to have a dependency on Guava.

There appears to be a gradle plugin for this:
https://github.com/uber-common/infer-plugin

This seems like a good idea to add this to the build.

yschimke:
I'm trying this out at the moment

yschimke:
The plugin seems to have version incompatibilities with newer gradle versions, but infer docs work fine externally for now.

https://github.com/rsocket/rsocket-java/pull/288

PR #433, opened 2017-10-22
Feature/infer check

qweek:
check build with infer


qweek:
Uber plugin for Infer (#286) is not compatible with new versions
Dependencies installed on Travis in 23 minutes
I think it's better to run Infer in a separate branch

.travis.yml
```diff
@@ -10,12 +10,22 @@ addons:
 
 dist: trusty
 #group: edge
-sudo: false
+sudo: true
 # as per http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
 
+install:
+  - chmod +x gradle/installViaTravis.sh
```
yschimke: shouldn't this be handled when you add the file to git?

.travis.yml
```diff
@@ -10,12 +10,22 @@ addons:
...
 script: gradle/buildViaTravis.sh
 
+after_script:
+  - gradle clean
+  - infer capture -- gradle test
+  - infer analyze -a checkers --print-active-checkers --quiet && cat infer-out/bugs.txt
+  - infer --no-default-checkers --bufferoverrun --compute-analytics --eradicate --repeated-calls --suggest-nullable --tracing --print-active-checkers --quiet && cat infer-out/bugs.txt
+
 # cache between builds
 cache:
```
yschimke: Can you add the infer builds to the cache and check before building them?

.travis.yml
```diff
@@ -10,12 +10,22 @@ addons:
 
 dist: trusty
 #group: edge
-sudo: false
+sudo: true
 # as per http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
```
yschimke: this contradicts the "sudo: true" here.

.travis.yml
qweek: Ok, I will check how it works with cache

robertroeser:
@qweek @yschimke what's outstanding for this request? The caching stuff?

qweek:
41 min without cache
5 min 47 sec with cache

yschimke:
I restarted the build - will land if its running in 5 minutes, not 41 :)

robertroeser:
How often is this run? Is it 5 minutes locally? The local build is pretty snappy now, it would nice to keep it that way.

yschimke:
5 minutes would be similar to existing travis builds - https://travis-ci.org/rsocket/rsocket-java/builds, but I haven't seen caching working yet.  I keep seeing 41 minute builds.

This won't run locally without you running extra steps, so shouldn't affect dev experience.

n.b. I love infer, thanks @qweek for adding this, assuming we get it faster than 40 minutes.

robertroeser:
@qweek @yschimke  What's the status of this? Does the caching work now?

yschimke:
I'd love to see this in place, but I didn't see caching work which I think makes it a non starter.

PR #437, opened 2017-10-26
netty tcp driver hangs on large channels

robertroeser:
When streaming data from a channel RSocket would hang. Moved the subscription to a single channel in the DuplexConnection. For some reason while this fixes things for TCP with WebSocket it breaks other tests. This doesn't show up until you start streaming more than 20k or so items.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,7 +107,7 @@ subprojects {
 
     dependencies {
         compile "io.projectreactor:reactor-core:3.1.0.RELEASE"
-        compile "io.netty:netty-buffer:4.1.15.Final"
+        compile "io.netty:netty-buffer:4.1.16.Final"
         compile "org.reactivestreams:reactive-streams:1.0.1"
         compile "org.slf4j:slf4j-api:1.7.25"
         compile "com.google.code.findbugs:jsr305:3.0.2"
--- a/rsocket-test/src/main/java/io/rsocket/test/BaseClientServerTest.java
+++ b/rsocket-test/src/main/java/io/rsocket/test/BaseClientServerTest.java
@@ -17,6 +17,7 @@
 package io.rsocket.test;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 import io.rsocket.Payload;
 import io.rsocket.util.PayloadImpl;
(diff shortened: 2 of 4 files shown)


smaldini:
We should add rsocket in some of our test coverage directly in Reactor netty. This looks weird. Also looking at the usage it prompted me that further review could help for some optimization. Cc @violetagg

robertroeser:
@smaldini yep it's weird :)   Something is not lining up correctly. Ideally we'd like to be able to call out.send(Publisher<ByteBuf>) and just stream in data. It looks like it only flushes on when the stream completes?  Problem is we don't complete until the socket closes.  I would be great figure out how to get that to work as it would really simplify the code.

smaldini:
@robertroeser you can currently use ```NettyOutbound.options(NettyPipeline.SendOptions::flushOnEach)``` if you want to flus on each frame, that optimizes a bit the flow vs custom delimiting via multiple send.

smaldini:
Another thing, if you were to use UnicastProcessor you must remember that it obeys by default to RS and forbids concurrent onNext. In rx or reactor you can solve that with serialization, but I see you wanted to include MPSC queues from JCTool which you can pass to UnicastProcessor via create(Queue) factory.

robertroeser:
@smaldini Just tried it -seems to look up and not emit anything. Here's how we'd hope the class should be - doesn't seem to work though
```
public class NettyDuplexConnection implements DuplexConnection {
  private final NettyInbound in;
  private final NettyOutbound out;
  private final NettyContext context;
  private final MonoProcessor<Void> onClose;

  public NettyDuplexConnection(NettyInbound in, NettyOutbound out, NettyContext context) {
    this.in = in;
    this.out = out;
    this.out.options(NettyPipeline.SendOptions::flushOnEach);
    this.context = context;
    this.onClose = MonoProcessor.create();

    context.onClose(onClose::onComplete);
    this.onClose
        .doFinally(
            s -> {
              this.context.dispose();
              this.context.channel().close();
            })
        .subscribe();
  }

  @Override
  public Mono<Void> send(Publisher<Frame> frames) {
    **return out.send(Flux.from(frames).map(Frame::content)).then();**
  }

  @Override
  public Flux<Frame> receive() {
    return in.receive().map(buf -> Frame.from(buf.retain()));
  }

  @Override
  public Mono<Void> close() {
    return Mono.fromRunnable(onClose::onComplete);
  }

  @Override
  public Mono<Void> onClose() {
    return onClose;
  }

  @Override
  public double availability() {
    return onClose.isTerminated() ? 0.0 : 1.0;
  }
}
```

PR #438, opened 2017-10-30
RSocket hanging on large channels

robertroeser:
When streaming data from a channel RSocket would hang. Replaced the processors in the RSocketClient and RSocketServer with ones that can emit when the onNext method is called to prevent a live lock. Basically a Unicast processor with the operator fusion code removed, and  a thread-safe queue.  Also fixes a bug when an channel receiving an empty publisher wouldn't complete.

### DIFF
--- a/rsocket-test/src/main/java/io/rsocket/test/BaseClientServerTest.java
+++ b/rsocket-test/src/main/java/io/rsocket/test/BaseClientServerTest.java
@@ -112,7 +112,7 @@ public abstract class BaseClientServerTest<T extends ClientSetupRule<?, ?>> {
     assertEquals(100, outputCount);
   }
 
-  @Test//(timeout = 20000)
+  @Test(timeout = 20000)
   public void testRequestResponse10_000() {
     long outputCount =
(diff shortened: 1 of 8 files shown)


rsocket-test/src/main/java/io/rsocket/test/BaseClientServerTest.java
```diff
@@ -54,7 +57,7 @@ public void testPushMetadata10() {
     assertEquals(0, outputCount);
   }
 
-  @Test(timeout = 10000)
+  @Test//(timeout = 10000)
```
yschimke: Can we put some timeout in?

yschimke:
Does this resolve https://github.com/rsocket/rsocket/issues/241

robertroeser:
@yschimke yes it fixes that

rsocket-test/src/main/java/io/rsocket/test/BaseClientServerTest.java
robertroeser: Yeah I'll add the timeouts back

PR #442, opened 2017-11-04
Server Connection Closing on Exception

robertroeser:
RSocket server was being closed when it receives an exception that isn't being handle. It should call the error consumer, and continue processing requests.


yschimke:
Is there a unit test we should add for this and similar cases?

PR #450, opened 2017-12-01
Terminate subscribers before clearing map

### LINKED ISSUE #449: Bug: RSocketClient subscribers are not terminated on connection close
In current snapshot [RSocketClient.cleanup()](https://github.com/rsocket/rsocket-java/blob/ad372dad68c0cc73c42ce44597f3008146fcfb5b/rsocket-core/src/main/java/io/rsocket/RSocketClient.java#L432) is supposed to terminate its publishers and subscribers, but before it has chance to do that, containing maps are cleared. This leaves publishers and subscribers in stalled state - they never emit or get terminated.

robertroeser:
Fixes #449 by terminating subscriptions before clearing maps.


robertroeser:
@mostroverkhov pull request for #449

yschimke:
This is a tough review because of the format changes?  Could you reformat and push to master in a separate diff, then update?

PR #454, opened 2017-12-14
Add documentation to flag constants, mark STRICT as deprecated

simonbasle:
This commit improves a few pieces of javadoc, notably documenting most FLAG constants and marking STRICT as deprecated (as it has been removed from the protocol after v0.2).


yschimke:
Awesome, thanks for the documentation fixes!

PR #466, opened 2018-02-01
Update dependencies, cleanup RSocketClient/Server implementation

rdegnan:
Merge some changes from JCTools, clean up a few places where synchronization is no longer needed after switching to NonBlockingHashMapLong


rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/NettyDuplexConnection.java
```diff
@@ -64,16 +52,16 @@ public NettyDuplexConnection(NettyInbound in, NettyOutbound out, NettyContext co
 
   @Override
   public void dispose() {
-    onClose.onComplete();
+    context.dispose();
```
robertroeser: This didn't seem to work reliably. Did this get fixed?

PR #467, opened 2018-02-03
Simplify RSocketClient/Server implementation and avoid unnecessary flatMap

rdegnan:
Most of the diff here is from running googleJavaFormat -- the real changes are in RSocketClient/Server.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/RSocketClient.java
+++ b/rsocket-core/src/main/java/io/rsocket/RSocketClient.java
@@ -22,18 +22,17 @@ import io.rsocket.exceptions.Exceptions;
 import io.rsocket.internal.LimitableRequestPublisher;
 import io.rsocket.internal.UnboundedProcessor;
 import io.rsocket.util.NonBlockingHashMapLong;
-import org.reactivestreams.Publisher;
-import org.reactivestreams.Subscriber;
-import reactor.core.Disposable;
-import reactor.core.publisher.*;
-
-import javax.annotation.Nullable;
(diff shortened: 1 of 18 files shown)


rsocket-core/src/main/java/io/rsocket/RSocketServer.java
```diff
@@ -70,34 +69,19 @@
...
-                            t -> {
-                              errorConsumer.accept(t);
-                              return Mono.empty();
-                            }))
-            .doOnError(errorConsumer)
-            .then()
-            .subscribe();
+        .subscribe(null, this::handleSendProcessorError);
+
+    this.receiveDisposable = connection.receive().subscribe(this::handleFrame, errorConsumer);
```
robertroeser: This no long composes the backpressure from the client code and netty. That's why it's flap map - now it's disconnected because you have another subscription.

rsocket-core/src/main/java/io/rsocket/RSocketClient.java
```diff
@@ -88,36 +87,25 @@
...
-        .doFinally(
-            signalType -> {
-              cleanup();
-            })
-        .doOnError(errorConsumer)
-        .subscribe();
+    connection.onClose().doFinally(signalType -> cleanup()).subscribe(null, errorConsumer);
 
     connection
         .send(sendProcessor)
```
robertroeser: I wrote them this way for readability - operator fusion should be happening so the difference should be negligible.

PR #484, opened 2018-04-03
Fix issue with metadataPush not being subscribed to

### LINKED ISSUE #470: Connection Metadata push from Server to Client does not work
Actual frames are sent and received  correctly, but routed by ConnectionMultiplexer to Requester RSocket instead of Responder one. 

Tested with following [code](https://gist.github.com/mostroverkhov/8154bd41202b3850f3548569d85ed6b8)

Got an error
```
05 Mar 2018 09:18:21,063 DEBUG [reactor-tcp-nio-1] FrameLogger - receiving -> Frame => Stream ID: 0 Type: METADATA_PUSH Payload: metadata: "Metadata sent by server " 
05 Mar 2018 09:18:21,401 DEBUG [reactor-tcp-nio-2] TcpServer - [id: 0xcc7e142e, L:/127.0.0.1:7000 - R:/127.0.0.1:33218] READ COMPLETE
java.lang.IllegalStateException: Client received supported frame on stream 0: Frame => Stream ID: 0 Type: METADATA_PUSH Payload: metadata: "Metadata sent by server " 
	at io.rsocket.RSocketClient.handleStreamZero(RSocketClient.java:465)
	at io.rsocket.RSocketClient.handleIncomingFrames(RSocketClient.java:439)
```

rdegnan:
This commit fixes #470

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/RSocketClient.java
+++ b/rsocket-core/src/main/java/io/rsocket/RSocketClient.java
@@ -41,7 +41,6 @@ class RSocketClient implements RSocket {
   private final Function<Frame, ? extends Payload> frameDecoder;
   private final Consumer<Throwable> errorConsumer;
   private final StreamIdSupplier streamIdSupplier;
-  private final MonoProcessor<Void> started;
   private final NonBlockingHashMapLong<LimitableRequestPublisher> senders;
   private final NonBlockingHashMapLong<UnicastProcessor<Payload>> receivers;
   private final AtomicInteger missedAckCounter;
@@ -72,7 +71,6 @@ class RSocketClient implements RSocket {
     this.frameDecoder = frameDecoder;
     this.errorConsumer = errorConsumer;
     this.streamIdSupplier = streamIdSupplier;
-    this.started = MonoProcessor.create();
     this.senders = new NonBlockingHashMapLong<>(256);
     this.receivers = new NonBlockingHashMapLong<>(256);
     this.missedAckCounter = new AtomicInteger();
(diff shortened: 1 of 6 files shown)


rsocket-core/src/main/java/io/rsocket/internal/ClientServerInputMultiplexer.java
```diff
@@ -74,7 +74,7 @@ public ClientServerInputMultiplexer(DuplexConnection source, PluginRegistry plug
                 if (frame.getType() == FrameType.SETUP) {
                   type = Type.STREAM_ZERO;
                 } else {
-                  type = Type.CLIENT;
+                  type = isServer ? Type.CLIENT : Type.SERVER;
```
mostroverkhov: I don't think this change is sufficient - previously Client Requester also handled Keep-Alive frames, and now Connection gets closed eventually because of missing Keep-Alive acks.

rdegnan: That's a good point -- I think what really needs to be done is separating stream zero into a DuplexConnection that gets handled independently of the client and server rather than trying to mix stream zero frames in with one or the other. That's a bit more work, but I had started some changes in that direction so probably time to dust that off and solve this properly.

nebhale: I came to this same conclusion working on the `DuplexConnection` straw man today.  It’s just so different (almost completely disjoint set of frame types, can’t be fragmented, special lifecycle, etc.) that it deserves to be treated differently.

rdegnan: Indeed -- also the metadataPush method really doesn't belong in the RSocket interface, but in this stream zero handler, whatever we end up calling it.

nebhale: Yep, agreed.

PR #508, opened 2018-05-21
Ensure sendProcessor is disposed

rdegnan:


nebhale:
@rdegnan I’ve been tweaking the timeout of the 2M test, but it’s probably time to take it back to the way it was before my refactoring and `@Dispose` it.  I believe that timeout is now at 2 minutes, which, if it can’t complete successfully in that time, is probably not useful to test regularly.

PR #530, opened 2018-09-18
Support for rejecting setup

mostroverkhov:


rsocket-core/src/test/java/io/rsocket/SetupRejectionTest.java
```diff
@@ -0,0 +1,148 @@
...
+    public RejectingAcceptor(String errorMessage) {
+      this.errorMessage = errorMessage;
+    }
+
+    private final UnicastProcessor<RSocket> senderRSockets = UnicastProcessor.create();
+
+    @Override
+    public Mono<RSocket> accept(ConnectionSetupPayload setup, RSocket sendingSocket) {
+      senderRSockets.onNext(sendingSocket);
+      return Mono.error(new RuntimeException(errorMessage));
```
yschimke: Why not use RejectedSetupException?

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/TcpServerTransport.java
```diff
@@ -98,9 +98,7 @@ public static TcpServerTransport create(TcpServer server) {
             (in, out) -> {
               in.context().addHandler(new RSocketLengthCodec());
               TcpDuplexConnection connection = new TcpDuplexConnection(in, out, in.context());
-              acceptor.apply(connection).subscribe();
```
yschimke: nice! I think this removes a warning.

rsocket-core/src/test/java/io/rsocket/SetupRejectionTest.java
mostroverkhov: I thought It is not feasible to force (and expect) people use correct exception, so this test verifies any exception is translated  to RejectedSetupException

PR #531, opened 2018-09-27
draft impl for transmit processor

OlegDokuka:
This PR is aimed to refactor current inefficient frame routing.
At a current point, I provided some initial changes of internals that should improve performance and fixes a few issues in `SwitchTransform`.

For now, I added an implementation of TransimProcessor which simply transmit all signals between upstream and downstream. I'm going to send this impl to Reactor itself, so we will use it once we merge `TransmitProcessor` PR to Reactor-Core.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/RSocketClient.java
+++ b/rsocket-core/src/main/java/io/rsocket/RSocketClient.java
@@ -16,19 +16,24 @@
 
 package io.rsocket;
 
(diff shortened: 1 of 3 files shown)


yschimke:
Is there more context behind the change?

OlegDokuka:
@yschimke yeah. Fixed that. Sant this PR tonight so forgot about my manners ;)

PR #533, opened 2018-10-01
Upgrade reactor to Californium-RELEASE

### LINKED ISSUE #532: Upgrade reactor to Californium-RELEASE
Spring 5.1 and Spring Boot 2.1 uses Californium-RELEASE, which comes with breaking changes in its APIs, but also a change in the package names and artifact coordinates from Bismuth-SR11.
So currently we can't use Spring 5.1 and Spring Boot 2.1 with rsocket-java, which currently uses reactor Bismuth-SR11.

rdegnan:
This resolves #532


rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
```diff
@@ -115,7 +114,7 @@ public boolean isDisposed() {
     return delegate
         .receive()
         .map(AbstractionLeakingFrameUtils::fromAbstractionLeakingFrame)
-        .concatMap(function(this::toReassembledFrames));
+        .concatMap(t2 -> toReassembledFrames(t2.getT1(), t2.getT2()));
```
nebhale: Makes me sad, but I get it.

Issue #540, opened 2018-10-26
Detect closed server connection and automatically reconnect.


joshfix:
I have a simple setup between two services that I've built from provided examples.  I have one service that listens for incoming connections:

```
        TcpServerTransport tcp = TcpServerTransport.create(configProps.getPort());
        RSocketFactory.receive()
                .acceptor(itemSocketAcceptor)
                .transport(tcp)
                .start().log()
                .subscribe();
```

Another service attempts to make a connection:

```
        this.socket = RSocketFactory
                .connect()
                .transport(TcpClientTransport.create(<host>, <port>)
                .start()
                .block();
```

If I redeploy the receiving service, the client service never reports an error (unless I set a timeout on the requestResponse method).  I can't seem to find a way for the client side to detect that the underlying connection has closed and to automatically try to reconnect.  To boot, these are web applications in Spring Boot, so I'm never able to attempt to call block() during a web request.

Not sure if this is an actual issue or if I've simply overlooked some documentation or examples.  I'm anxious to adopt rsocket in my company, so any direction would be greatly appreciated.

yschimke:
@robertroeser does the load balanced socket do this for you even if you only had a single backend?

https://github.com/rsocket/rsocket-java/blob/1.0.x/rsocket-load-balancer/src/main/java/io/rsocket/client/LoadBalancedRSocketMono.java

robertroeser:
@joshfix @yschimke yes you can use the LoadBalancedRSocketMono for. You just have your RSocketSupplier create a new connection when it's called. If the connection breaks, it will call the supplier for a new RSocket and create the connection again.

joshfix:
Thanks for the quick responses... I'm playing with the load balancer but still can't seem to get it to work.  Here is my client code:

```
    @PostConstruct
    public void connect() {
        log.info("Attempting RSocket connection to " + configProps.getHost() + ":" + configProps.getPort());
        socket = LoadBalancedRSocketMono
                .create(Flux.just(Collections.singleton(new RSocketSupplier(() -> RSocketFactory
                        .connect()
                        .transport(TcpClientTransport.create(configProps.getHost(), configProps.getPort()))
                        .start()
                        .doOnSubscribe(s -> log.info("RSocket connection established.", s))))))
                .block();
    }
```
When I kill the server/acceptor, I see the following log messages:
```
[...]

PR #541, opened 2018-10-28
Fix current and future streams not terminating after connection dispose

### LINKED ISSUE #468: RSocketClient requests never complete if performed after Connection send error
Also, state (senders and receivers) of those new requests is never released. Likely because
 `connection.send(sendProcessor)...` stream gets terminated and `sendProcessor` itself becomes cancelled. 

Expected behavior is once `Connection.send` is terminated with error, all subsequent requests must be terminated with same error too.

Can be reproduced with [following test](https://gist.github.com/mostroverkhov/fc0e6202ea7dd201ab36cbc5600309b7)

mostroverkhov:
Fixes #468 , likely relates to #540

### DIFF
--- a/rsocket-core/src/test/java/io/rsocket/RSocketClientTerminationTest.java
+++ b/rsocket-core/src/test/java/io/rsocket/RSocketClientTerminationTest.java
@@ -1,0 +1,64 @@
+package io.rsocket;
+
+import io.rsocket.RSocketClientTest.ClientSocketRule;
+import io.rsocket.util.EmptyPayload;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.test.StepVerifier;
+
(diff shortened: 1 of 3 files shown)


rsocket-core/src/test/java/io/rsocket/RSocketClientTerminationTest.java
```diff
@@ -0,0 +1,64 @@
...
+  public void testCurrentStreamIsTerminatedOnConnectionClose() {
+    RSocketClient rSocket = rule.socket;
+
+    Mono.delay(Duration.ofSeconds(1))
+        .doOnNext(v -> rule.connection.dispose())
+        .subscribe();
+
+    StepVerifier.create(interaction.apply(rSocket))
+        .expectError(ClosedChannelException.class)
+        .verify(Duration.ofSeconds(5));
```
yschimke: Not specific to this PR: Are these tests "wall clock" slow?  Should we tag someway to run in CI, but not in normal developer builds?

mostroverkhov: Yes, this will be great improvement for local builds - will do that in follow-up. Also I was thinking about VirtualTimeScheduler, but eventually preferred approach with no magic in It as this way tests look less fragile

PR #547, opened 2018-11-04
loadbalancer reconnects when downstream connection goes away

robertroeser:
This addresses the issue https://github.com/rsocket/rsocket-java/issues/540 where load balancer does not reconnect.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,7 +65,7 @@ atlassian-ide-plugin.xml
 
 # NetBeans specific files/directories
 .nbattrs
-/bin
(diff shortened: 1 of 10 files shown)


yschimke:
This is a tough review, 2 failing tests in CI as well.  I'll let you work through those but generally give you a shipit then as I think active use of this from 1.0.x branch will really help.  I can add support in rsocket-cli and help test that way.

I wonder whether this sort of library code can eventually be owned outside of RSocket?  Is there a natural home for distributed systems algos like this?

PR #548, opened 2018-11-11
Switching FrameType flags back to bits

robertroeser:
Switching FrameType flag back to bits. Current version uses an EnumSet which requires a contains method call. This method traverses and array, which requires you do this every time you created a Frame. This one does a bitmask and equals.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/framing/FrameType.java
+++ b/rsocket-core/src/main/java/io/rsocket/framing/FrameType.java
@@ -17,8 +17,8 @@
 package io.rsocket.framing;
 
 import io.rsocket.Frame;
+
 import java.util.Arrays;
-import java.util.EnumSet;
 
 /**
  * Types of {@link Frame} that can be sent.
@@ -27,12 +27,12 @@ import java.util.EnumSet;
  *     Types</a>
  */
 public enum FrameType {
-
+  
   /** Reserved. */
   RESERVED(0x00),
-
+  
   // CONNECTION
-
+  
   /**
    * Sent by client to initiate protocol processing.
    *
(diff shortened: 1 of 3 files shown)


rsocket-core/src/main/java/io/rsocket/framing/FrameType.java
```diff
@@ -189,53 +185,51 @@
...
+  
   private final int encodedType;
-
-  private final EnumSet<Flags> flags;
-
+  private final int flags;
+  
   FrameType(int encodedType) {
-    this(encodedType, EnumSet.noneOf(Flags.class));
+    this(encodedType, 0);
```
OlegDokuka: 0 -> Flags.NONE or Flags.EMPTY

rsocket-core/src/main/java/io/rsocket/framing/FrameType.java
```diff
@@ -278,47 +276,41 @@ public boolean canHaveMetadata() {
...
-    CAN_HAVE_METADATA,
-
-    HAS_INITIAL_REQUEST_N,
-
-    IS_FRAGMENTABLE,
-
-    IS_REQUEST_TYPE;
+  
+  private static class Flags {
+    private static final int CAN_HAVE_DATA         = 0b10000;
```
OlegDokuka: private static final int EMPTY                   = 0b00000;

rsocket-core/src/main/java/io/rsocket/framing/FrameType.java
OlegDokuka:
```suggestion
    this(encodedType, Flags.EMPTY);
```

rsocket-core/src/main/java/io/rsocket/framing/FrameType.java
OlegDokuka:
```suggestion
    private static final int EMPTY                 = 0b00000;
    private static final int CAN_HAVE_DATA         = 0b10000;
```

PR #549, opened 2018-11-14
updates to netty's duplex connection

robertroeser:
Some performance changes to netty's duplex connection

### DIFF
--- a/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/TcpDuplexConnection.java
+++ b/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/TcpDuplexConnection.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *       http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


smaldini:
Should continue with specific flatMapEmpty operator and/or fusion support

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/TcpDuplexConnection.java
```diff
@@ -60,11 +67,19 @@ public boolean isDisposed() {
...
   }
 
   @Override
   public Mono<Void> sendOne(Frame frame) {
-    return connection.outbound().sendObject(frame.content()).then();
+    return outbound.sendObject(frame.content()).then();
+  }
+
+  @Override
+  public Publisher<Void> apply(ByteBuf frame) {
```
yschimke: What is the utility of this? Also noticeable that because of the internal implementation, this one is

apply(ByteBuf), sendOne(Frame)

While WS is

apply(Frame), sendOne(Frame)

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/TcpDuplexConnection.java
```diff
@@ -1,11 +1,11 @@
 /*
- * Copyright 2015-2018 the original author or authors.
+ * Copyright (c) 2011-2018 Pivotal Software Inc, All Rights Reserved.
```
yschimke: Doesn't match the contributing or license recommendations.  What is the backstory behind this change?

robertroeser: copy and paste from another branch - should be the original author or authors.

smaldini: my bad sorry i'll update my intelliJ template for the next time

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/TcpDuplexConnection.java
smaldini: i just avoid an extra function allocation (inlining the flatMap function) but we could actually inline map as well so its `apply(Frame)`

smaldini: Also we can just not inline anything the function cost is low enough if send is just called once every request.

PR #550, opened 2018-11-14
Client server input multiplexer tuning

robertroeser:
Removes the groupBy operation in the ClientServerInputMultiplexer class.

Also switches the StreamIdSupplier to use CAS operations instead of synchronized


robertroeser:
Can't seem to reproduce the failing test locally - maybe its a race condition from a slower machine?

rsocket-core/src/main/java/io/rsocket/StreamIdSupplier.java
```diff
@@ -16,28 +16,37 @@
...
     return new StreamIdSupplier(0);
   }
+
+  int nextStreamId() {
+    int n;
+    int o;
+    do {
+      o = streamId;
+      n = o + 2;
+    } while (!STREAM_ID.compareAndSet(this, o, n));
```
yschimke: Can you replace this with `STREAM_ID.addAndGet(this, 2)`

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/TcpDuplexConnection.java
```diff
@@ -60,11 +67,19 @@ public boolean isDisposed() {
...
   }
 
   @Override
   public Mono<Void> sendOne(Frame frame) {
-    return connection.outbound().sendObject(frame.content()).then();
+    return outbound.sendObject(frame.content()).then();
+  }
+
+  @Override
+  public Publisher<Void> apply(ByteBuf frame) {
```
yschimke: What is the utility of this?  Also noticeable that because of the internal implementation, this one is

apply(ByteBuf), sendOne(Frame)

While WS is

apply(Frame), sendOne(Frame)

yschimke: Think this comes from https://github.com/rsocket/rsocket-java/pull/549/files

PR #552, opened 2018-11-19
perf changes

robertroeser:
Changes the RScocket client and server to use the a synchronized version of the Agrona Int2ObjectHashMap. The NonBlockingHashMapLong was leading to GC pauses.

Uses SendPublisher that writes to the Netty channel directly - on my laptop the performance increase is approx 200% throughput improvement, with better tail latency on the Ping/Pong test as well our internal code that relies on RSocket.


yschimke:
LGTM

I keep expecting the use of libraries like reactor to mean that the engine of RSocket is this shiny simple "only one way to logically write this" chrome engine.  I'm still getting surprised that we end up writing so much stream infra code.  In this case, things like SendPublisher.

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
```diff
@@ -42,8 +49,8 @@
   private final Function<Frame, ? extends Payload> frameDecoder;
   private final Consumer<Throwable> errorConsumer;
 
-  private final NonBlockingHashMapLong<Subscription> sendingSubscriptions;
```
rdegnan: I think handleSendProcessorError and handleSendProcessorCancel should now be synchronized as they iterate over sendingSubscriptions/channelProcessors (previously this wasn't necessary because of Cliff Click magic).

rsocket-core/src/main/java/io/rsocket/RSocketClient.java
```diff
@@ -40,8 +46,8 @@
   private final Function<Frame, ? extends Payload> frameDecoder;
   private final Consumer<Throwable> errorConsumer;
   private final StreamIdSupplier streamIdSupplier;
-  private final NonBlockingHashMapLong<LimitableRequestPublisher> senders;
```
rdegnan: I think handleSendProcessorError and handleSendProcessorCancel and terminate should now be synchronized as they iterate over sendingSubscriptions/channelProcessors (previously this wasn't necessary because of Cliff Click magic).

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
robertroeser: I wrapped the map with Collections.synchronizedMap so it should effectively be doing that.

robertroeser:
@yschimke I'm hoping these changes make it back into reactor-*

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
```diff
@@ -16,23 +16,30 @@
...
 package io.rsocket;
 
-import static io.rsocket.Frame.Request.initialRequestN;
-import static io.rsocket.frame.FrameHeaderFlyweight.FLAGS_C;
-import static io.rsocket.frame.FrameHeaderFlyweight.FLAGS_M;
-
 import io.rsocket.exceptions.ApplicationErrorException;
 import io.rsocket.exceptions.ConnectionErrorException;
 import io.rsocket.framing.FrameType;
+import io.rsocket.internal.Int2ObjectHashMap;
```
hepin1989: Netty already has one.

robertroeser: Switched to it

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
robertroeser: I looked at the code again. I think we are fine - the problem would be if something adds something to map while we're iterating over it. Looking at the non blocking map the same thing would happen. We should probably add something to the request* methods so that you can't add items to the map when handling a cancel or an error, and then this isn't a problem.

PR #558, opened 2019-01-02
Resume support

mostroverkhov:
Work in progress. Changes are somewhat bigger than expected due to related keep-alive update (new impl is for resumption, for now It is part of this PR ) and formatter work on sources

implement stream resume

use keep-alive frames last-received position for resume cache truncation

in order to implement above, remove keep-alive support from `RSocketClient/Server`, and reimplement It as `KeepAliveDuplexConnection`


yschimke:
Can we reformat master and split that out from this PR, from the first few files it's tough to review with it mixed.

yschimke:
Do you plan to test against rsocket-js? n.b. I think rsocket-cpp are still not in sync.

https://github.com/rsocket/rsocket-cpp/blob/629d3d47753f99b2036f2c3cd191188bb9471394/rsocket/internal/WarmResumeManager.cpp#L60

https://github.com/rsocket/rsocket-js/blob/master/packages/rsocket-core/src/RSocketResumableTransport.js#L415

I suggest matching the rsocket-js approach as IIRC it's where the spec ended up.

mostroverkhov:
@yschimke good call - I will cleanup this pr from formatter changes. Frames count is used for position tracking, so it is on par with js implementation - going to check how those work together

yschimke:
@mostroverkhov A reminder there is a demo project of rsocket-js -> rsocket-java here

https://github.com/yschimke/rsocket-demo

Might be a useful test base to start with.

yschimke:
cc @lexs who knows the rsocket-js side, and to be honest the resumption code as well, much better than I do.

mostroverkhov:
@yschimke I am looking forward this too :).  If not this weekend then next one - quite busy weeks atm. Other than splitting things into separate commits and testing against other jvm implementations, It would be good to have support for out-of-process storage of  messages to support server-to-server resumption use case.

yschimke:
If it isn't a useful part of your testing, I can try adding support for the js demo and cli.  More active test cases should help.

mostroverkhov:
this will be superseded by something on top of https://github.com/rsocket/rsocket-java/tree/feature/resumeOnTopOfFragmentation - basically above changes on top of bytebuf frames, but w/o randomly changed unrelated sources & split into several commits

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumeStore.java
```diff
@@ -0,0 +1,113 @@
...
+    this.tag = tag;
+    this.cacheLimit = cacheLimit;
+    this.cachedFrames = cachedFramesQueue(cacheLimit);
+  }
+
+  @Override
+  public void saveFrame(ByteBuf frame) {
+    cachedFramesSize.incrementAndGet();
+    cachedFrames.offer(frame.retain());
+    if (cachedFramesSize.get() == cacheLimit) {
```
yschimke: incrementAndGet can be done here in a single op.

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumeStore.java
```diff
@@ -0,0 +1,113 @@
...
+  public InMemoryResumeStore(String tag, int cacheLimit) {
+    this.tag = tag;
+    this.cacheLimit = cacheLimit;
+    this.cachedFrames = cachedFramesQueue(cacheLimit);
+  }
+
+  @Override
+  public void saveFrame(ByteBuf frame) {
+    cachedFramesSize.incrementAndGet();
+    cachedFrames.offer(frame.retain());
```
yschimke: add seems better than offer considering you don't check return value.

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumeStore.java
```diff
@@ -0,0 +1,113 @@
...
+    long processedCount = 0;
+    while (processedCount < removeCount) {
+      releaseFrame();
+      processedCount++;
+    }
+    logger.info("{} Removed frames. Current size: {}", tag, cachedFramesSize.get());
+  }
+
+  private void releaseFrame() {
+    ByteBuf content = cachedFrames.poll();
```
yschimke: remove instead of poll?

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumeStore.java
```diff
@@ -0,0 +1,113 @@
...
+
+  /*must be subscribed at most once on resumeStart -> resumeFrames cycle*/
+  @Override
+  public Flux<ByteBuf> resumeStream() {
+    return Flux.create(s -> {
+      int size = cachedFramesSize.get();
+      logger.info("{} Resuming stream size: {}",tag, size);
+      /*spsc queue has no iterator - iterating by consuming*/
+      for (int i = 0; i < size; i++) {
+        ByteBuf frame = cachedFrames.poll();
```
yschimke: Should this be destructive in case this resume get disconnected?

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
```diff
@@ -104,11 +105,14 @@ public static ServerRSocketFactory receive() {
 
     private boolean resumeEnabled;
     private Supplier<ResumeToken> resumeTokenSupplier = ResumeToken::generate;
-    private int resumeCacheSize = 32768;
+    private Function<? super ResumeToken, ? extends ResumeStore> resumeStoreFactory =
+        token -> new InMemoryResumeStore("client", 32768);
```
yschimke: This seems like a really generous default, given each message could be quite large.  at 1k, this is 32mb per connection.

yschimke: At some point the pain of resumption may be worse than establishing a new connection for clients, e.g. if too much history is retained.

rsocket-core/src/main/java/io/rsocket/frame/ResumeFrameFlyweight.java
```diff
@@ -0,0 +1,85 @@
...
+  public static long firstAvailableClientPos(ByteBuf byteBuf) {
+    FrameHeaderFlyweight.ensureFrameType(FrameType.RESUME, byteBuf);
+
+    byteBuf.markReaderIndex();
+    // header + version
+    int tokenPos = FrameHeaderFlyweight.size() + Integer.BYTES;
+    byteBuf.skipBytes(tokenPos);
+    //token
+    int tokenLength = byteBuf.readShort() & 0xFFFF;
+    byteBuf.skipBytes(tokenLength);
```
yschimke: assume WIP?

rsocket-core/src/main/java/io/rsocket/plugins/DuplexConnectionInterceptor.java
```diff
@@ -23,7 +23,7 @@
 public @FunctionalInterface interface DuplexConnectionInterceptor
     extends BiFunction<DuplexConnectionInterceptor.Type, DuplexConnection, DuplexConnection> {
   enum Type {
-    STREAM_ZERO,
+    SETUP,
```
yschimke: Why change this from STREAM_ZERO which is anything related to the connection, to setup?

yschimke:
Skim review - tough to review without basically running this in place.  But looking really good.  Proof will be in self resumption and real world cases, particularly against the js Impl.

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumeStore.java
mostroverkhov: Queue implementation (`SpscArrayQueue`) does not support that method. I can check return value and throw if queue is full.  Also, because `saveFrame` wont be called concurrently (access is serialized by calling code), current logic looks sufficient . Another thing - looks like atomic is not needed here as `cachedFramesSize` (and other fields except `impliedPosition`) due to above - just should be safe to access from different threads

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumeStore.java
mostroverkhov: Queue implementation (SpscArrayQueue) does not support that method

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumeStore.java
mostroverkhov: Most apps dealing with `reactor` and refcounted objects have `Refcounted.release()` in `Hooks.onNextDropped` - otherwise cancelled streams would lead to memory leaks. The idea here is to retain frames for future resumes, and assume calling code will release accordingly (either transport or `Hooks.onNextDropped`)

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
mostroverkhov: I can change It to 1024?

rsocket-core/src/main/java/io/rsocket/plugins/DuplexConnectionInterceptor.java
mostroverkhov: Because interceptors are applied in `ClientServerInputMultiplexer`, and It considers SETUP frame only as part of STREAM_ZERO connection - https://github.com/rsocket/rsocket-java/blob/e48658e28676b6dc6dcef20b2fb70b38f220ad8d/rsocket-core/src/main/java/io/rsocket/internal/ClientServerInputMultiplexer.java#L77.

rsocket-core/src/main/java/io/rsocket/resume/ResumedFramesCalculator.java
```diff
@@ -0,0 +1,52 @@
...
+import reactor.core.publisher.Mono;
+
+interface ResumedFramesCalculator {
+
+  ResumedFramesCalculator ofClient = new ClientResumedFramesCalculator();
+  ResumedFramesCalculator ofServer = new ServerResumedFramesCalculator();
+
+  Mono<Long> calculate(ResumptionState local, ResumptionState remote);
+
+  class ClientResumedFramesCalculator implements ResumedFramesCalculator {
```
robertroeser: Is there are reason this is async? It doesn't look like its doing anything sync

mostroverkhov: It captures error if resumption is not possible with states provided, which then is used in RSocket sessions like https://github.com/mostroverkhov/rsocket-java/blob/42314ce746aab0c7908a677f72fabbe8c872b54d/rsocket-core/src/main/java/io/rsocket/resume/ServerRSocketSession.java#L105

robertroeser: Sorry - this comment was supposed to be on `   Mono<Long> calculate(ResumptionState local, ResumptionState remote); ` 

I still don't still don't see why this async?

PR #566, opened 2019-01-07
add parameters to make weighted socket retry configurable

robertroeser:


robertroeser:
@joshfix added parameter to the load balancer to let you control how often a weighted socket retries.

Issue #569, opened 2019-01-24
Add specific requestChannel(Payload, Publisher<Payload>) support for RSocket Server


smaldini:
An `RSocketServer` extending `RSocket` could be a runtime detected feature that allows invoking a 2-arguments Payload/Publisher<Payload> RSocket handler method.

The reason for a 2-arguments is to explicitly separate the first Payload possibly carrying routing informations  (https://github.com/rsocket/rsocket/pull/254) or any other such as tracing, composite etc. All others interactions do not have this issues having a single Payload argument readily available for introspection where requestChannel requires the user to peek the first emitted Payload to achieve the same effect.

yschimke:
I like this option because the wrinkles with this interaction are

- needing first message to route
- supporting a channel that starts with messages from the server to the client before first message
- different types specifically for the first message

Should this interaction be flagged in the channel metadata when possible?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

robertroeser:
I would prefer we don't characterize this as something for metadata.  I think if we're doing this it's to optimize requestChannel so you don't need to call the SwitchTransform to peek at the first payload. It is equally likely that an end user may want to look at their first payload and see where to send the request.

> The reason for a 2-arguments is to explicitly separate the first Payload possibly carrying routing informations (rsocket/rsocket#254) ...

To prevent confusion we aren't separating the first payload. Like you said - we are letting you peak at the first item emitted from a stream without using SwitchTranform.  The first emitted payload is the same payload as the Payload argument passed in requestChannel(Payload payload, Publishers<Payload> publishers)

Regarding metadata sent once - that is possible now. Its up to the Payload creator if they want to send the metadata each time.

smaldini:
fixed by the PR above

PR #572, opened 2019-01-28
Adds specific requestChannel(Payload, Publisher<Payload>) 

robertroeser:
Optimize request channel to let you peak at the first item in the stream #569


rsocket-core/src/main/java/io/rsocket/RequestHandler.java
```diff
@@ -0,0 +1,19 @@
+package io.rsocket;
+
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Flux;
+
+/**
+ * Extends the {@link RSocket} that allows an implementor to peek at the first request payload of a channel.
```
yschimke:
```suggestion
 * Extends the {@link RSocket} that allows an implementer to peek at the first request payload of a channel.
```

rsocket-core/src/main/java/io/rsocket/RequestHandler.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public interface RequestHandler extends RSocket {
+  /**
+   * Implement this method to peak at the first payload of the incoming request stream
```
yschimke: Maybe clarify that peak means that payloads still includes the peaked payload.

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
```diff
@@ -39,15 +42,17 @@
 import static io.rsocket.frame.FrameHeaderFlyweight.FLAGS_M;
 
 /** Server side RSocket. Receives {@link Frame}s from a {@link RSocketClient} */
-class RSocketServer implements RSocket {
+class RSocketServer implements RequestHandler {
 
   private final DuplexConnection connection;
   private final RSocket requestHandler;
+  private final RequestHandler optimizedRequestHandler;
+  private final boolean hasOptimizedRequestHandler;
```
smaldini: No need for an extra boolean property, just checking optimizedRequestHandler != null should be fine

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
```diff
@@ -69,12 +74,22 @@
       Consumer<Throwable> errorConsumer,
       long tickPeriod,
       long ackTimeout) {
+    
+    if (requestHandler instanceof RequestHandler) {
+      this.optimizedRequestHandler = (RequestHandler) requestHandler;
+      this.hasOptimizedRequestHandler = true;
+      this.requestHandler = null;
```
smaldini: I would keep the requestHandler assigned anyway for consistency.

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
```diff
@@ -191,6 +211,15 @@ private void handleSendProcessorCancel(SignalType t) {
     }
   }
 
+  @Override
+  public Flux<Payload> requestChannel(Payload payload, Publisher<Payload> payloads) {
+    try {
+      return responderRSocket.requestChannel(payloads);
```
yschimke: Shouldn't this call the other form with payload as the first argument?

rstoyanchev:
Not sure if this is expected or not but I tried using the `ResponderRSocket` interface and ran into `UnsupportedOperationException` because of an `RSocketProxy` interceptor that was only aware of the `RSocket` interface. In my case it was an easy fix but I can imagine a case where the interceptor is not mine to change. 

Perhaps there needs to be an `ResponderRSocketProxy` and that should be expected for a receiving side interceptor?

PR #575, opened 2019-02-08
Feature/framing

robertroeser:
Originally RSocket didn't use ByteBuf, and the existing frame classes are a legacy of that.  This change finishes the framing work @nebhale  started, and turns all the Frames into flyweights. It removes the Frame class completely. The frames originally required offsets and this has been replaced using ByteBuf's reader and writer indexes. 

Frame length is not long part of the header flyweight. It is a separate flyweight that can be used by transports that require a length. It only needs to be applied to in the duplex connection before sending the frame.

The Payload classes remained unchanged.

Only two things are now recycled - the Payloads, and ByteBufs.

There is an API change in the RSocketFactory. The default remains unchanged but instead of taking a function that turns a ByteBuf into a Payload it now takes a specific type called a PayloadDecoder. There are constants in the interface to a zero copy implementation and a default copying implementation. There is also a method added to let the call specify a ByteBufAllocator if they want to change from ByteBufAllocator.DEFAULT

These changes - aside from RSocketFactory  - are transparent.


rsocket-core/src/main/java/io/rsocket/ConnectionSetupPayload.java
```diff
@@ -16,117 +16,115 @@
...
 public abstract class ConnectionSetupPayload extends AbstractReferenceCounted implements Payload {
-
-  public static ConnectionSetupPayload create(final Frame setupFrame) {
-    Frame.ensureFrameType(FrameType.SETUP, setupFrame);
+  
+  public static ConnectionSetupPayload create(final ByteBuf setupFrame) {
     return new DefaultConnectionSetupPayload(setupFrame);
   }
-
+
```
yschimke: Are we having some format war? Every PR recently seems to reformat more lines than it changes...

robertroeser: I didn't run a formatter on purpose I was going to wait till after this pull request and then make the formatter required for a build to pass.

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
```diff
@@ -72,7 +84,7 @@ public FragmentationDuplexConnection(DuplexConnection delegate, int maxFragmentS
    *     be fragmented.
    * @throws NullPointerException if {@code byteBufAllocator} or {@code delegate} are {@code null}
    * @throws IllegalArgumentException if {@code maxFragmentSize} is not {@code positive}
-   */
+   *//*
```
yschimke: Do you want to check in this commented code?

robertroeser: yeah - should have added to my pull request that its not quite done yet, and there's a couple things left. I wanted people to start looking at it.

rsocket-transport-local/src/test/java/io/rsocket/transport/local/LocalTransportTest.java
```diff
@@ -17,19 +17,29 @@
...
 
 import io.rsocket.test.TransportTest;
+import org.junit.jupiter.api.Test;
+
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicInteger;
 
-final class LocalTransportTest implements TransportTest {
-
+final class LocalTransportTest {// implements TransportTest {
```
yschimke: Is this WIP or you want to land, if the latter can you link to a github issue?

yschimke:
Overall the change looks sensible, but I'm probably not the best to give the final shipit.

Awkward with review with the large amount of reformatted and also large commented out blocks.

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
mostroverkhov: @yschimke just removed one commented block (not this one though), so those will be cleaned out also - one change at a time.

rsocket-core/src/main/java/io/rsocket/frame/LeaseFlyweight.java
```diff
@@ -0,0 +1,48 @@
...
+    byteBuf.resetReaderIndex();
+    return numRequests;
+  }
+
+  public static ByteBuf metadata(final ByteBuf byteBuf) {
+    boolean hasMetadata = FrameHeaderFlyweight.hasMetadata(byteBuf);
+    FrameHeaderFlyweight.ensureFrameType(FrameType.LEASE, byteBuf);
+    byteBuf.markReaderIndex();
+    byteBuf.skipBytes(FrameHeaderFlyweight.size() + Integer.BYTES * 2);
+    ByteBuf metadata = DataAndMetadataFlyweight.metadataWithoutMarking(byteBuf, hasMetadata);
```
mostroverkhov: `metadataWithoutMarking` assumes there is also `metadata` length encoded - but according to spec, `Lease` frame should not encode `metadata` length

rsocket-core/src/main/java/io/rsocket/frame/ExtensionFrameFlyweight.java
```diff
@@ -0,0 +1,34 @@
...
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
+
+public class ExtensionFrameFlyweight {
+  private ExtensionFrameFlyweight() {}
+
+  public static ByteBuf encode(
+      ByteBufAllocator allocator, int streamId, int extendedType, ByteBuf data) {
+    int flags = FrameHeaderFlyweight.FLAGS_I | FrameHeaderFlyweight.FLAGS_M;
```
mostroverkhov: Metadata is not written - flag should not be set (or support writing both data and metadata - and update logic accordingly)

rsocket-core/src/main/java/io/rsocket/frame/FrameHeaderFlyweight.java
```diff
@@ -31,29 +12,13 @@
...
-  private FrameHeaderFlyweight() {}
-
-  public static final int FRAME_HEADER_LENGTH;
-
-  private static final int FRAME_TYPE_BITS = 6;
-  private static final int FRAME_TYPE_SHIFT = 16 - FRAME_TYPE_BITS;
-  private static final int FRAME_FLAGS_MASK = 0b0000_0011_1111_1111;
-
+public final class FrameHeaderFlyweight {
   public static final int FRAME_LENGTH_SIZE = 3;
```
mostroverkhov: nit: `FRAME_LENGTH_SIZE` and `FRAME_LENGTH_MASK` are TCP transport specific now, so maybe move them to `FrameLengthFlyweight`?

robertroeser:
@mostroverkhov - you good with the changes? Can I merge this?

mostroverkhov:
@robertroeser There is still missing support for fragmentation, but that can be done is follow-up PR

PR #585, opened 2019-02-26
add fragmentation support using new frames

robertroeser:
Signed-off-by: Robert Roeser <rroeserr@gmail.com>


rsocket-core/src/main/java/io/rsocket/RSocketServer.java
```diff
@@ -345,11 +348,28 @@ private void handleFireAndForget(int streamId, Mono<Void> result) {
...
-        .map(payload -> PayloadFrameFlyweight.encodeNextComplete(allocator, streamId, payload))
+        .map(
+            payload -> {
+              ByteBuf byteBuf = null;
+              try {
+                byteBuf = PayloadFrameFlyweight.encodeNextComplete(allocator, streamId, payload);
+              } catch (Throwable t) {
+                if (byteBuf != null) {
+                  ReferenceCountUtil.safeRelease(byteBuf);
+                  ReferenceCountUtil.safeRelease(payload);
```
yschimke: Why this release as well as line 362?

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
```diff
@@ -30,139 +39,88 @@
...
-    this.byteBufAllocator =
-        Objects.requireNonNull(byteBufAllocator, "byteBufAllocator must not be null");
-    this.delegate = Objects.requireNonNull(delegate, "delegate must not be null");
-
-    NumberUtils.requireNonNegative(maxFragmentSize, "maxFragmentSize must be positive");
+      DuplexConnection delegate, ByteBufAllocator allocator, int mtu, boolean encodeLength) {
+    Objects.requireNonNull(delegate, "delegate must not be null");
+    Objects.requireNonNull(allocator, "byteBufAllocator must not be null");
+    if (mtu < MIN_MTU_SIZE) {
+      throw new IllegalArgumentException("smallest allowed mtu size is 64 bytes");
```
yschimke: for a non hot path exception, worth referring to MIN_MTU_SIZE in the message instead of 64

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
```diff
@@ -30,139 +39,88 @@
...
-            signalType -> {
-              Collection<FrameReassembler> values;
-              synchronized (FragmentationDuplexConnection.this) {
-                values = frameReassemblers.values();
-              }
-              values.forEach(FrameReassembler::dispose);
-            })
-        .subscribe();
+  private boolean shouldFragment(FrameType frameType, int readableBytes) {
+    if (frameType.isFragmentable()) {
```
yschimke: tioli: more readable as a one liner

return frameType.isFragmentable() && readableBytes > mtu

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameFragmenter.java
```diff
@@ -24,168 +34,212 @@
...
+        return PayloadFrameFlyweight.encode(
+            allocator, streamId, true, false, true, metadataFragment, dataFragment);
+      case NEXT_COMPLETE:
+        return PayloadFrameFlyweight.encode(
+            allocator, streamId, true, true, true, metadataFragment, dataFragment);
+      case COMPLETE:
+        return PayloadFrameFlyweight.encode(
+            allocator, streamId, true, true, false, metadataFragment, dataFragment);
+      default:
+        throw new IllegalStateException("unsupported fragment type");
```
yschimke: nit: useful to include frameType in the message

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameFragmenter.java
```diff
@@ -24,168 +34,212 @@
...
-
-      if (initialFragmentCreated) {
-        return createPayloadFrame(byteBufAllocator, !complete, data == null, metadata, data);
-      } else {
-        initialFragmentCreated = true;
-        return frame.createFragment(byteBufAllocator, metadata, data);
-      }
-    }
+    boolean follows =
+        metadata == Unpooled.EMPTY_BUFFER
```
yschimke: why check this? doesn't metadata.isReadable() do this?

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameFragmenter.java
```diff
@@ -24,168 +34,212 @@
...
+          break;
+          // Payload and synthetic types
+        case PAYLOAD:
+        case NEXT:
+        case NEXT_COMPLETE:
+        case COMPLETE:
+          metadata = PayloadFrameFlyweight.metadata(frame);
+          break;
+        default:
+          throw new IllegalStateException("unsupported fragment type");
```
yschimke: as above include frameType

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameReassembler.java
```diff
@@ -25,143 +37,252 @@
  *     href="https://github.com/rsocket/rsocket/blob/master/Protocol.md#fragmentation-and-reassembly">Fragmentation
  *     and Reassembly</a>
  */
-final class FrameReassembler implements Disposable {
+final class FrameReassembler extends AtomicBoolean implements Disposable {
```
yschimke: why extend AtomicBoolean instead of a field?  memory?

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameReassembler.java
```diff
@@ -25,143 +37,252 @@
...
+        handleNoFollowsFlag(frame, sink, streamId);
+      } else {
+        handleFollowsFlag(frame, streamId, frameType);
       }
 
-      return this.metadata == null
-          ? metadata.retain()
-          : Unpooled.wrappedBuffer(this.metadata, metadata.retain());
+    } catch (Throwable t) {
+      logger.error("error reassemble frame", t);
```
yschimke:
```suggestion
      logger.error("error reassembling frame", t);
```

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameReassembler.java
```diff
@@ -25,143 +37,252 @@
...
+      if (!hasFollows) {
+        handleNoFollowsFlag(frame, sink, streamId);
+      } else {
+        handleFollowsFlag(frame, streamId, frameType);
       }
 
-      return this.metadata == null
-          ? metadata.retain()
-          : Unpooled.wrappedBuffer(this.metadata, metadata.retain());
+    } catch (Throwable t) {
```
yschimke: This is a very generic error to handle here when there doesn't seem to be a handoff to client code, i.e. it's either an encoding problem on the wire or a logic bug in this code.

rsocket-core/src/main/java/io/rsocket/frame/FrameUtil.java
```diff
@@ -0,0 +1,108 @@
...
+import io.netty.buffer.ByteBufUtil;
+import io.netty.buffer.Unpooled;
+
+public class FrameUtil {
+  private FrameUtil() {}
+
+  public static String toString(ByteBuf frame) {
+    FrameType frameType = FrameHeaderFlyweight.frameType(frame);
+    int streamId = FrameHeaderFlyweight.streamId(frame);
+    StringBuilder payload = new StringBuilder();
```
yschimke: worth presizing, 16 default will require a realloc.

rsocket-core/src/main/java/io/rsocket/frame/RequestChannelFrameFlyweight.java
```diff
@@ -2,13 +2,35 @@
...
 
+  public static ByteBuf encode(
+      ByteBufAllocator allocator,
+      int streamId,
+      boolean fragmentFollows,
+      boolean complete,
+      long requestN,
+      Payload payload) {
+
+    int reqN = requestN > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) requestN;
```
yschimke: suggest asserting on negative values as a public util method

rsocket-core/src/main/java/io/rsocket/frame/RequestFlyweight.java
```diff
@@ -31,7 +32,7 @@ ByteBuf encode(
       ByteBuf data) {
     int flags = 0;
 
-    if (metadata != null) {
+    if (metadata != null && Unpooled.EMPTY_BUFFER != metadata) {
```
yschimke: Isn't EMPTY_BUFFER the "" case? and null is the case where flag should be unset?

rsocket-core/src/main/java/io/rsocket/internal/ClientServerInputMultiplexer.java
```diff
@@ -148,7 +149,7 @@ public InternalDuplexConnection(
     @Override
     public Mono<Void> send(Publisher<ByteBuf> frame) {
       if (debugEnabled) {
```
yschimke: being pedantic, shouldn't debug levels be able to change in a running app?

rsocket-core/src/main/java/io/rsocket/util/ByteBufPayload.java
```diff
@@ -179,4 +123,70 @@ public static Payload create(Payload payload) {
...
   }
+
+  @Override
+  public boolean hasMetadata() {
+    return metadata != null;
+  }
+
+  @Override
+  public ByteBuf sliceMetadata() {
+    return metadata == null ? Unpooled.EMPTY_BUFFER : metadata.slice();
```
yschimke: I wonder whether this implicit conversion from null to EMPTY_BUFFER could be confusing for the null/"" edge case

rsocket-core/src/test/java/io/rsocket/fragmentation/FragmentationIntegrationTest.java
```diff
@@ -0,0 +1,56 @@
...
+  private static byte[] metadata = new byte[128];
+
+  static {
+    ThreadLocalRandom.current().nextBytes(data);
+    ThreadLocalRandom.current().nextBytes(metadata);
+  }
+
+  private ByteBufAllocator allocator = ByteBufAllocator.DEFAULT;
+
+  @DisplayName("fragments and reassmbles data")
```
yschimke:
```suggestion
  @DisplayName("fragments and reassembles data")
```

rsocket-core/src/test/java/io/rsocket/fragmentation/FragmentationIntegrationTest.java
```diff
@@ -0,0 +1,56 @@
...
+  @Test
+  void fragmentAndReassembleData() {
+    System.out.println("original");
```
yschimke: delete?

rsocket-core/src/test/java/io/rsocket/frame/RequestFlyweightTest.java
```diff
@@ -40,7 +40,7 @@ void testEncodingWithEmptyMetadata() {
 
     frame = FrameLengthFlyweight.encode(ByteBufAllocator.DEFAULT, frame.readableBytes(), frame);
 
-    assertEquals("00000e0000000119000000000100000064", ByteBufUtil.hexDump(frame));
+    assertEquals("00000b0000000118000000000164", ByteBufUtil.hexDump(frame));
```
yschimke: Was this a bug?

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/TcpDuplexConnection.java
```diff
@@ -35,12 +35,25 @@
...
    */
   public TcpDuplexConnection(Connection connection) {
+    this(connection, true);
+  }
+
+  /**
+   * Creates a new instance
+   *
+   * @param encodeLength indicates if this connection should encode the length or not.
+   * @param connection the {@link Connection} to for managing the server
```
yschimke: "to for"?

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameReassembler.java
```diff
@@ -25,143 +37,252 @@
...
+      }
 
-    private @Nullable ByteBuf accumulateMetadata(FragmentableFrame fragment) {
-      ByteBuf metadata = fragment.getUnsafeMetadata();
+      if (!frameType.isFragmentable()) {
+        sink.next(frame);
+        return;
+      }
+
+      boolean hasFollows = FrameHeaderFlyweight.hasFollows(frame);
```
mostroverkhov: nit: there is a mention in a spec that `follows` flag should be ignored if `complete` flag is set

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameReassembler.java
```diff
@@ -25,143 +37,252 @@
...
+
+    return FragmentationFlyweight.encode(allocator, header, metadata, data);
+  }
+
+  private ByteBuf assembleData(ByteBuf frame, int streamId) {
+    ByteBuf data;
+    CompositeByteBuf cd = removeData(streamId);
+    if (cd != null) {
+      ByteBuf d = PayloadFrameFlyweight.data(frame);
+      if (d != null) {
```
mostroverkhov: nit: data is not nullable

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameFragmenter.java
```diff
@@ -24,168 +34,212 @@
...
+      case PAYLOAD:
+        return PayloadFrameFlyweight.encode(
+            allocator, streamId, true, false, false, metadataFragment, dataFragment);
+      case NEXT:
+        return PayloadFrameFlyweight.encode(
+            allocator, streamId, true, false, true, metadataFragment, dataFragment);
+      case NEXT_COMPLETE:
+        return PayloadFrameFlyweight.encode(
+            allocator, streamId, true, true, true, metadataFragment, dataFragment);
+      case COMPLETE:
```
mostroverkhov: Is this possible for first fragment? `COMPLETE` is synthetic type for frame not having `Payload` (data/metadata) - unlikely to be fragmented

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameFragmenter.java
```diff
@@ -24,168 +34,212 @@
...
+        // Payload and synthetic types
+      case PAYLOAD:
+        return PayloadFrameFlyweight.encode(
+            allocator, streamId, true, false, false, metadataFragment, dataFragment);
+      case NEXT:
+        return PayloadFrameFlyweight.encode(
+            allocator, streamId, true, false, true, metadataFragment, dataFragment);
+      case NEXT_COMPLETE:
+        return PayloadFrameFlyweight.encode(
+            allocator, streamId, true, true, true, metadataFragment, dataFragment);
```
mostroverkhov: nit: spec mentions `follows` flag should be ignored if `complete` flag is set - wondering if `complete` flag should be deferred to last fragment?

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameReassembler.java
robertroeser: something I picked up from rxjava

rsocket-core/src/main/java/io/rsocket/frame/FrameUtil.java
robertroeser: maybe? this was kind of for debugging purposes, but I can add a size

rsocket-core/src/main/java/io/rsocket/frame/RequestFlyweight.java
robertroeser: yeah I  that is correct - that would explain the other thing I changed in the unit - it was kind of early - I will go through and check the flyweights.

rsocket-core/src/main/java/io/rsocket/internal/ClientServerInputMultiplexer.java
robertroeser: I set this way so that eventually it would get JIT'd away and you could only set it startup so you don't accidentally turn on frame logging.

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameFragmenter.java
robertroeser: not sure - the spec is a little unclear around fragmentation. If we send it last we can't generate the header until the last frame, and the metadata is supposed to come complete before the data - so how do you read it? Also I don't sent next_complete frames in the fragments i send next frames. Maybe we should change the spec because it seems more complicated to defer this till the end.

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameFragmenter.java
```diff
@@ -24,168 +34,212 @@
...
+            streamId,
+            true,
+            false,
+            RequestChannelFrameFlyweight.initialRequestN(frame),
+            metadataFragment,
+            dataFragment);
+        // Payload and synthetic types
+      case PAYLOAD:
+        return PayloadFrameFlyweight.encode(
+            allocator, streamId, true, false, false, metadataFragment, dataFragment);
```
mostroverkhov: `next` should be set - cant have both `next`  and `complete` false

rsocket-core/src/main/java/io/rsocket/RSocketServer.java
robertroeser: I  can probably remove the payload one. I might have added it when I was chasing done some memory leaks.

rsocket-core/src/main/java/io/rsocket/internal/ClientServerInputMultiplexer.java
yschimke: Your call, it's not common.

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameReassembler.java
robertroeser: I've gotten in the habit of doing this with publishers to maker sure exceptions don't slip by - its less about handling the error and more about making sure that I call sink.error less the exception disappears in the ether.

Issue #589, opened 2019-02-27
cannot connect to remote socket


zapphyre:
your example works, but I can not conect to some remote server:

```
WebsocketClientTransport ws = WebsocketClientTransport.create(URI.create("wss://ws.kraken.com"));
RSocket client = RSocketFactory.connect().keepAlive().transport(ws).start().block();
```

produces:
```
Caused by: java.lang.IllegalArgumentException: Unable to parse url ''
	at reactor.netty.http.client.UriEndpointFactory.createUriEndpoint(UriEndpointFactory.java:69) ~[reactor-netty-0.8.5.RELEASE.jar:0.8.5.RELEASE]
	at reactor.netty.http.client.UriEndpointFactory.createUriEndpoint(UriEndpointFactory.java:44) ~[reactor-netty-0.8.5.RELEASE.jar:0.8.5.RELEASE]
	at reactor.netty.http.client.HttpClientConnect$HttpClientHandler.<init>(HttpClientConnect.java:503) ~[reactor-netty-0.8.5.RELEASE.jar:0.8.5.RELEASE]
	at reactor.netty.http.client.HttpClientConnect$MonoHttpConnect.subscribe(HttpClientConnect.java:230) ~[reactor-netty-0.8.5.RELEASE.jar:0.8.5.RELEASE]
	at reactor.core.publisher.MonoMap.subscribe(MonoMap.java:55) ~[reactor-core-3.2.6.RELEASE.jar:3.2.6.RELEASE]
	at reactor.core.publisher.MonoFlatMap.subscribe(MonoFlatMap.java:60) ~[reactor-core-3.2.6.RELEASE.jar:3.2.6.RELEASE]
	at reactor.core.publisher.Mono.block(Mono.java:1493) ~[reactor-core-3.2.6.RELEASE.jar:3.2.6.RELEASE]
```

OlegDokuka:
@zapphyre, thank you for this issue. Fixed by #590 and merged into develop branch. You can find it in the latest snapshot

zapphyre:
thanks for fixing!
when release?

OlegDokuka:
Not sure. For now you can workaround it by adding additional / at the end of your link

zapphyre:
with trailing '/' I get:
```
java.lang.IllegalArgumentException: Frame type 28 is unknown
	at io.rsocket.framing.FrameType.fromEncodedType(FrameType.java:243)
[...]

PR #590, opened 2019-02-27
Bugfix/#589

OlegDokuka:
This PR provides bugfix for #589

### DIFF
--- a/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/client/WebsocketClientTransport.java
+++ b/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/client/WebsocketClientTransport.java
@@ -40,6 +40,8 @@ import reactor.netty.tcp.TcpClient;
  */
 public final class WebsocketClientTransport implements ClientTransport, TransportHeaderAware {
 
+  private static final String DEFAULT_PATH = "/";
+
   private final HttpClient client;
 
   private String path;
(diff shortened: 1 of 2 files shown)


rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/client/WebsocketClientTransport.java
```diff
@@ -130,6 +132,8 @@ public static WebsocketClientTransport create(HttpClient client, String path) {
     Objects.requireNonNull(client, "client must not be null");
     Objects.requireNonNull(path, "path must not be null");
 
+    path = path.startsWith(DEFAULT_PATH) ? path : (DEFAULT_PATH + path);
```
yschimke: Checking for "" seems safer than handling all non "/.*" cases.  Is it deliberate to convert something like ":443" to "/:443"

OlegDokuka: guess it is better to check this on regex for allowed symbols. Since it is possible to do something like "test" for the `path` field, which also must start with "/"

OlegDokuka: In general, path allows any symbols, so it does not matter. `URI#getPath` returns everything after "/" so if this is ":443" then, it is, probably, the right path. On the other hand, if it will be ":443", netty consider that as a `baseURL` and after some modification try to validate it against 
```
(?:(\w+)://)?((?:\[.+?])|(?<!\[)(?:[^/?]+?))(?::(\d{2,5}))?([/?].*)?
```

which fails validation.

See -> https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/http/client/HttpClient.java#L106
and -> https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L482

OlegDokuka: Thus, it is safe for any of possible cases

PR #591, opened 2019-02-28
added missing retain on metadata when reassembling frames

robertroeser:
Signed-off-by: Robert Roeser <rroeserr@gmail.com>

### DIFF
--- a/rsocket-transport-netty/src/test/java/io/rsocket/integration/FragmentTest.java
+++ b/rsocket-transport-netty/src/test/java/io/rsocket/integration/FragmentTest.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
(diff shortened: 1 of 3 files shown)


rsocket-transport-netty/src/test/java/io/rsocket/integration/FragmentTest.java
```diff
@@ -0,0 +1,171 @@
...
+public class FragmentTest {
+  private static final int frameSize = 128;
+  private AbstractRSocket handler;
+  private CloseableChannel server;
+  private String message = null;
+  private String metaData = null;
+
+  @BeforeEach
+  public void startup() {
+    int randomPort = ThreadLocalRandom.current().nextInt(10_000, 20_000);
```
yschimke: This seems somewhat flaky, particularly allocating ports in the non-ephemeral range.  IDeally this would be 0 and retrieved after starting.

PR #593, opened 2019-03-05
fixes issue with requestChannel never completes if server cancels (#592)

### LINKED ISSUE #592: requestChannel never completes if server cancels
Given the code below, the client hangs with the following output:

```
18:31:15.813 [main] INFO  r.F.C.1 - onSubscribe(FluxConcatArray.ConcatArraySubscriber)
18:31:15.827 [main] INFO  r.F.C.1 - request(unbounded)
```

```java

  CloseableChannel serverCloseable = RSocketFactory.receive()
      .acceptor((setup, sendingSocket) ->
          Mono.just(new AbstractRSocket() {

            @Override
            public Flux<Payload> requestChannel(Publisher<Payload> payloads) {
              return Flux.from(payloads).log().next()
                  .flatMapMany(in -> Flux.range(1, 3).map(i -> {
                    String data = in.getDataUtf8();
                    return DefaultPayload.create(data + "-" + i);
                  }));
            }
          }))
      .transport(TcpServerTransport.create("localhost", 7000))
      .start()
      .block();

  RSocket rsocket = RSocketFactory.connect()
      .dataMimeType(MimeTypeUtils.TEXT_PLAIN_VALUE)
      .transport(TcpClientTransport.create("localhost", 7000))
      .start()
      .block();

  Flux<Payload> payloadFlux = Flux.just(DefaultPayload.create("foo"), DefaultPayload.create("bar"));
  rsocket.requestChannel(payloadFlux).log().blockLast();

  rsocket.dispose();
  serverCloseable.dispose();
}
```

I think the scenario is invalid because the server cancels the input stream and then continues to produce output, but still the client probably shouldn't hang.

OlegDokuka:
This PR fixes #592 

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/RSocketClient.java
+++ b/rsocket-core/src/main/java/io/rsocket/RSocketClient.java
@@ -519,7 +519,6 @@ class RSocketClient implements RSocket {
         case CANCEL:
           {
             LimitableRequestPublisher sender = senders.remove(streamId);
-            receivers.remove(streamId);
             if (sender != null) {
               sender.cancel();
             }
--- a/rsocket-core/src/test/java/io/rsocket/RSocketClientTest.java
+++ b/rsocket-core/src/test/java/io/rsocket/RSocketClientTest.java
@@ -35,6 +35,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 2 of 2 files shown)


robertroeser:
this needs to be merged against the 1.0.x branch as well

OlegDokuka:
I will cherrypick that to 1.0.x as well

Issue #514, opened 2018-07-10
Possible OoME due to lack of coordination between Transport and RSocket


mostroverkhov:
Untrusted requsters may send large RequestN, and  if handler is fast compared to Transport, response frames queued on `UnboundedProcessor` may lead to large memory usage.

yschimke:
+1 

Given accounting is number of frame based, supporting max message sizes (fragmentation aware) sounds like a good tuning parameter.  I would expect to be able to size the memory of an rsocket server predictably for the number of supported clients and configured credits etc.

mostroverkhov:
That may help also. More specifically, my suggestion is to limit the number of frames on queue - `if `Subscription` does `request(1_000_000)` over network, I'd expect responder `Processor` does not request that 1m items blindly from `RSocketHandler`, but also accounts for `requests` from transport `DuplexConnection.send(Publisher)`.

yschimke:
Just in case you haven't hit it, we inherit some behaviour from reactor-core here also, i.e. it has some default behaviour of take(5) being limit(MAX) and cancel after 5.  limitRate can workaround that. 

Mentioning for context

mostroverkhov:
Yeah, I remember heated discussion on https://github.com/reactor/reactor-core/issues/879, sad it is not used here yet despite insistent demand. Still I believe it is slightly different problem here: `limitRequest` allows to tackle this problem for well-behaved requester, so there is no huge `requestN` over network, but one cannot assume untrusted client be well behaved. IMO now its to easy to blow up server with rogue request (even unintentionally)

yschimke:
@mostroverkhov I use it externally

https://github.com/rsocket/rsocket-cli/blob/1acfdea0d638cc6604eff53458602e586bd60a9e/src/main/kotlin/io/rsocket/cli/util.kt#L184-L185

robertroeser:
The LimitableRequestPublisher arbitrages between the underlying transport and the request n for the caller. It shouldn't be using up large amounts of memory because it shouldn't be processing data the transport buffers can handle it? We could try adding limitRate with a low and high water mark somewhere in the change to keep running smoother.

I guess the problem is that on the Java one then that would cause request n of 1,000,000 to be sent as request n of 128 over the network. That is probably fine but it might confuse people if they were to debug it.

OlegDokuka:
[...]

PR #595, opened 2019-03-12
fixes uncontrolled data sending in case of direct propagation of request from requester

OlegDokuka:
This PR provide a partial fix for #514 and reintroduce usage of `LimitableRequestPublisher` as of its initial design.

At a moment all connected `LimitableRequestPublisher` receives the same request size as from the connection. It means that upstream still an overflow UnboundProcessor but with less probability (it can happen in case we have super many requestChannel/requestStream connections which infinitely send data)

In the future PRs we have to revise the implementation of UnboundedProcessor in order to have proper guarantees on overflow prevention.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/RSocketServer.java
+++ b/rsocket-core/src/main/java/io/rsocket/RSocketServer.java
@@ -94,15 +94,15 @@ class RSocketServer implements ResponderRSocket {
     connection
         .send(
             sendProcessor.doOnRequest(
-                r ->
-                    sendingSubscriptions.forEach(
-                        (__, s) -> {
-                          if (s instanceof LimitableRequestPublisher) {
-                            LimitableRequestPublisher lrp = (LimitableRequestPublisher) s;
+                r -> {
+                  for (Subscription s : sendingSubscriptions.values()) {
+                    if (s instanceof LimitableRequestPublisher) {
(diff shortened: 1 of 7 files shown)


rsocket-core/src/main/java/io/rsocket/RSocketServer.java
```diff
@@ -92,7 +92,17 @@
     this.sendProcessor = new UnboundedProcessor<>();
 
     connection
-        .send(sendProcessor)
+        .send(
+            sendProcessor.doOnRequest(
+                r ->
+                    sendingSubscriptions.forEach(
```
bsideup: I would avoid `forEach` in performance critical paths

rsocket-core/src/test/java/io/rsocket/RSocketClientTest.java
```diff
@@ -216,6 +219,32 @@ public void testChannelRequestServerSideCancellation() {
     Assertions.assertThat(request.isDisposed()).isTrue();
   }
 
+  @Test(timeout = 2_000900)
```
bsideup: 2000 seconds? A typo?

OlegDokuka: fixed

PR #598, opened 2019-03-17
General performance improvements

OlegDokuka:
## This PR provides a list of dramatical performance improvements:

1) Reduce the number of used Atomics within the code base. As you can see in PR, there were a few places with applied `first` elements strategy. In all of those cases, usage of atomic was redundant since Reactive-Streams spec and Reactors implementation guarantees happens-before and [serialized](https://github.com/reactive-streams/reactive-streams-java#term_serially) access to the object (Subscriber / Subscription) so having an additional object with volatile access is redundant and give more overhead
2) Reduce the number of allocated objects. In many cases, we use too many Reactors operators where actual logic can be implemented within one `BaseSubscriber`. Thus, this PR eliminates redundant transformations and minimize function flow. Also, dynamic invocations give its overhead, so there are upcoming changes in Reactor 3 (https://github.com/reactor/reactor-core/issues/1556) that improves performance as well.
3) Simplifies Lifecycle to more specific cases, so the number of used operators (hence allocated objects) also is reduced to the minimum.

## General bugfixes and improvements

Along with performance improvements, this PR introduces Jmh tests so you can run benchmarks locally and observe 10-20% perf improvements comparing to the previous implementation.

Finally, this PR provides a couple of bugfixes found during development and testing 

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/RSocketClient.java
+++ b/rsocket-core/src/main/java/io/rsocket/RSocketClient.java
@@ -41,7 +41,6 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.SignalType;
 import reactor.core.publisher.UnicastProcessor;
-import reactor.util.concurrent.Queues;
 
 /** Client Side of a RSocket socket. Sends {@link ByteBuf}s to a {@link RSocketServer} */
 class RSocketClient implements RSocket {
@@ -226,8 +225,7 @@ class RSocketClient implements RSocket {
           int streamId = streamIdSupplier.nextStreamId();
 
           final UnboundedProcessor<ByteBuf> sendProcessor = this.sendProcessor;
-          final UnicastProcessor<Payload> receiver =
-              UnicastProcessor.create(Queues.<Payload>one().get());
+          final UnicastProcessor<Payload> receiver = UnicastProcessor.create();
 
           receivers.put(streamId, receiver);
 
@@ -277,8 +275,7 @@ class RSocketClient implements RSocket {
     return lifecycle.activeFlux(
         () -> {
           final UnboundedProcessor<ByteBuf> sendProcessor = this.sendProcessor;
-          final UnicastProcessor<Payload> receiver =
-              UnicastProcessor.create(Queues.<Payload>one().get());
+          final UnicastProcessor<Payload> receiver = UnicastProcessor.create();
           final int streamId = streamIdSupplier.nextStreamId();
 
           return receiver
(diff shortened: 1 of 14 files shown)


rsocket-core/src/main/java/io/rsocket/RSocketClient.java
```diff
@@ -490,8 +493,7 @@ private void handleFrame(int streamId, FrameType type, ByteBuf frame) {
             LimitableRequestPublisher sender = senders.get(streamId);
             if (sender != null) {
               int n = RequestNFrameFlyweight.requestN(frame);
-              sender.increaseRequestLimit(n);
-              sendProcessor.drain();
+              sender.request(n >= Integer.MAX_VALUE ? Long.MAX_VALUE : n);
```
yschimke: I'm trying to remember where we specify this rule?  Integer.MAX_VALUE == Inifinite?  Reactor? Reactive Streams? RSocket?

rsocket-core/src/test/java/io/rsocket/fragmentation/FragmentationDuplexConnectionTest.java
```diff
@@ -125,7 +124,7 @@ void reassembleData() {
         .assertNext(
             byteBuf -> {
               Assert.assertEquals(data, RequestResponseFrameFlyweight.data(byteBuf));
-              ReferenceCountUtil.safeRelease(byteBuf);
+              //              ReferenceCountUtil.safeRelease(byteBuf);
```
yschimke: deliberate?

yschimke:
Nice perf win!

rsocket-core/src/main/java/io/rsocket/RSocketClient.java
OlegDokuka: https://github.com/rsocket/rsocket-java/blob/develop/rsocket-core/src/main/java/io/rsocket/RSocketServer.java#L469

rsocket-core/src/test/java/io/rsocket/fragmentation/FragmentationDuplexConnectionTest.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/RSocketClient.java
yschimke: I guess what I'm asking is whether this rule needs to migrate into the spec

OlegDokuka: If it is not clarify that Int. Max Value should be considered as unbounded stream, then we have to mention that

PR #606, opened 2019-03-27
update tcp performance test with resume case

mostroverkhov:


robertroeser:
@mostroverkhov looks good can fix DCO?

PR #609, opened 2019-03-28
resume: reject setup if there is existing session for provided token

mostroverkhov:


robertroeser:
I don't know if this is always valid - the session token isn't necessarily unique - like a session token tied to a particular users.

mostroverkhov:
spec prohibits that, and It is intuitive - there must be at most one session for given token: either reconnect with existing token, or start new session with new one.

mostroverkhov:
this is out of sync with ongoing "byte-count for positions" PR - I will reapply changes there

PR #610, opened 2019-04-01
fixes bug with incorrect WS framesize setup.

OlegDokuka:
This PR provides a WebSocket transport's frame size setup with regards
of current RSocket fragment size.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/frame/FrameUtil.java
+++ b/rsocket-core/src/main/java/io/rsocket/frame/FrameUtil.java
@@ -5,7 +5,6 @@ import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.Unpooled;
 
 public class FrameUtil {
-  public static final int FRAME_MAX_SIZE = 16_777_215;
 
   private FrameUtil() {}
 
--- a/rsocket-core/src/main/java/io/rsocket/frame/FrameUtil.java
+++ b/rsocket-core/src/main/java/io/rsocket/frame/FrameUtil.java
@@ -5,6 +5,8 @@ import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.Unpooled;
 
 public class FrameUtil {
+  public static final int FRAME_MAX_SIZE = 16_777_215;
+
   private FrameUtil() {}
 
   public static String toString(ByteBuf frame) {
--- a/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketRouteTransport.java
+++ b/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketRouteTransport.java
@@ -16,7 +16,7 @@
 
 package io.rsocket.transport.netty.server;
 
-import static io.rsocket.frame.FrameUtil.FRAME_MAX_SIZE;
+import static io.rsocket.frame.FrameLengthFlyweight.FRAME_LENGTH_MASK;
 
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.handler.codec.http.HttpMethod;
@@ -87,7 +87,7 @@ public final class WebsocketRouteTransport implements ServerTransport<Closeable>
                     return acceptor.apply(connection).then(out.neverComplete());
                   },
                   null,
-                  Math.max(DEFAULT_FRAME_SIZE, mtu == 0 ? FRAME_MAX_SIZE : mtu));
+                  Math.max(DEFAULT_FRAME_SIZE, mtu == 0 ? FRAME_LENGTH_MASK : mtu));
             })
         .bind()
         .map(CloseableChannel::new);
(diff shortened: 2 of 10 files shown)


rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketRouteTransport.java
```diff
@@ -74,9 +85,132 @@ public WebsocketRouteTransport(
...
-                  });
+                  },
+                  null,
+                  Math.max(DEFAULT_FRAME_SIZE, mtu == 0 ? FRAME_MAX_SIZE : mtu));
             })
         .bind()
         .map(CloseableChannel::new);
   }
+
+  static final class UriPathTemplate {
```
OlegDokuka: This is port from Reactor-Netty internals 

cc/
 @violetagg

rsocket-core/src/main/java/io/rsocket/frame/FrameUtil.java
```diff
@@ -5,6 +5,8 @@
 import io.netty.buffer.Unpooled;
 
 public class FrameUtil {
+  public static final int FRAME_MAX_SIZE = 16_777_215;
```
OlegDokuka: @robertroeser I did not find a place which limits `mtu` to that size? Is that limitation still valid?

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketRouteTransport.java
```diff
@@ -74,9 +85,132 @@ public WebsocketRouteTransport(
...
+    private static final Pattern FULL_SPLAT_PATTERN = Pattern.compile("[\\*][\\*]");
+    private static final String FULL_SPLAT_REPLACEMENT = ".*";
+
+    private static final Pattern NAME_SPLAT_PATTERN = Pattern.compile("\\{([^/]+?)\\}[\\*][\\*]");
+    // JDK 6 doesn't support named capture groups
+    private static final String NAME_SPLAT_REPLACEMENT = "(?<%NAME%>.*)";
+    // private static final String  NAME_SPLAT_REPLACEMENT = "(.*)";
+
+    private static final Pattern NAME_PATTERN = Pattern.compile("\\{([^/]+?)\\}");
+    // JDK 6 doesn't support named capture groups
```
yschimke: We are JDK8, so can be cleaned up.  Generally anything on Spring Reactor is JDK8 IIUC

rsocket-core/src/main/java/io/rsocket/frame/FrameUtil.java
robertroeser: https://github.com/rsocket/rsocket-java/blob/develop/rsocket-core/src/main/java/io/rsocket/frame/FrameLengthFlyweight.java#L17

robertroeser: Can you switch the length to  0xFFFFFF?

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketRouteTransport.java
robertroeser: Yes - the java version of RSocket only supports 1.8 are greater so we don't need to worry about anything in JDK 1.6

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketServerTransport.java
```diff
@@ -114,6 +118,8 @@ public void setTransportHeaders(Supplier<Map<String, String>> transportHeaders)
             (request, response) -> {
               transportHeaders.get().forEach(response::addHeader);
               return response.sendWebsocket(
+                  null,
+                  Math.max(DEFAULT_FRAME_SIZE, mtu == 0 ? FRAME_MAX_SIZE : mtu),
```
robertroeser: what is this doing? Is there a reason we are setting the max frame size to anything but rsocket max frame size?

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketRouteTransport.java
OlegDokuka: good catchup, I did not notice that when copy past that :D

rsocket-core/src/main/java/io/rsocket/frame/FrameUtil.java
OlegDokuka: Done. Uses `FRAME_LENGTH_MASK` from `FrameLengthFlyweight`

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketServerTransport.java
OlegDokuka: So, the entire fragmentation logic is built on top of the assumption that mtu is higher than 0. So if `mtu` is 0, then fragmentation is avoided. It means that max size of the fragment should be less || equal to maximum possible size frame. At least I was thinking like that when writing that code. In case 0xFFFFFF is a specific limit to fragmentation and in general payload size is unlimited (or limited to `(Integer.MAX_VALUE - 2)`), then I can fix that.

OlegDokuka: simplified to max frame size for all possible cases

PR #611, opened 2019-04-01
Switch resume position tracking to byte count

mostroverkhov:
Also some simplification & performance improvements


yschimke:
I think this deserves more discussion. RSocket cpp implements this. But it wasn’t clear in the spec IIRC that it should be bytes vs frame counts. And RSocket js went with frame counts. The complexity of bytes vs frame counts doesn’t seem like a win. It’s a terrible checksum and frame counting seems like a better way forward

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mostroverkhov:
@yschimke positions are little simpler to implement. There are multiple PRs hanging on spec with opposite propositions. I think if the goal is having all 3 implementations aligned, It is easier to update java & js complared to C++

yschimke:
No objection from me if all converge. And spec is clear on bytes vs frames. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

robertroeser:
@yschimke can you fix your pull request that makes bytes explicit and then we can update the rsocket spec.

yschimke:
@robertroeser done, but it's still a bit awkward.  So critical extension frames still skip counting which doesn't seem great for a client and server staying in sync in reality.  There are still holes here...

https://github.com/rsocket/rsocket/pull/236

yschimke:
I don't think next iteration of this part of the spec should come from me, but I predict you will need one :)

robertroeser:
@yschimke I merged it - we could extent the extension frames. We could add a flag to the extension frames that indicates that it is resumable.

PR #614, opened 2019-04-09
Feature/tuple frames

OlegDokuka:
This PR provides `TupleXByteBuf` as a replacement for `CompositByteBuf`

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/buffer/Tuple2ByteBuf.java
+++ b/rsocket-core/src/main/java/io/rsocket/buffer/Tuple2ByteBuf.java
@@ -1,0 +1,394 @@
+package io.rsocket.buffer;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.Unpooled;
+import io.netty.util.ReferenceCountUtil;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.channels.GatheringByteChannel;
+import java.nio.charset.Charset;
+
+class Tuple2ByteBuf extends AbstractTupleByteBuf {
+
+  private static final long ONE_MASK = 0x100000000L;
+  private static final long TWO_MASK = 0x200000000L;
+  private static final long MASK = 0x700000000L;
+
+  private final ByteBuf one;
+  private final ByteBuf two;
+  private final int oneReadIndex;
+  private final int twoReadIndex;
+  private final int oneReadableBytes;
+  private final int twoReadableBytes;
+  private final int twoRelativeIndex;
+
+  private boolean freed;
+
+  Tuple2ByteBuf(ByteBufAllocator allocator, ByteBuf one, ByteBuf two) {
+    super(allocator, one.readableBytes() + two.readableBytes());
+
+    this.one = one;
+    this.two = two;
+
+    this.oneReadIndex = one.readerIndex();
+    this.twoReadIndex = two.readerIndex();
+
+    this.oneReadableBytes = one.readableBytes();
+    this.twoReadableBytes = two.readableBytes();
+
+    this.twoRelativeIndex = oneReadableBytes;
+
+    this.freed = false;
+  }
+
+  long calculateRelativeIndex(int index) {
+    checkIndex(index, 0);
+
+    long relativeIndex;
+    long mask;
+    if (index >= twoRelativeIndex) {
+      relativeIndex = twoReadIndex + (index - oneReadableBytes);
+      mask = TWO_MASK;
+    } else {
+      relativeIndex = oneReadIndex + index;
+      mask = ONE_MASK;
+    }
+
+    return relativeIndex | mask;
+  }
+
+  ByteBuf getPart(int index) {
+    long ri = calculateRelativeIndex(index);
+    switch ((int) ((ri & MASK) >>> 32L)) {
+      case 0x1:
+        return one;
+      case 0x2:
+        return two;
+      default:
+        throw new IllegalStateException();
+    }
+  }
+
+  @Override
+  public boolean isDirect() {
+    return one.isDirect() && two.isDirect();
+  }
+
+  @Override
+  public int nioBufferCount() {
+    return one.nioBufferCount() + two.nioBufferCount();
+  }
+
+  @Override
+  public ByteBuffer nioBuffer() {
+    ByteBuffer[] oneBuffers = one.nioBuffers();
+    ByteBuffer[] twoBuffers = two.nioBuffers();
+
+    ByteBuffer merged =
+        BufferUtil.allocateDirectAligned(capacity, DEFAULT_DIRECT_MEMORY_CACHE_ALIGNMENT)
+            .order(order());
+
+    for (ByteBuffer b : oneBuffers) {
+      merged.put(b);
+    }
+
+    for (ByteBuffer b : twoBuffers) {
+      merged.put(b);
+    }
+
+    merged.flip();
+    return merged;
+  }
+
+  @Override
+  public ByteBuffer[] nioBuffers(int index, int length) {
+    if (length == 0) {
+      return new ByteBuffer[] {EMPTY_NIO_BUFFER};
+    }
+
+    long ri = calculateRelativeIndex(index);
+    index = (int) (ri & Integer.MAX_VALUE);
+    switch ((int) ((ri & MASK) >>> 32L)) {
+      case 0x1:
+        ByteBuffer[] oneBuffer;
+        ByteBuffer[] twoBuffer;
+        int l = Math.min(oneReadableBytes - index, length);
+        oneBuffer = one.nioBuffers(index, l);
+        length -= l;
+        if (length != 0) {
+          twoBuffer = two.nioBuffers(twoReadIndex, length);
+          ByteBuffer[] results = new ByteBuffer[oneBuffer.length + twoBuffer.length];
+          System.arraycopy(oneBuffer, 0, results, 0, oneBuffer.length);
+          System.arraycopy(twoBuffer, 0, results, oneBuffer.length, twoBuffer.length);
+          return results;
+        } else {
+          return oneBuffer;
+        }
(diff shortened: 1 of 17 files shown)


rsocket-core/src/test/java/io/rsocket/frame/RequestFlyweightTest.java
```diff
@@ -40,7 +40,7 @@ void testEncodingWithEmptyMetadata() {
 
     frame = FrameLengthFlyweight.encode(ByteBufAllocator.DEFAULT, frame.readableBytes(), frame);
 
-    assertEquals("00000e0000000119000000000100000064", ByteBufUtil.hexDump(frame));
+    assertEquals("00000b0000000118000000000164", ByteBufUtil.hexDump(frame));
```
robertroeser: The value for this test in the develop branch is the correct value. Changing to this value is a bug.

rsocket-core/src/main/java/io/rsocket/frame/DataAndMetadataFlyweight.java
```diff
@@ -32,11 +33,13 @@ private static int decodeLength(final ByteBuf byteBuf) {
...
   static ByteBuf encodeOnlyMetadata(
       ByteBufAllocator allocator, final ByteBuf header, ByteBuf metadata) {
-    return allocator.compositeBuffer(2).addComponents(true, header, metadata);
+    // return allocator.compositeBuffer(2).addComponents(true, header, metadata);
+    return TupleByteBuffs.of(allocator, header, metadata);
   }
 
   static ByteBuf encodeOnlyData(ByteBufAllocator allocator, final ByteBuf header, ByteBuf data) {
-    return allocator.compositeBuffer(2).addComponents(true, header, data);
+    // return allocator.compositeBuffer(2).addComponents(true, header, data);
```
robertroeser: remove these commented fields

rsocket-core/src/main/java/io/rsocket/frame/ExtensionFrameFlyweight.java
```diff
@@ -16,13 +17,16 @@ public static ByteBuf encode(
 
     int flags = FrameHeaderFlyweight.FLAGS_I;
 
-    if (metadata != null) {
+    if (metadata != null && metadata != Unpooled.EMPTY_BUFFER) {
```
robertroeser: Don't change any of the frame logic with this pull request expect switching to TupleByteBuf - the branch original comes from an older branch and we should defer the the framing logic to the develop branch

rsocket-core/src/test/java/io/rsocket/buffer/Tuple3ByteBufTest.java
```diff
@@ -0,0 +1,24 @@
+package io.rsocket.buffer;
+
+import static io.netty.buffer.Unpooled.wrappedBuffer;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
+import org.junit.Ignore;
+
+@Ignore
+public class Tuple3ByteBufTest extends AbstractByteBufTest {
```
robertroeser: don't ignore the tests - make sure they work

rsocket-core/build.gradle
```diff
@@ -30,7 +30,9 @@ dependencies {
     implementation 'org.slf4j:slf4j-api'
 
     compileOnly 'com.google.code.findbugs:jsr305'
-
+    
+    implementation 'org.agrona:agrona:0.9.33'
```
robertroeser: We probably shouldn't pull this in - we can just add the BufferUtil.allocateDirectAligned method into a util class in RSocket so we don't need a new dependency.

rsocket-core/src/main/java/io/rsocket/buffer/Tuple2ByteBuf.java
```diff
@@ -0,0 +1,415 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.channels.GatheringByteChannel;
+import java.nio.charset.Charset;
+import org.agrona.BufferUtil;
+
+class Tuple2ByteBuf extends AbstractTupleByteBuf {
+  private static final int MEMORY_CACHE_ALIGNMENT = 64;
```
robertroeser: this should be changed to 8

rsocket-core/src/main/java/io/rsocket/buffer/Tuple2ByteBuf.java
```diff
@@ -0,0 +1,415 @@
...
+    this.oneReadableBytes = one.readableBytes();
+    this.twoReadableBytes = two.readableBytes();
+
+    this.twoRelativeIndex = oneReadableBytes;
+
+    this.freed = false;
+  }
+
+  public long calculateRelativeIndex(int index) {
+    if (CHECK_BOUNDS && index > capacity) {
```
robertroeser: We should be able to disable bounds checking - this should be a property. I believe there could be a property netty uses for its bounds checking - we might want to use use the same one.

rsocket-core/src/main/java/io/rsocket/buffer/Tuple3ByteBuf.java
```diff
@@ -0,0 +1,597 @@
...
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.channels.GatheringByteChannel;
+import java.nio.charset.Charset;
+import org.agrona.BufferUtil;
+
+class Tuple3ByteBuf extends AbstractTupleByteBuf {
+
+  private static final int MEMORY_CACHE_ALIGNMENT = 64;
```
robertroeser: same as above - make this a property and default it to 8

rsocket-core/src/test/java/io/rsocket/frame/RequestFlyweightTest.java
OlegDokuka: I got an issue on the decoding side in case we send `EMPTY` `bytebuf` metadata, so the decoding side seeing meta flag tries to read another one byte and gets the index out of bound exception

rsocket-core/src/main/java/io/rsocket/frame/ExtensionFrameFlyweight.java
OlegDokuka: I'm just wondering whether empty metadata can be counted as "we have metadata, but it is empty"

rsocket-core/src/test/java/io/rsocket/buffer/Tuple3ByteBufTest.java
OlegDokuka: I tried to port tests from netty but it seems that I don't want to fix them all for now. I will try to cut testsuit a little to the minimum we need right now

rsocket-core/build.gradle
OlegDokuka: Alright

rsocket-core/src/main/java/io/rsocket/buffer/Tuple2ByteBuf.java
OlegDokuka: rigth

rsocket-core/src/main/java/io/rsocket/buffer/Tuple3ByteBuf.java
OlegDokuka: What does that magic number do?

rsocket-core/src/test/java/io/rsocket/frame/RequestFlyweightTest.java
robertroeser: yeah - I think you're pulling in older framing logic or something. The value 00000e0000000119000000000100000064 is correct

rsocket-core/src/main/java/io/rsocket/buffer/Tuple3ByteBuf.java
robertroeser: https://stackoverflow.com/questions/381244/purpose-of-memory-alignment

rsocket-core/src/test/java/io/rsocket/frame/RequestFlyweightTest.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/frame/DataAndMetadataFlyweight.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/frame/ExtensionFrameFlyweight.java
OlegDokuka: fixed

rsocket-core/src/test/java/io/rsocket/buffer/Tuple3ByteBufTest.java
OlegDokuka: fixed

rsocket-core/build.gradle
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/buffer/Tuple2ByteBuf.java
OlegDokuka: fixed. Now I'm using native Netty property `SystemPropertyUtil.getInt("io.netty.allocator.directMemoryCacheAlignment", 0);`

rsocket-core/src/main/java/io/rsocket/buffer/Tuple2ByteBuf.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/buffer/Tuple3ByteBuf.java
OlegDokuka: done

Issue #615, opened 2019-04-11
Exception while re-assembling Fragment


robertroeser:
When fragmentation is enabled, depending on the size of the payload and MTU, there is a refCnt exception prevent fragmentation from completing.

PR #616, opened 2019-04-11
fixes issue with fragmentation when the MTU size is set to 64 bytes

robertroeser:
Fix for issue #615

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/frame/decoder/DefaultPayloadDecoder.java
+++ b/rsocket-core/src/main/java/io/rsocket/frame/decoder/DefaultPayloadDecoder.java
@@ -11,7 +11,7 @@ import java.nio.ByteBuffer;
 class DefaultPayloadDecoder implements PayloadDecoder {
 
   @Override
-  public Payload apply(ByteBuf byteBuf) {
+  public synchronized Payload apply(ByteBuf byteBuf) {
     ByteBuf m;
     ByteBuf d;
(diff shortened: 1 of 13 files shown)


rsocket-core/src/main/java/io/rsocket/frame/decoder/DefaultPayloadDecoder.java
```diff
@@ -11,7 +11,7 @@
 class DefaultPayloadDecoder implements PayloadDecoder {
 
   @Override
-  public Payload apply(ByteBuf byteBuf) {
+  public synchronized Payload apply(ByteBuf byteBuf) {
```
SerCeMan: Hi, @robertroeser! I've run a benchmark with multiple clients sending messages concurrently, and lock profiler (a part of [async-profiler](https://github.com/jvm-profiling-tools/async-profiler)) showed that there was lock contention in `DefaultPayloadDecoder` which seems to be introduced in this PR. 

Screen Shot 2019-11-29 at 6 42 24 pm

I've looked through the code, but it's not obvious to me why it requires synchronisation since the class seems to be stateless. So, I'm wondering whether this `synchronized` can be safely removed in which case I'll be happy to send a PR or it serves some reason. 

Thanks :)

OlegDokuka: Hello, @SerCeMan!

Thanks for finding that. I guess that was done by mistake and there is no need to have `synchronized` keyword here.

PR #618, opened 2019-04-15
fix issue when long running channels eventually hang

mostroverkhov:


yschimke:
I’ll test tomorrow. Thanks for the fix

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mostroverkhov:
@yschimke credit goes to Robert here..

PR #623, opened 2019-04-18
Refresh sockets on `select` in LoadBalancedRSocketMono before checking for active sockets

### LINKED ISSUE #622: Load balancer doesn't recover from supplier pool going down to 0 instances
I've been attempting to use the `LoadBalancedRSocketMono` class by providing it a `Publisher<Collection<RSocketSupplier>>` that may deliver zero to many suppliers at any given point. I wanted to prove that if we lost all instances of a service, the load balancer would recover when new instances came online. 

In its current form, the load balancer will default to returning a `FailingRSocket` if there are no active sockets available, which makes sense, but if new sockets do become available in the pool, the load balancer doesn't know about them.

I was able to resolve this by moving the `refreshSockets` call in the `select` method above the check for empty active sockets. This allows the load balancer to check for new instances in the supplier pool before returning the static FailingRSocket.

_Another potential solution would be to give the pool some type of callback so the load balancer is notified when new suppliers are delivered._

I'll have a PR shortly that addresses this with the former solution.

jareddellitt:
Fixes #622 

Also cleaned up a few things in the test that weren't being used.

### DIFF
--- a/rsocket-load-balancer/src/main/java/io/rsocket/client/LoadBalancedRSocketMono.java
+++ b/rsocket-load-balancer/src/main/java/io/rsocket/client/LoadBalancedRSocketMono.java
@@ -374,10 +374,11 @@ public abstract class LoadBalancedRSocketMono extends Mono<RSocket>
   }
 
   private synchronized RSocket select() {
+    refreshSockets();
+
     if (activeSockets.isEmpty()) {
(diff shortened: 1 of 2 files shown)


robertroeser:
@jareddellitt - thanks for the pull request

PR #630, opened 2019-05-14
Internal backpressure control

OlegDokuka:


rsocket-core/src/main/java/io/rsocket/util/BackpressureUtils.java
```diff
@@ -0,0 +1,29 @@
...
+
+  public static void shareRequest(
+      long requested,
+      SynchronizedIntObjectHashMap<LimitableRequestPublisher> limitableSubscriptions) {
+    try {
+      if (limitableSubscriptions.isEmpty()) {
+        return;
+      }
+      Object[] activeSubscriptions;
+      activeSubscriptions = limitableSubscriptions.getValuesCopy();
```
robertroeser: copy the values each time you read isn't exceptionable and will lead to bad performance.

PR #638, opened 2019-05-20
Fix bug when process OOMs due to retained LimitableRequestPublishers in RSocketServer

mostroverkhov:


rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/TcpServerTransport.java
```diff
@@ -114,6 +118,9 @@ public static TcpServerTransport create(TcpServer server) {
               acceptor.apply(connection).then(Mono.<Void>never()).subscribe(c.disposeSubscriber());
             })
         .bind()
+        .doOnNext(
```
smaldini: Do we need to do that ? I'm wondering because we are opening the door to a new way to retain channels:
- Is there a use case where the server life is shorter than the application life ?
- If so, should we just shutdown event loops to force channels to be closed ?
- Is this related at all to the OOM since this is only doing anything when the server is closed ?

mostroverkhov: @smaldini This is not related to OOM - just annoying bug living in reactor-netty for almost year, getting in my way when debugging. This change is my interpretation of what suggested as workaround by violeta at the bottom of https://github.com/reactor/reactor-netty/issues/495. If I close Rsocket's server `CloseableChannel`, I d expect connections -> rsockets are closed aswell. The reason many changes are squashed in PR is there exist long-running soak test, and we would like to run It fewer times for this release

PR #639, opened 2019-05-20
Fix logging formatting

yschimke:
The single arg form with a Throwable is a pre formatted message in slf4j

```
21:19:34.781	Starting session timeout due to error: {}
java.nio.channels.ClosedChannelException
	at io.rsocket.resume.ResumableDuplexConnection.disconnect(ResumableDuplexConnection.java:326)
```


mostroverkhov:
can you sign a commit to pass DCO check?

Issue #640, opened 2019-05-21
Server dies on keep-alive ack timeout


lksvenoy-r7:
The rsocket server dies on keep-alive ack timeout. I've tried adding onErrorResume, but to no avail. How can I prevent the server from closing its socket on error?

**Error**
```
[2019-05-20 11:12:21.438] ERROR [parallel-1] RegistryRSocketServer: Error occurred during session
io.rsocket.exceptions.ConnectionErrorException: No keep-alive acks for 60000 ms
at io.rsocket.keepalive.KeepAliveConnection.lambda$startKeepAlives$1(KeepAliveConnection.java:97)
at reactor.core.publisher.LambdaMonoSubscriber.onNext(LambdaMonoSubscriber.java:137)
at reactor.core.publisher.Operators$MonoSubscriber.complete(Operators.java:1476)
at reactor.core.publisher.MonoProcessor.onNext(MonoProcessor.java:389)
at io.rsocket.keepalive.KeepAliveHandler.doCheckTimeout(KeepAliveHandler.java:112)
at io.rsocket.keepalive.KeepAliveHandler$Server.onIntervalTick(KeepAliveHandler.java:128)
at io.rsocket.keepalive.KeepAliveHandler.lambda$start$0(KeepAliveHandler.java:63)
at reactor.core.publisher.LambdaSubscriber.onNext(LambdaSubscriber.java:130)
at reactor.core.publisher.FluxInterval$IntervalRunnable.run(FluxInterval.java:123)
at reactor.core.scheduler.PeriodicWorkerTask.call(PeriodicWorkerTask.java:59)
at reactor.core.scheduler.PeriodicWorkerTask.run(PeriodicWorkerTask.java:73)
at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
at java.base/java.lang.Thread.run(Thread.java:834)

io.rsocket.exceptions.ConnectionErrorException: No keep-alive acks for 60000 ms
[...]

PR #643, opened 2019-05-28
Rename internal components to match their purpose

mostroverkhov:
`RSocketClient` and `RSocketServer` names for requester and responder are misleading because each side of connection has both, and they can be confused with client and server of `RSocketFactory`. Happened more than once already so It makes sense to rename them appropriately https://github.com/rsocket/rsocket-java/issues/640#issuecomment-494313461.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
+++ b/rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
@@ -145,14 +145,14 @@ public class RSocketFactory {
       return this;
     }
 
-    /** Deprecated. Use {@link #addHandlerPlugin(RSocketInterceptor)} instead */
+    /** Deprecated. Use {@link #addResponderPlugin(RSocketInterceptor)} instead */
     @Deprecated
     public ClientRSocketFactory addServerPlugin(RSocketInterceptor interceptor) {
-      return addHandlerPlugin(interceptor);
+      return addResponderPlugin(interceptor);
     }
 
-    public ClientRSocketFactory addHandlerPlugin(RSocketInterceptor interceptor) {
-      plugins.addHandlerPlugin(interceptor);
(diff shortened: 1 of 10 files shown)


rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
```diff
@@ -134,14 +134,25 @@ public ClientRSocketFactory addConnectionPlugin(DuplexConnectionInterceptor inte
...
     }
 
+    /** Deprecated. Use {@link #addHandlerPlugin(RSocketInterceptor)} instead */
+    @Deprecated
     public ClientRSocketFactory addServerPlugin(RSocketInterceptor interceptor) {
-      plugins.addServerPlugin(interceptor);
+      return addHandlerPlugin(interceptor);
+    }
+
+    public ClientRSocketFactory addHandlerPlugin(RSocketInterceptor interceptor) {
```
robertroeser: Should this be addResponderPlugin?

rsocket-core/src/main/java/io/rsocket/plugins/PluginRegistry.java
```diff
@@ -23,39 +23,63 @@
...
+  }
+
+  /** Deprecated. Use {@link #addHandlerPlugin(RSocketInterceptor)} instead */
+  @Deprecated
   public void addServerPlugin(RSocketInterceptor interceptor) {
-    servers.add(interceptor);
+    addHandlerPlugin(interceptor);
+  }
+
+  public void addHandlerPlugin(RSocketInterceptor interceptor) {
```
robertroeser: should be this be addResponderPlugin

robertroeser:
Looks good - in some places you have RSocketRequest and some other places you have RSocketHandler. Should these be RSocketResponder instead to like up with the other Responder? Handler doens't exist anyway in the spec

Issue #641, opened 2019-05-21
Missing onComplete in requestStream and requestChannel with fragmentation


lksvenoy-r7:
When using fragmentation, the onComplete signal is sometimes missing with requestStream and requestChannel. This has been observed using many different fragmentation values (from 64 and up), and the behaviour is different locally than it is when using it over internet.

**Expected Behaviour:** When fragmentation is enabled, granted there are no errors, the entire payload is received by the client and the server finished by sending the COMPLETE frame.

**Actual Behaviour:** When fragmentation is enabled, when communicating over the internet, the socket is inconsistent, sometimes producing all payloads and the completion signal, and sometimes the payload amount varies, and no completion signal is issued.

**Steps to Reproduce**

1. Create an rsocket server using the websocket transport protocol on an external server
2. Create an rsocket client using the websocket transport protocol.
3. Enable fragmentation on both server and client
4. Send a big item and observe the onCompletion signal never arriving

**Additional Information**
Rsocket Version: 0.12.2-RC2
Infrastructure (Amazon): Rsocket Client -> Kong ALB -> Kong -> Service ALB -> Service

The behaviour works as expected when testing locally, but not over the internet. The load balancers have been tested and are not the issue. 

**Workaround**
The current workaround we are using is to disable fragmentation, which resolves the issue.

If you require additional information, please let me know.

robertroeser:
Haven't gotten a chance to look at this in depth but there is a change this could be related to some structural changes in 0.12.2-RC2. Have you had a chance to try this with 0.12.2-RC3?

lksvenoy-r7:
> Haven't gotten a chance to look at this in depth but there is a change this could be related to some structural changes in 0.12.2-RC2. Have you had a chance to try this with 0.12.2-RC3?

I haven't gotten a chance to look at this yet as I am too busy at this moment, but I'll update this ticket once it becomes relevant again. I had previously enabled it to fix an issue where my payload was too big, but this issue went away in a later update so fragmentation was no longer necessary for us.

robertroeser:
Ok - thanks for checking - @mostroverkhov is looking into this - we'll let you know.

robertroeser:
@lksvenoy-r7 did you get a chance to try @mostroverkhov fix?

rstoyanchev:
I can't reproduce this. The following:
```java
Payload payload =
    DefaultPayload.create(Flux.just("a").repeat(1024 * 1024).collect(Collectors.joining()).block());

[...]

PR #645, opened 2019-05-30
Fix leaks and issues related to fragmentation

mostroverkhov:
Fix memory leaks on frame reassembly.
Fix empty data on reassembly in some (payload size/mtu) combinations.
Server signals error on binding if incorrect fragmentation mtu is set.

Relates to https://github.com/rsocket/rsocket-java/issues/641


rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/TcpServerTransport.java
```diff
@@ -94,26 +94,30 @@ public static TcpServerTransport create(TcpServer server) {
...
-                        true,
-                        "server");
-              } else {
-                connection = new TcpDuplexConnection(c);
-              }
-              acceptor.apply(connection).then(Mono.<Void>never()).subscribe(c.disposeSubscriber());
-            })
-        .bind()
-        .map(CloseableChannel::new);
+    return FragmentationDuplexConnection.checkMtu(mtu)
```
mostroverkhov: This check could be done on RSocketFactory, added here for parity with client signalling error on connect.

mostroverkhov:
java11 build failed with javac related error - have same issue on one of my boxes with 11.0.3. Java8 and 12 are green.

robertroeser:
> java11 build failed with javac related [error](https://travis-ci.org/rsocket/rsocket-java/jobs/539361289) - have same issue on one of my boxes with 11.0.3. Java8 and 12 are green.

I've found that the builds are a bit flaky in the other jdks and you can restart them and they work.

mostroverkhov:
Did not help. I need to fiddle with commits to workaround that

mostroverkhov:
@robertroeser added workaround - there must better ways to fix this, but the build is green now

PR #649, opened 2019-06-11
Expose a Stream<Entry> out of CompositeMetadata

simonbasle:


simonbasle:
(note: tests fail due to the bug fixed in #651)

PR #650, opened 2019-06-12
ClientServerMultiplexer should properly route zero non-setup frames

mostroverkhov:
to either requester or responder, regardless of connection side. This would simplify lease by removing coupling between requester and responder. Also fixes MetadaPush related issue https://github.com/rsocket/rsocket-java/issues/470

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/internal/ClientServerInputMultiplexer.java
+++ b/rsocket-core/src/main/java/io/rsocket/internal/ClientServerInputMultiplexer.java
@@ -20,7 +20,6 @@ import io.netty.buffer.ByteBuf;
 import io.rsocket.Closeable;
 import io.rsocket.DuplexConnection;
 import io.rsocket.frame.FrameHeaderFlyweight;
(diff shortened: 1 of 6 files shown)


rsocket-core/src/main/java/io/rsocket/internal/ClientServerInputMultiplexer.java
```diff
@@ -79,10 +81,15 @@ public ClientServerInputMultiplexer(DuplexConnection source, PluginRegistry plug
               int streamId = FrameHeaderFlyweight.streamId(frame);
               final Type type;
               if (streamId == 0) {
-                if (isSetup(frame)) {
+                FrameType frameType = FrameHeaderFlyweight.frameType(frame);
+                if (isSetup(frameType)) {
```
robertroeser: instead of the isSetup and and isRequestFrame could we just those methods, the if/elses and replace it with switch statement? Seems like we don't have to the extra if/else here if there was one switch statement

PR #658, opened 2019-07-05
Tuple Frames

robertroeser:
Replaces composite buffers with specialized tuple frames. Fixes earlier issue where the buffers didn't properly deal with primitives that span multiple buffers.


OlegDokuka:
Does it fix that weird issue with websocket?

robertroeser:
Yep

PR #659, opened 2019-07-06
more performance improvements

robertroeser:
Replaces a UnboundedProcessor's queue with a more efficient queue, and replaces maps with a synchronized map that doesn't need autoboxing.


OlegDokuka:
What is the purpose of `RequestResponseProcessor` since it is used nowhere? (At least I did not find any usage of it)

Issue #603, opened 2019-03-22
CloseableChannel#dispose ignored


pmackowski:
I am trying to shutdown a receiver `RSocketFactory.receive()` through `CloseableChannel#dispose` method but it seems not to be working. Here is a test that I expect to pass:

```
@Test
public void closeReceiver() throws InterruptedException {

int randomPort = ThreadLocalRandom.current().nextInt(10_000, 20_000);

AtomicInteger results = new AtomicInteger();
CountDownLatch errorLatch = new CountDownLatch(1);

CloseableChannel closeableChannel = RSocketFactory.receive()
        .acceptor((setup, sendingSocket) -> {
            return Mono.just(new AbstractRSocket() {

                @Override
                public Flux<Payload> requestStream(Payload payload) {
                    return Flux.interval(Duration.ofMillis(100))
                            .onBackpressureDrop()
                            .map(i -> DefaultPayload.create("Response-" +i));
                }

            });
        })
        .transport(TcpServerTransport.create(randomPort))
        .start()
        .block();

RSocket sender = RSocketFactory.connect()
        .transport(TcpClientTransport.create(randomPort))
        .start()
        .block();

sender.requestStream(DefaultPayload.create("Request"))
        .subscribe(new BaseSubscriber<Payload>() {

            volatile Subscription subscription;

            @Override
            protected void hookOnSubscribe(Subscription subscription) {
                this.subscription = subscription;
                this.subscription.request(1);
                closeableChannel.onClose()
                        .doFinally(signalType-> {
                            // request more, should produce an error as receiver no longer produce
                            subscription.request(Integer.MAX_VALUE);
                        })
                        .subscribe();
            }

            @Override
            protected void hookOnNext(Payload value) {
                if (results.compareAndSet(0, 1)) {
                    // expecting it to stop the server
                    closeableChannel.dispose();
                } else {
                    results.getAndIncrement();
                }
            }

            @Override
            protected void hookOnError(Throwable throwable) {
                if (throwable instanceof ClosedChannelException) {
                    errorLatch.countDown();
                }
            }
        });

assertThat(errorLatch.await(1, TimeUnit.SECONDS))
        .withFailMessage("ClosedChannelException is expected").isTrue();
assertThat(results.get()).isEqualTo(1);

}
```

pmackowski:
[...]

PR #660, opened 2019-07-08
netty transports: close connections on server dispose

mostroverkhov:
should fix https://github.com/rsocket/rsocket-java/issues/603

### DIFF
--- a/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/CloseableChannel.java
+++ b/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/CloseableChannel.java
@@ -16,11 +16,15 @@
 
 package io.rsocket.transport.netty.server;
 
+import io.netty.channel.group.ChannelGroup;
 import io.rsocket.Closeable;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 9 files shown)


robertroeser:
I don't think we should be creating channel groups or change the executors - esp. with reactor-netty which sets up it's own event loops. This could also screw up other netty instances. When you receive a connection you could just subscribe to an mono processor and that when it completes it calls dispose.

smaldini:
I would advise caution on that change - Maybe we should look at the problem in a different fashion, e.g. you need to dispose a server mostly in test scenarios. Should we provide rsocket-test to help with this ?
The workaround is not too hard neither, disposing the resources (and shutting down the event loop group) will force the close in an `@After` callback .

mostroverkhov:
@smaldini In that case I'd suggest removing `dispose` on rsocket server, given that all transports implemented are based on netty

sleoniy:
Still having this issue, it appears PR is still open.

OlegDokuka:
Closing in favor of native graceful shutdown support in Reactor-Netty. See https://github.com/reactor/reactor-netty/releases and https://github.com/reactor/reactor-netty/pull/1034

PR #667, opened 2019-07-19
stream id wrap fix

robertroeser:
wraps the stream id when it reaches 2^31 - it also checks for active streams and increments the id if it finds a collision

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/RSocketRequester.java
+++ b/rsocket-core/src/main/java/io/rsocket/RSocketRequester.java
@@ -205,7 +205,7 @@ class RSocketRequester implements RSocket {
       return Mono.error(err);
     }
 
-    final int streamId = streamIdSupplier.nextStreamId();
+    final int streamId = streamIdSupplier.nextStreamId(receivers);
 
     return emptyUnicastMono()
         .doOnSubscribe(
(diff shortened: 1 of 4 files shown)


rsocket-core/src/main/java/io/rsocket/RSocketRequester.java
```diff
@@ -205,7 +205,7 @@ public boolean isDisposed() {
       return Mono.error(err);
     }
 
-    final int streamId = streamIdSupplier.nextStreamId();
+    final int streamId = streamIdSupplier.nextStreamId(receivers);
```
hepin1989: why not make it a long?

robertroeser: According to the protocol the stream id is an unsigned 32-bit number

yschimke:
0 shouldn’t be the first wrapped stream Id should it?

rsocket-core/src/test/java/io/rsocket/StreamIdSupplierTest.java
```diff
@@ -82,4 +88,32 @@ public void testServerIsValid() {
...
     assertTrue(s.isBeforeOrCurrent(1));
   }
+
+  @Test
+  public void testWrap() {
+    IntObjectMap<Object> map = new SynchronizedIntObjectHashMap<>();
+    StreamIdSupplier s = new StreamIdSupplier(Integer.MAX_VALUE - 3);
+
+    assertEquals(2147483646, s.nextStreamId(map));
+    assertEquals(0, s.nextStreamId(map));
```
yschimke: 2 should be the first result here.

PR #672, opened 2019-07-29
Feature/rate limiter publisher

OlegDokuka:
In this PR I want to introduce rate-limited request coordinator which basically prevents direct Long.MAX_VALUE propagation.

I'm considering moving this part to "addons" and instrument RSocket via `plugin` API.

For now, leave it in this for pre-reviewing

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/internal/RateLimitableRequestPublisher.java
+++ b/rsocket-core/src/main/java/io/rsocket/internal/RateLimitableRequestPublisher.java
@@ -1,0 +1,238 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.rsocket.internal;
+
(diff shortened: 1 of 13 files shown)


yschimke:
I thought this was already directly possible based on optional Spring Reactor operations?  I hate the infinite then cancel default mode of Reactor, but why the need to implement yourself?

OlegDokuka:
@yschimke I consider that as a replacement for LimitableRequestPublisher which does `internaRequesN(Long.MAX_VALUE)` and then do coordination based on the external request size. In case external consumer requests `Long.MAX_VALUE` it can lead to -> https://github.com/rsocket/rsocket-java/pull/638. This still an issue. Previously, I was not able to end up with efficient enough and properly tunned request coordination algorithm, and do not think it makes any sense (see Flux#flatMap request coordination between producers. You may find that it is based on the prefetch and subsequent additional requests over the time). Thus, my idea is to get rid of `LimitableRequestPublisher`, so it will not bring any additional synchronization overhead during runtime, and in case of need, provide Plugin with that functionality. Note, default Reactor#limitRate is too overhead. It creates `SPSCArrayQueue` (in fact there are no reasons to do that) and then introduces tasks rescheduling over `Schedulers.immediate`, so it leads to function invocation overhead

yschimke:
Thanks for the explanation. Is this worth opening an issue with reactor as well?

OlegDokuka:
I think, I'm going to port it eventually, but it worth, indeed

rsocket-core/src/main/java/io/rsocket/internal/RateLimitableRequestPublisher.java
```diff
@@ -0,0 +1,242 @@
...
+
+    source.subscribe(s);
+    destination.onSubscribe(s);
+  }
+
+  @Override
+  public void request(long n) {
+    synchronized (this) {
+      long requested = externalRequested;
+      if (requested == Long.MAX_VALUE) {
```
yschimke: This doesn't appear to match the logic on line 56.  Integer vs Long MAX_VALUE.

Issue #665, opened 2019-07-18
Add resumeToken() in ConnectionSetupPayload class


linux-china:
Please add resumeEnabled() and resumeToken() in ConnectionSetupPayload class
 
```
public abstract class ConnectionSetupPayload {
  public boolean abstract resumeEnabled()
  public ByteBuf abstract resumeToken()
}
```

resumeEnabled() and  resumeToken()  already implemented in  ConnectionSetupPayload class.
   
```
class SetupFrameFlyweight {
 public static boolean resumeEnabled(ByteBuf byteBuf)
  public static ByteBuf resumeToken(ByteBuf byteBuf); 
}
```

PR #677, opened 2019-07-31
ConnectionSetupPayload: expose resume flag & token

mostroverkhov:
Relates to #665

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/ConnectionSetupPayload.java
+++ b/rsocket-core/src/main/java/io/rsocket/ConnectionSetupPayload.java
@@ -17,10 +17,10 @@
 package io.rsocket;
 
 import io.netty.buffer.ByteBuf;
-import io.netty.buffer.Unpooled;
 import io.netty.util.AbstractReferenceCounted;
 import io.rsocket.frame.FrameHeaderFlyweight;
 import io.rsocket.frame.SetupFrameFlyweight;
+import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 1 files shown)


mostroverkhov:
looks like `develop` build (openjdk) is not green - there are failing tests `testThatSetupWithSpecifiedFrameSizeButLowerThanWsDefaultShouldSetToWsDefault`

rsocket-core/src/main/java/io/rsocket/ConnectionSetupPayload.java
```diff
@@ -101,6 +106,18 @@ public boolean willClientHonorLease() {
       return SetupFrameFlyweight.honorLease(setupFrame);
     }
 
+    @Override
+    public boolean isResumeEnabled() {
+      return SetupFrameFlyweight.resumeEnabled(setupFrame);
+    }
+
+    @Override
+    public ByteBuf resumeToken() {
```
yschimke: Should this be null instead?  For things like metadata, null means not present.

mostroverkhov: resume token cant be empty - there is no need to distinguish with null.  Looks like this change can be reduced to one token returning method - and empty would mean non-resumable

mostroverkhov: @yschimke checking spec `Unsigned 16-bit integer of Resume Identification Token Length in bytes. (Not present if R flag is not set)`, having both `isResumeEnabled` and nullable `resumeToken` seems best option.

PR #681, opened 2019-08-07
Implement  Routing Metadata Extension

linux-china:
Routing Metadata Extension: 
 https://github.com/rsocket/rsocket/blob/master/Extensions/Routing.md

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/metadata/TaggingMetadata.java
+++ b/rsocket-core/src/main/java/io/rsocket/metadata/TaggingMetadata.java
@@ -1,9 +1,10 @@
 package io.rsocket.metadata;
 
+import static io.netty.buffer.Unpooled.buffer;
+
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.CompositeByteBuf;
 import io.netty.buffer.Unpooled;
-
 import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


rsocket-core/src/main/java/io/rsocket/metadata/TaggingMetadata.java
```diff
@@ -0,0 +1,88 @@
...
+  private ByteBuf content;
+
+  public TaggingMetadata(String type, ByteBuf content) {
+    this.type = type;
+    this.content = content;
+  }
+
+  public TaggingMetadata(String type, Collection<String> tags) {
+    this.type = type;
+    CompositeByteBuf compositeByteBuf = Unpooled.compositeBuffer();
```
rstoyanchev: `CompositeMetadataFlyweight` allows any `ByteBufAllocator` to be passed in while here it is fixed.

rsocket-core/src/main/java/io/rsocket/metadata/TaggingMetadata.java
```diff
@@ -0,0 +1,88 @@
...
+    tags.stream()
+        .map(tag -> tag.getBytes(StandardCharsets.UTF_8))
+        .filter(bytes -> bytes.length > 0 && bytes.length < TAG_LENGTH_MAX)
+        .forEach(
+            bytes -> {
+              int capacity = bytes.length + 1;
+              ByteBuf tagByteBuf = buffer(capacity, capacity);
```
rstoyanchev: Also no option to use alternative `ByteBufAllocator` here.

rsocket-core/src/main/java/io/rsocket/metadata/TaggingMetadata.java
```diff
@@ -0,0 +1,88 @@
...
+        if (tagLength > 0) {
+          byte[] tagBytes = new byte[tagLength];
+          content.readBytes(tagBytes);
+          return new String(tagBytes, StandardCharsets.UTF_8);
+        } else {
+          return "";
+        }
+      }
+    };
+  }
```
rstoyanchev: A test that proves `iterator()` can be used more than once would be useful. As far as I can see that would require a `ByteBuf` slice for independent iteration.

linux-china:
@robertroeser,  you are correct  and we should use flyweight pattern to implement this feature.  I close this PR and take a look.  If somebody can implement this feature, please implement it first.  I close this PR now and take a look at flyweight pattern.  Thanks for you to review my code.  :)

PR #685, opened 2019-08-20
This PR drafts high-performant low-overhead reconnecting rsocket

OlegDokuka:
This PR introduces High-Performant low-object overhead reconnecting rsocket, which could be useful for many engineering cases.

The work is in progress, but the draft is ready to review.

Plan:

- [x] Working Reconnecting RSocket impl with customized `flatMap` `flatMapMany` implementations
- [ ] More flexible reconnecting strategy (possibly requires bringing reactor-addons with Retry and Backoff (or I will reimplement it counting the fact we need a low object overhead)
- [ ] Built-in retry logic for RSocket methods. As of common usage, `RSocket#<logicalCall>` (e.g. `RSocket#requestResponse.retryBackoff(...)`) should be followed with retry operator on logical streams, so it would make sense to have it built-in as well (so we will get rid of 3 more objects)

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/util/ReconnectingRSocket.java
+++ b/rsocket-core/src/main/java/io/rsocket/util/ReconnectingRSocket.java
@@ -1,0 +1,539 @@
+package io.rsocket.util;
+
+import io.netty.util.internal.ThreadLocalRandom;
+import io.rsocket.Payload;
+import io.rsocket.RSocket;
+import io.rsocket.exceptions.ConnectionCloseException;
+import io.rsocket.exceptions.ConnectionErrorException;
+import java.nio.channels.ClosedChannelException;
+import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscription;
+import reactor.core.CoreSubscriber;
+import reactor.core.Exceptions;
+import reactor.core.Scannable;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.core.publisher.MonoProcessor;
+import reactor.core.publisher.Operators;
+import reactor.core.scheduler.Scheduler;
+import reactor.core.scheduler.Schedulers;
+import reactor.util.annotation.Nullable;
+import reactor.util.context.Context;
+
+@SuppressWarnings("unchecked")
+public class ReconnectingRSocket implements CoreSubscriber<RSocket>, RSocket, Runnable {
+
+  private final long backoffMinInMillis;
+  private final long backoffMaxInMillis;
+  private final Mono<? extends RSocket> source;
+  private final Scheduler scheduler;
+  private final Predicate<Throwable> errorPredicate;
+  private final MonoProcessor<Void> onDispose;
+
+  RSocket value;
+
+  volatile Consumer<RSocket>[] subscribers;
+
+  @SuppressWarnings("rawtypes")
+  static final AtomicReferenceFieldUpdater<ReconnectingRSocket, Consumer[]> SUBSCRIBERS =
+      AtomicReferenceFieldUpdater.newUpdater(
+          ReconnectingRSocket.class, Consumer[].class, "subscribers");
+
+  @SuppressWarnings("rawtypes")
+  static final Consumer<RSocket>[] EMPTY_UNSUBSCRIBED = new Consumer[0];
+
+  @SuppressWarnings("rawtypes")
+  static final Consumer<RSocket>[] EMPTY_SUBSCRIBED = new Consumer[0];
+
+  @SuppressWarnings("rawtypes")
+  static final Consumer<RSocket>[] TERMINATED = new Consumer[0];
+
+  static final ClosedChannelException ON_CLOSE_EXCEPTION = new ClosedChannelException();
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  ReconnectingRSocket(
+      Mono<? extends RSocket> source,
+      Predicate<Throwable> errorPredicate,
+      Scheduler scheduler,
+      long backoffMinInMillis,
+      long backoffMaxInMillis) {
+
+    this.source = source;
+    this.backoffMinInMillis = backoffMinInMillis;
+    this.backoffMaxInMillis = backoffMaxInMillis;
+    this.scheduler = scheduler;
+    this.errorPredicate = errorPredicate;
+    this.onDispose = MonoProcessor.create();
+
+    SUBSCRIBERS.lazySet(this, EMPTY_UNSUBSCRIBED);
+  }
+
+  public void subscribe(Consumer<RSocket> actual) {
+    if (!add(actual)) {
+      actual.accept(value);
+    }
+  }
+
+  @Override
+  public void onSubscribe(Subscription subscription) {
+    subscription.request(Long.MAX_VALUE);
+  }
+
+  @Override
+  public void onComplete() {
+    final RSocket value = this.value;
+
+    if (value == null) {
+      reconnect();
+    } else {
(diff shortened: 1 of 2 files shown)


rsocket-core/src/main/java/io/rsocket/util/ReconnectingRSocket.java
```diff
@@ -0,0 +1,539 @@
...
+                || throwable instanceof ConnectionCloseException
+                || throwable instanceof ConnectionErrorException;
+
+    Duration backoffMin;
+    Duration backoffMax;
+    Supplier<Mono<RSocket>> sourceSupplier;
+    Scheduler scheduler = Schedulers.parallel();
+    Predicate<Throwable> errorPredicate = DEFAULT_ERROR_PREDICATE;
+
+    public Builder withSourceRSocket(Supplier<Mono<RSocket>> sourceSupplier) {
```
yschimke: why not just Mono<RSocket> here?  That could be an instance of Mono.defer called externally.  I always assumed Mono.defer taking Supplier was more about having a SAM and being able to call some sync block that returned e.g. Mono.just(...)

rsocket-core/src/main/java/io/rsocket/util/ReconnectingRSocket.java
```diff
@@ -0,0 +1,539 @@
...
+  }
+
+  @Override
+  public void run() {
+    source.subscribe(this);
+  }
+
+  private void reconnect() {
+    ThreadLocalRandom random = ThreadLocalRandom.current();
+    long nextRandomDelay = random.nextLong(backoffMinInMillis, backoffMaxInMillis);
```
yschimke: Could the backoff strategy be hidden behind a Flux<Long> without losing much?  Or would that negate "high-performant low-overhead"?

yschimke:
Would the intention be to make this the default implementation (used in below code)?  Or is there any likely functional differences?

```
    if (resume) {
      clientRSocketFactory.resume()
        .resumeSessionDuration(ofSeconds(30))
        .resumeStrategy { PeriodicResumeStrategy(ofSeconds(3)) }
    }
```

rsocket-core/src/main/java/io/rsocket/util/ReconnectingRSocket.java
```diff
@@ -0,0 +1,539 @@
...
+import reactor.core.publisher.Mono;
+import reactor.core.publisher.MonoProcessor;
+import reactor.core.publisher.Operators;
+import reactor.core.scheduler.Scheduler;
+import reactor.core.scheduler.Schedulers;
+import reactor.util.annotation.Nullable;
+import reactor.util.context.Context;
+
+@SuppressWarnings("unchecked")
+public class ReconnectingRSocket implements CoreSubscriber<RSocket>, RSocket, Runnable {
```
yschimke: Does Runnable need to be part of the public API? Is this to remove one allocation of an anonymous Runnable instance?

yschimke:
Are you working against some higher level performance/allocation targets?  Are those things that can be made visible.  This seems cool, I can understand in concept why you're doing it, but maybe just not when is it good enough?

simonbasle:
@OlegDokuka putting it here just in case since I noticed the mention of addons with retry/backoff: there is now a `retryBackoff` default implementation right into core `Flux`. Not sure if you wanted to reuse the builders in a standalone way or combined with the vanilla `retryWhen` operator, though...

rstoyanchev:
I can't review, so here are a few initial comments:

1. I like how this encapsulates the reconnect logic which can be quite complex for a shared (multiplexed) connection, given that multiple requests can fail and then compete to re-establish the connection. It also eliminates the need to `flatMap` on a `Mono<RSocket>` since `ReconnectingRSocket` has the built-in ability to connect and reconnect.

1. One downside is less flexibility, e.g. you can't vary reconnect parameters per request/stream, and likewise you are always connected even if not making any requests. This may be possible to address by evolving the design but I haven't thought further on that.

1. The builder offers only a very simple never ending reconnect at the moment. Reactor users will expect what they are familiar with from Flux/Mono `retry`. There is a `Retry` contract being introduced in reactor-core, with specs for max retries or backoff strategy, and that will replace the one from reactor-addons. That will make it easier to integrate with here, see https://github.com/reactor/reactor-core/pull/1979.

1. I haven't proven these but they look like issues in `FlatMapInner` and `FlatMapManyInner`:

    1.  The cached input `Payload` can be leaked if we connection never succeeds and we stop trying, currently only via call to dispose.

    1. If a request fails, e.g. with `ClosedChannelException`, the cached Payload is re-used but is probably released at this point because it's been given to be sent. The window for this to happen on FNF or on RR is small but for Request-Stream it is much wider.

    1. The `onError` seems like should not pass the error downstream if reconnection is in progress, i.e. when `tryResubscribe` returns true.

Overall a promising direction. Personally I'm also interested in parallel to see if we can achieve the same with just Reactor Core operators. I've discussed with @simonbasle a potential `cache(..)` variant which works with a predicate to decide if the cached thing is still valid or needs to be re-obtained. That could make it reasonably straight-forward to put together a reconnect chain by yourself and that would be less convenient but more flexible.

OlegDokuka:
> One downside is less flexibility, e.g. you can't vary reconnect parameters per request/stream, and likewise you are always connected even if not making any requests which may or may not be what you want. There may be a way to evolve the design but I haven't thought further on that.


I assume users will be writing something like 

```java
rsocket.requestStream().retryBackoff(...).subscriber();
rsocket.requestStream().retryWhen(...).subscriber();
```
Every stream can be retried independently. 

If you mean that after the error we immediately try to reconnect using the specified strategy, then, in theory, I can do a lazy reconnection, but does it make any sense? I see the case when the lazy (lazy in this context is a reconnection which happens only when there is a need for that) reconnection happens immediately, so there is no backoff. As a subsequence of that, one can require reconnection with backoff, so it may complicate internals but it is possible.


> The builder offers only a very simple never-ending reconnect at the moment. Reactor users will expect what they are familiar with from Flux/Mono retry. Note there is a Retry contract being introduced in reactor-core (with specs for max retries or backoff strategy) to replace the one in reactor-addons and that should make it easier to use here and make it consistent with Reactor Core options, see reactor/reactor-core#1979.

Correct, this is a draft. Just to make sure it is useful. Once POC is good to go I will provide api identical to Retry from project reactor.

> The cached input Payload can be leaked if we connection never succeeds and we stop trying, currently only via call to dispose.

Indeed. Going to fix that.

> The onError seems like should not pass the error downstream if reconnection is in progress, i.e. when tryResubscribe returns true.

I propagate exception because it is up to business logic whether it is necessary to retire it or not. Imagine flux of elements. In the middle of the stream, we observed an exception. IF we swallow it internally and then reconnect, the user will never know that the stream is not working anymore and itis necessary to call it again

rstoyanchev:
So let me see if I get the logic on request failure in `FlatMap[Many]Inner` right:
```java
if (this.parent.tryResubscribe(rSocket, t)) {
  this.actual = null;
  this.state = NONE;
} else {
  done = true;
}
actual.onError(t);
```

If the error is connection related, as per `DEFAULT_ERROR_PREDICATE`, then start trying to reconnect and reset the state, hence leaving it open for downstream to re-subscribe. For any other error set `done`, hence now any further interaction.

Assuming this is correct:
* Why does the reconnect start immediately if it is up to each downstream to retry?
* What if downstream want to retry for any other error? It seems not possible.

> I assume users will be writing something like
rsocket.requestStream().retryBackoff(...).subscriber();
rsocket.requestStream().retryWhen(...).subscriber();

Okay, so this would be required in any scenario with reconnect then. That increases the need to use consistent retry API. 

> Once POC is good to go I will provide api identical to Retry from project reactor.
 
In the proposal there is a `Retry` contract (functional interface) with `RetrySpec` (basic number based) and `RetryBackOffSpec` implementations. It would be much better if these classes were re-used. It would be quite a bit to duplicate all that and there is the challenge of keeping them in sync over time.

Downstream retry logic would also need to check the type of error with something like:
```java
private static final Predicate<Throwable> DEFAULT_ERROR_PREDICATE =
    throwable ->
        throwable instanceof ClosedChannelException
            || throwable instanceof ConnectionCloseException
            || throwable instanceof ConnectionErrorException;
```

This is also where a `Retry` spec could be provided with the error predicate set for connection errors, so that it can be more convenient to plug in. It might look like this:

```java
rsocket.requestStream().retry(ReconnectRetry.max(5)).subscriber();
```

OlegDokuka:
After the discussion with @rstoyanchev, we ended up to derive common as `ReconnectMono` and put it in the reactor-addons. For more info see ->  https://github.com/reactor/reactor-addons/pull/229

rstoyanchev:
@OlegDokuka this can be closed now?

OlegDokuka:
@rstoyanchev yes

PR #687, opened 2019-08-30
Implement Routing Metadata Extension

linux-china:
Routing Metadata Extension:
https://github.com/rsocket/rsocket/blob/master/Extensions/Routing.md

Changes:  TaggingMetadata, RoutingMetadata and TaggingMetadataFlyweight

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/metadata/TaggingMetadataFlyweight.java
+++ b/rsocket-core/src/main/java/io/rsocket/metadata/TaggingMetadataFlyweight.java
@@ -30,6 +30,16 @@ public class TaggingMetadataFlyweight {
   }
 
   /**
+   * create tagging metadata from composite metadata entry
+   *
+   * @param entry composite metadata entry
+   * @return tagging metadata
+   */
+  public static TaggingMetadata createTaggingMetadata(CompositeMetadata.Entry entry) {
+    return new TaggingMetadata(entry.getMimeType(), entry.getContent());
+  }
+
+  /**
    * create tagging metadata
    *
    * @param allocator the {@link ByteBufAllocator} to use to create intermediate buffers as needed.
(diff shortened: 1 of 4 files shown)


rsocket-core/src/main/java/io/rsocket/metadata/TaggingMetadataFlyweight.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /**
+   * create tagging content
+   *
+   * @param allocator the {@link ByteBufAllocator} to use to create intermediate buffers as needed.
+   * @param tags tag values
+   * @return tagging content
+   */
+  public static ByteBuf createTaggingContent(ByteBufAllocator allocator, Collection<String> tags) {
```
robertroeser: Something like this is about 225k+ ops/sec faster 

```java
CompositeByteBuf taggingContent = allocator.compositeBuffer();
    for (String key : tags) {
      int length = ByteBufUtil.utf8Bytes(key);
      
      if (length > TAG_LENGTH_MAX) {
        continue;
      }

      ByteBuf byteBuf = allocator.buffer().writeByte(length);
      byteBuf.writeCharSequence(key, StandardCharsets.UTF_8);
      taggingContent.addComponent(true, byteBuf);
    }

    return taggingContent;
```

_Original_
TaggingMetadataFlyweightPerf.benchmarkEncode  thrpt   10  678914.817 ± 3637.060  ops/s

_Suggestion_
TaggingMetadataFlyweightPerf.benchmarkEncode  thrpt   10  903925.613 ± 13036.852  ops/s

linux-china: Thanks.   it's library and performance is critical.

PR #691, opened 2019-09-02
provides draft of the enhanced websocket handling

OlegDokuka:
This PR provides some enhancement in WebSocket client/server API which allows manipulating connection handling + receiving a human-readable reason for closed WebSocket. 

For more info see forum discussion -> https://community.netifi.com/t/accessing-http-connection-headers-and-path/180

Also related (https://github.com/rsocket/rsocket-js/issues/42)

### DIFF
--- a/rsocket-examples/src/main/java/io/rsocket/examples/transport/ws/WebSocketHeadersSample.java
+++ b/rsocket-examples/src/main/java/io/rsocket/examples/transport/ws/WebSocketHeadersSample.java
@@ -16,27 +16,25 @@
 
 package io.rsocket.examples.transport.ws;
 
-import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.websocketx.WebSocketCloseStatus;
 import io.rsocket.AbstractRSocket;
 import io.rsocket.ConnectionSetupPayload;
-import io.rsocket.DuplexConnection;
 import io.rsocket.Payload;
 import io.rsocket.RSocket;
 import io.rsocket.RSocketFactory;
 import io.rsocket.SocketAcceptor;
 import io.rsocket.frame.decoder.PayloadDecoder;
-import io.rsocket.transport.ServerTransport;
-import io.rsocket.transport.netty.WebsocketDuplexConnection;
 import io.rsocket.transport.netty.client.WebsocketClientTransport;
+import io.rsocket.transport.netty.server.CloseableChannel;
+import io.rsocket.transport.netty.server.WebsocketServerTransport;
 import io.rsocket.util.ByteBufPayload;
 import java.time.Duration;
 import java.util.HashMap;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
+import reactor.core.publisher.MonoProcessor;
 import reactor.core.scheduler.Schedulers;
-import reactor.netty.Connection;
-import reactor.netty.DisposableServer;
 import reactor.netty.http.server.HttpServer;
 
 public class WebSocketHeadersSample {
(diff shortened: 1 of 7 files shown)


rsocket-examples/src/main/java/io/rsocket/examples/transport/ws/WebSocketHeadersSample.java
```diff
@@ -16,71 +16,69 @@
...
-                              HttpResponseStatus.UNAUTHORIZED.reasonPhrase());
-                        }))
-            .bindNow();
+            .transport(
+                WebsocketServerTransport.builder()
+                    .filteringInbound(
+                        headers -> headers.containsValue("Authorization", "test", true))
+                    .closingWithStatus(headers -> new WebSocketCloseStatus(4404, "Unauthorized"))
+                    .build(HttpServer.create().host("localhost").port(8080))
+                // Same could be done with routing transport
```
robertroeser: Is the comment needed?

OlegDokuka: fixed

rsocket-examples/src/main/java/io/rsocket/examples/transport/ws/WebSocketRouteTransportHeadersSample.java
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            @Override
+            public Mono<Payload> requestResponse(Payload payload) {
+              return Mono.just(payload);
+            }
+
+            @Override
+            public Flux<Payload> requestChannel(Publisher<Payload> payloads) {
+              return Flux.from(payloads).subscribeOn(Schedulers.single());
```
yschimke: Is the subscribe on required here?

OlegDokuka: Actually it is not necessarily needed, thus just removed it at all

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketRouteTransport.java
```diff
@@ -236,4 +308,47 @@ private Matcher matcher(String uri) {
...
+      Objects.requireNonNull(headersPredicate, "Header predicate must not be null");
+      this.headersPredicate = headersPredicate;
+      return this;
+    }
+
+    public Builder closingWithStatus(
+        Function<HttpHeaders, WebSocketCloseStatus> webSocketCloseStatusSupplier) {
+      this.webSocketCloseStatusSupplier =
+          Objects.requireNonNull(
+              webSocketCloseStatusSupplier, "WebSocketCloseStatusSupplier must not be null");
```
yschimke:
```suggestion
              webSocketCloseStatusSupplier, "webSocketCloseStatusSupplier must not be null");
```

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketServerTransport.java
```diff
@@ -147,27 +161,92 @@ public void setTransportHeaders(Supplier<Map<String, String>> transportHeaders)
...
-                      });
-                })
-            .bind()
-            .map(CloseableChannel::new);
+
+    if (isError != null) {
+      return isError;
+    }
+
+    if (headersPredicate != null) {
```
yschimke: Couldn't this be simplified by having a default headers predicate that always returns true?

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketServerTransport.java
```diff
@@ -147,27 +161,92 @@ public void setTransportHeaders(Supplier<Map<String, String>> transportHeaders)
...
+        .map(CloseableChannel::new);
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static class Builder {
+
+    private Predicate<HttpHeaders> headersPredicate;
```
yschimke: Why split headersPredicate and webSocketCloseStatusSupplier?  Seems like this could be combined into a single function call instead?

OlegDokuka: Do you think that returning `null` in case everything is fine is a better approach? I was thinking about API which allows a user to filter incoming connection out and in case it is `rejected` to provide a specific close status. One of the options is to use SynchronousSink (as it is in Reactor) but it seems way complex as of normal developer standpoint (e.g. calling error or success)

yschimke: In this case splitting it means possibly evaluating the rules twice, so probably.  Or use optional?

OlegDokuka:
close in favor of #835

PR #704, opened 2019-10-13
Extra method for checking ByteBuf is a composite metadata

OlegDokuka:
This PR provides an additional utility method which allows checking that given bytebuf correspond to the structure of the composite metadata

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/metadata/CompositeMetadataFlyweight.java
+++ b/rsocket-core/src/main/java/io/rsocket/metadata/CompositeMetadataFlyweight.java
@@ -306,7 +306,7 @@ public class CompositeMetadataFlyweight {
           int mimeLength = Byte.toUnsignedInt(mimeIdOrLength) + 1;
 
           if (compositeMetadata.isReadable(
-                  mimeLength)) { // need to be able to read an extra mimeLength bytes
+              mimeLength)) { // need to be able to read an extra mimeLength bytes
             // here we need a way for the returned ByteBuf to differentiate between a
             // 1-byte length mime type and a 1 byte encoded mime id, preferably without
             // re-applying the byte mask. The easiest way is to include the initial byte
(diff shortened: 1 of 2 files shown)


rsocket-core/src/main/java/io/rsocket/metadata/CompositeMetadataFlyweight.java
```diff
@@ -291,6 +291,55 @@ public static boolean isWellKnownMimeType(ByteBuf header) {
     return header.readableBytes() == 1;
   }
 
+  public static boolean isCompositeMetadata(ByteBuf compositeMetadata) {
+    compositeMetadata.markReaderIndex();
+    compositeMetadata.readerIndex(0);
+
+    int ridx = 0;
+    while (hasEntry(compositeMetadata, ridx)) {
+      if (compositeMetadata.isReadable()) {
```
yschimke: isn't this provably true because of hasEntry above?

yschimke:
What's the motivation for the PR?  Are you hitting a case where you happen to have a ByteBuf and want to guess if it's composite metadata?  Shouldn't the mimetype being `message/x.rsocket.composite-metadata.v0` be definitive?

rsocket-core/src/main/java/io/rsocket/metadata/CompositeMetadataFlyweight.java
```diff
@@ -291,6 +291,55 @@ public static boolean isWellKnownMimeType(ByteBuf header) {
...
+          }
+        } else {
+          compositeMetadata.resetReaderIndex();
+          return false;
+        }
+      }
+      ridx = compositeMetadata.readerIndex();
+    }
+
+    compositeMetadata.resetReaderIndex();
```
yschimke: would this resetReaderIndex be cleaner in a finally block around the entire method?

OlegDokuka:
> What's the motivation for the PR? Are you hitting a case where you happen to have a ByteBuf and want to guess if it's composite metadata? Shouldn't the mimetype being message/x.rsocket.composite-metadata.v0 be definitive?
> 
> Or is this a validate method?

Not a 100 percent sure that should be here but the idea was to use that method at rsocket-rpc to have Backward compatibility (e.g check whether give bb is composite) but I totally agree that it should work only when setupPayload specified that interaction metadata should be a composite metadata

OlegDokuka:
alright. I'm going to move that to RPC since that is done for RPC only

PR #707, opened 2019-10-17
fix websockets ping/pong control frames handling

### LINKED ISSUE #700: Unhandled pong from IE11 causes connection to die
Issue: IE11 sends a pong every 30 seconds which is unhandled and eventually the application tries to decode the pong frame and fails. This kills the connection from ie11 to the server. 

Simple demo app to reproduce the issue:
https://github.com/jeisea/violet-demo
Node web app. 

Spring boot rsocket server in the zip.

mostroverkhov:
closes #700

### DIFF
--- a/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketServerTransport.java
+++ b/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketServerTransport.java
@@ -19,12 +19,6 @@ package io.rsocket.transport.netty.server;
 import static io.rsocket.frame.FrameLengthFlyweight.FRAME_LENGTH_MASK;
 
 import io.netty.buffer.ByteBufAllocator;
-import io.netty.buffer.Unpooled;
-import io.netty.channel.ChannelHandlerContext;
-import io.netty.channel.ChannelInboundHandlerAdapter;
-import io.netty.handler.codec.http.websocketx.CloseWebSocketFrame;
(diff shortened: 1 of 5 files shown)


rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketServerTransport.java
```diff
@@ -107,33 +101,7 @@ public static WebsocketServerTransport create(InetSocketAddress address) {
...
-                tcpServer.doOnConnection(
-                    connection ->
-                        connection.addHandlerLast(
-                            new ChannelInboundHandlerAdapter() {
-                              @Override
-                              public void channelRead(ChannelHandlerContext ctx, Object msg)
-                                  throws Exception {
-                                if (msg instanceof PongWebSocketFrame) {
-                                  logger.debug("received WebSocket Pong Frame");
-                                } else if (msg instanceof PingWebSocketFrame) {
```
yschimke: What is handling the incoming Ping frames after this PR?  Surely we still need to respond?

mostroverkhov: pings are handled by reactor-netty https://github.com/reactor/reactor-netty/blob/cbd4c588abfb852931cb81d5a605f9ab5c844595/src/main/java/reactor/netty/http/server/WebsocketServerOperations.java#L147

PR #709, opened 2019-10-29
Revert "Performance Optimization Session (#682)"

mostroverkhov:


OlegDokuka:
Hey.

What is the reason for doing that?

mostroverkhov:
The reason is It makes soak test unstable, probably due to impl does not respect downstream demand request, and improper use of `lazySet`

OlegDokuka:
Let me take a look at that, may be I will be able to provide a fix quickly

OlegDokuka:
Alright. I see the problem. Totally agree - my bad. Going to fix that later cuz it  is going to take a while

PR #715, opened 2019-11-11
Remove rsocket-examples from generated BOM.

### LINKED ISSUE #714: rsocket-bom includes unreleased artifacts
The BOM published by this project (https://search.maven.org/artifact/io.rsocket/rsocket-bom/1.0.0-RC5/pom) includes unpublished artifacts:

```
      <dependency>
        <groupId>io.rsocket</groupId>
        <artifactId>rsocket-examples</artifactId>
        <version>1.0.0-RC5</version>
      </dependency>
```

I think this should be removed. This is currently a problem for us because we apply everything that is there in the Spring BOM to a dependency set in order to scan for security vulnerabilities.

franzbecker:
Fixes #714

### DIFF
--- a/rsocket-bom/build.gradle
+++ b/rsocket-bom/build.gradle
@@ -24,7 +24,7 @@ description = 'RSocket Java Bill of materials.'
 
 dependencies {
     constraints {
-        parent.subprojects.findAll { it.name != project.name }.sort { "$it.name" }.each {
+        parent.subprojects.findAll { it.name != project.name && it.name != "rsocket-examples" }.sort { "$it.name" }.each {
             api it
         }
     }


rsocket-bom/build.gradle
```diff
@@ -24,7 +24,7 @@ description = 'RSocket Java Bill of materials.'
 
 dependencies {
     constraints {
-        parent.subprojects.findAll { it.name != project.name }.sort { "$it.name" }.each {
+        parent.subprojects.findAll { it.name != project.name && it.name != "rsocket-examples" }.sort { "$it.name" }.each {
```
franzbecker: Alternatively you could create a more generic logic that goes something like this:
```
        parent.subprojects.findAll { it.name != project.name }.sort { "$it.name" }.each { subproject ->
            subproject.plugins.withType(MavenPublishPlugin.class) {
                api subproject
            }
        }
```

OlegDokuka:
Looks good. Thanks for contribution

franzbecker:
@OlegDokuka would you mind releasing that change as another RC or could you tell me when you plan to release the next one? Thanks!

PR #727, opened 2019-12-07
fixes error handling to match specification

OlegDokuka:
This PR extends the current implementation of error handling to match what is defined in the spec and let the user use a range of user allowed error codes using `CastomRSocketException` class

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/exceptions/CustomRSocketException.java
+++ b/rsocket-core/src/main/java/io/rsocket/exceptions/CustomRSocketException.java
@@ -1,0 +1,51 @@
+package io.rsocket.exceptions;
+
+import io.rsocket.frame.ErrorType;
+
+public class CustomRSocketException extends RSocketException {
+  private static final long serialVersionUID = 7873267740343446585L;
+
+  private final int errorCode;
+
+  /**
+   * Constructs a new exception with the specified message.
+   *
+   * @param errorCode customizable error code. Should be in range [0x00000301-0xFFFFFFFE]
+   * @param message the message
+   * @throws NullPointerException if {@code message} is {@code null}
+   * @throws IllegalArgumentException if {@code errorCode} is out of allowed range
+   */
+  public CustomRSocketException(int errorCode, String message) {
+    super(message);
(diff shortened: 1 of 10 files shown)


OlegDokuka:
@gregwhitaker ☝️

rsocket-core/src/main/java/io/rsocket/exceptions/Exceptions.java
```diff
@@ -34,36 +34,50 @@ private Exceptions() {}
...
+          return new ApplicationErrorException(message);
+        case REJECTED:
+          return new RejectedException(message);
+        case CANCELED:
+          return new CanceledException(message);
+        case INVALID:
+          return new InvalidException(message);
+        default:
+          if (errorCode >= MIN_USER_ALLOWED_ERROR_CODE
+              || errorCode <= MAX_USER_ALLOWED_ERROR_CODE) {
```
yschimke: On re-reading on a computer (not Github iOS tesatflight app :), this seems good.

```
          if (errorCode >= 769
              || errorCode <= -2) {
```

rsocket-core/src/main/java/io/rsocket/exceptions/CustomRSocketException.java
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @param errorCode customizable error code. Should be in range [0x00000301-0xFFFFFFFE]
+   * @param message the message
+   * @throws NullPointerException if {@code message} is {@code null}
+   * @throws IllegalArgumentException if {@code errorCode} is out of allowed range
+   */
+  public CustomRSocketException(int errorCode, String message) {
+    super(message);
+    if (errorCode > ErrorType.MAX_USER_ALLOWED_ERROR_CODE
+        && errorCode < ErrorType.MIN_USER_ALLOWED_ERROR_CODE) {
```
yschimke: Looks legit, sorry.

```
    if (errorCode > -2
        && errorCode < 769) {
```

OlegDokuka:
So, yeah, to clarify, since error API expect int and return int, we cannot use long for that purpose, that is why had to deal with such unobvious comparison

PR #731, opened 2019-12-18
Authentication Metadata Extension support 

OlegDokuka:
This PR provides the implementation of the Authentication CompositeMetadata. For more information see - > https://github.com/rsocket/rsocket/blob/master/Extensions/Security/Authentication.md


rsocket-core/src/main/java/io/rsocket/buffer/Tuple3ByteBuf.java
```diff
@@ -235,15 +235,13 @@ public ByteBuf getBytes(int index, ByteBuf dst, int dstIndex, int length) {
...
-    return getBytes(0, dstBuf, index, min);
+    return getBytes(index, dstBuf, dstIndex, length);
   }
 
   @Override
   public ByteBuf getBytes(int index, ByteBuffer dst) {
     ByteBuf dstBuf = Unpooled.wrappedBuffer(dst);
-    int min = Math.min(dst.limit(), capacity);
-    return getBytes(0, dstBuf, index, min);
+    return getBytes(index, dstBuf);
```
yschimke: Isn't this now recursive?

rsocket-core/src/main/java/io/rsocket/metadata/security/AuthMetadataFlyweight.java
```diff
@@ -0,0 +1,338 @@
...
+  static final int AUTH_TYPE_ID_LENGTH = 1;
+
+  static final char[] EMPTY_CHARS_ARRAY = new char[0];
+
+  private AuthMetadataFlyweight() {}
+
+  /**
+   * Encode a Authentication CompositeMetadata payload using custom authentication type
+   *
+   * @throws IllegalArgumentException if customAuthType is non US_ASCII string if customAuthType is
```
yschimke: nit, the sentence is awkward with ifs.

rsocket-core/src/main/java/io/rsocket/metadata/security/AuthMetadataFlyweight.java
```diff
@@ -0,0 +1,338 @@
...
+   */
+  public static ByteBuf encodeMetadata(
+      ByteBufAllocator allocator, String customAuthType, ByteBuf metadata) {
+
+    int actualASCIILength = ByteBufUtil.utf8Bytes(customAuthType);
+    if (actualASCIILength != customAuthType.length()) {
+      metadata.release();
+      throw new IllegalArgumentException("custom auth type must be US_ASCII characters only");
+    }
+    if (actualASCIILength < 1 || actualASCIILength > 128) {
```
yschimke: should this be 127?

rsocket-core/src/main/java/io/rsocket/metadata/security/AuthMetadataFlyweight.java
```diff
@@ -0,0 +1,338 @@
...
+      metadata.release();
+      throw new IllegalArgumentException(
+          "custom auth type must have a strictly positive length that fits on 7 unsigned bits, ie 1-128");
+    }
+
+    int capacity = 1 + actualASCIILength;
+    ByteBuf headerBuffer = allocator.buffer(capacity, capacity);
+    // encoded length is one less than actual length, since 0 is never a valid length, which gives
```
yschimke: Is this worthwhile doing?

rsocket-core/src/main/java/io/rsocket/metadata/security/AuthMetadataFlyweight.java
```diff
@@ -0,0 +1,338 @@
...
+   * @param allocator the {@link ByteBufAllocator} to use to create intermediate buffers as needed.
+   * @param authType the custom mime type to encode.
+   * @param metadata the metadata value to encode.
+   */
+  public static ByteBuf encodeMetadata(
+      ByteBufAllocator allocator, WellKnownAuthType authType, ByteBuf metadata) {
+
+    if (authType == WellKnownAuthType.UNPARSEABLE_AUTH_TYPE
+        || authType == WellKnownAuthType.UNKNOWN_RESERVED_AUTH_TYPE) {
+      metadata.release();
```
yschimke: Probably easiest to leave the release here, but it seems a little bit weird for debugging etc, and when catching an exception you may want to try to recover.

rsocket-core/src/main/java/io/rsocket/metadata/security/AuthMetadataFlyweight.java
```diff
@@ -0,0 +1,338 @@
...
+   * @throws IllegalArgumentException if username length is greater than 256
+   * @param allocator the {@link ByteBufAllocator} to use to create intermediate buffers as needed.
+   * @param username the char sequence which represents user name.
+   * @param password the char sequence which represents user password.
+   */
+  public static ByteBuf encodeSimpleMetadata(
+      ByteBufAllocator allocator, char[] username, char[] password) {
+
+    int usernameLength = CharByteBufUtil.utf8Bytes(username);
+    if (usernameLength > 256) {
```
yschimke: Isn't 256 an overflow in the writeByte below?

rsocket-core/src/main/java/io/rsocket/metadata/security/WellKnownAuthType.java
```diff
@@ -0,0 +1,121 @@
...
+
+  static final WellKnownAuthType[] TYPES_BY_AUTH_ID;
+  static final Map<String, WellKnownAuthType> TYPES_BY_AUTH_STRING;
+
+  static {
+    // precompute an array of all valid auth ids, filling the blanks with the RESERVED enum
+    TYPES_BY_AUTH_ID = new WellKnownAuthType[128]; // 0-127 inclusive
+    Arrays.fill(TYPES_BY_AUTH_ID, UNKNOWN_RESERVED_AUTH_TYPE);
+    // also prepare a Map of the types by auth string
+    TYPES_BY_AUTH_STRING = new HashMap<>(128);
```
yschimke: TIOLI: favour LinkedHashMap for predictability and debuggability.

rsocket-core/src/main/java/io/rsocket/metadata/security/AuthMetadataFlyweight.java
OlegDokuka: Kind of pattern used in composite metadata impl. Since spec mentions that max length is 128 then it is the only way to make it. If we are not going to encode it in such a way it will not be possible to distinguish size from encoded wellknownauth type. (0-127 for auth types masked with 0x80 `1000 0000` which means that value overflows) so if one is going to use 128 length for encoded custom auth type then it will be identified as -1&0x7F == 0 well-known mime type

rsocket-core/src/main/java/io/rsocket/metadata/security/AuthMetadataFlyweight.java
OlegDokuka: fixed

PR #736, opened 2020-01-24
[Bugfix] Incorrect Behaviour of RequestChannel 

OlegDokuka:
This PR provides a fix for the incorrect behavior of RequestChannel.

At this moment there is a couple of issues with that:

1.1) There was an incorrect expectation of signal ordering which is `requestN` from the local downstream, onNext from the local upstream. In fact, signal order can be totally different and while we wait for first onNext signal, nothing prevents to perform a few more `requestN` which is going to break correctness (see https://github.com/rsocket/rsocket-java/pull/736/files#diff-a87c12662c200847a65e4b44b57e2278R236)

1.2) There was an incorrect behavior in case upstream terminates earlier than the remote responder so the downstream observed onComplete before it should do so. The RSocket-Spec says that the stream can be half-closed until both sides send the complete signal. (https://github.com/rsocket/rsocket-java/pull/736/files#diff-b0fea90eb4e3536e65de608e97f637f9L369)

2) I found that there could be racing between sending first frame and cancelation, so I added an additional flag that ensures that we released the given frame and no leaks going to appear. (see https://github.com/rsocket/rsocket-java/pull/736/files#diff-b0fea90eb4e3536e65de608e97f637f9R394)

3) In order to implement correct behavior of request channel I reimplemented the switchOnFirst operator to follow the behavior defined by RSocket-Spec (https://github.com/rsocket/rsocket-java/pull/736/files#diff-ec51e7f167b7b31f5b3c4b25f5233e8dR1)

4) This PR is a temporary fix on the upcoming rewriting of every stream to state-machine it is done in the RSocket-CPP https://github.com/rsocket/rsocket-cpp/tree/master/rsocket/statemachine


yschimke:
Did you miss a file? or is this a draft?

OlegDokuka:
@yschimke pushed fix

rsocket-core/src/main/java/io/rsocket/RSocketResponder.java
```diff
@@ -488,7 +458,21 @@ private void handleChannel(int streamId, Payload payload, int initialRequestN) {
...
-                l -> sendProcessor.onNext(RequestNFrameFlyweight.encode(allocator, streamId, l)))
+                new LongConsumer() {
+                  boolean first = true;
+
+                  @Override
+                  public void accept(long l) {
+                    long n;
+                    if (first) {
+                      first = false;
+                      n = l - 1;
```
simonbasle: just for the sake of readability (with a non-coding font), consider using either a different letter than `l` or the literal `1L` instead of `1`

rsocket-core/src/main/java/io/rsocket/internal/FluxSwitchOnFirst.java
```diff
@@ -0,0 +1,657 @@
...
+import reactor.core.publisher.Signal;
+import reactor.util.annotation.Nullable;
+import reactor.util.context.Context;
+
+/**
+ * @author Oleh Dokuka
+ * @param <T>
+ * @param <R>
+ */
+public final class FluxSwitchOnFirst<T, R> extends FluxOperator<T, R> {
```
simonbasle: this should at least very explicitly document the differences with the current core implementation and why this was (re)introduced by this PR when there already is an operator in core... and at least link to the PR/issue in core (if it is as relevant as I presume)

OlegDokuka: In general, this implementation exposes the intent I want to mirror in the switchOnFirst in the reactor. However, this code will be sitting there until I'm going to write custom operators for all the communication types. I want to do something similar to what is done in rsocket-cpp -> https://github.com/rsocket/rsocket-cpp/tree/master/rsocket/statemachine

OlegDokuka:
suppressed by #761

rsocket-core/src/main/java/io/rsocket/RSocketRequester.java
```diff
@@ -255,10 +257,12 @@ public void doOnTerminal(
 
     final UnboundedProcessor<ByteBuf> sendProcessor = this.sendProcessor;
     final UnicastProcessor<Payload> receiver = UnicastProcessor.create();
+    final AtomicBoolean payloadReleasedFlag = new AtomicBoolean(false);
 
     receivers.put(streamId, receiver);
 
     return receiver
+        .log()
```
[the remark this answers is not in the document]
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/RSocketResponder.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/internal/FluxSwitchOnFirst.java
OlegDokuka: Illuminated. Using the fixed one from the latest Reactor-Core

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -320,85 +377,47 @@ public void accept(long n) {
...
-                              receiver.dispose();
-                            }
-                          });
-                } else {
-                  if (contains(streamId) && !receiver.isDisposed()) {
-                    sendProcessor.onNext(RequestNFrameFlyweight.encode(allocator, streamId, n));
+                  senders.put(streamId, upstreamSubscriber);
+                  receivers.put(streamId, receiver);
+
+                  inboundFlux.limitRate(Queues.SMALL_BUFFER_SIZE).subscribe(upstreamSubscriber);
```
OlegDokuka, line 383: @rstoyanchev for your attention. I got to read of the custom rate limiter and use a built-in one

rsocket-core/src/main/java/io/rsocket/core/RSocketResponder.java
```diff
@@ -431,55 +398,56 @@ protected void hookOnComplete() {
...
+          }
+
+          @Override
+          protected void hookFinally(SignalType type) {
+            sendingSubscriptions.remove(streamId);
+          }
+        };
+
+    sendingSubscriptions.put(streamId, subscriber);
+    response.limitRate(Queues.SMALL_BUFFER_SIZE).subscribe(subscriber);
```
OlegDokuka, line 450: @rstoyanchev for your attention. I got to rid of the custom rate limiter and use a built-in one

rsocket-test/src/main/java/io/rsocket/test/TransportTest.java
```diff
@@ -215,14 +217,19 @@ default void requestChannel2_000_000() {
...
     getClient()
         .requestChannel(payloads)
-        .as(StepVerifier::create)
+        .as(publisher -> StepVerifier.create(publisher, 3))
         .expectNextCount(3)
         .expectComplete()
         .verify(getTimeout());
+
+    Assertions.assertThat(requested.get())
+        .isEqualTo(257L); // 257 because of eager behavior of limitRate
```
OlegDokuka, line 232: @rstoyanchev for your attention. Here is the result of eager prefetch 1L (from the switchOnFirst + 256 extra)

OlegDokuka: oh... found a bug because of skip does extra request 1 but it should not

OlegDokuka: fixed. prefetch should be exactly 256

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -306,10 +313,67 @@ public void accept(long n) {
...
+          boolean first = true;
+
+          @Override
+          protected void hookOnSubscribe(Subscription subscription) {
+            // noops
+          }
+
+          @Override
+          protected void hookOnNext(Payload payload) {
+            if (first) {
```
OlegDokuka, line 346: @rstoyanchev for your attention. Here how I skip first payload

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -306,10 +313,67 @@ public void accept(long n) {
...
+              // need to skip first since we have already sent it
+              first = false;
+              return;
+            }
```
rstoyanchev, line 350: The `if (first)` check could be replaced with `inboundFlux.skip(1)`?

OlegDokuka: It was with skip initially, but, I figured out that skip does extra request(1) when it skips, so we have extra 1 element requested (I mentioned that in the comments that expected prefetch is 256, but it was 257)

The strategy is the following 

1) we do at the very beginning `request(1)` (from the `switchOnFirst`) to obtain the first element and then do the transformation. 
2) Then, we send that obtained element as a part of `RequestChannel` frame so it appears on the moment of delivery
3) Then, in order to no brake backpressure because of the eager first element prefetch, we reduce 1 from the first upcoming request to the `switchOnFirst`.
4) However, the first prefetched element is sent to the flux along the way, and that would be OK if we have just normal routing case, but, in case of RSocket, the first payload has already been sent so if we are not going to `skip` the first element, it will be delivered twice
5) Here comes the issue, if we use `skip` then `skip` operator think that we need to fulfill the demand by requesting extra one element, but in fact we just need not skip but `DROP` element instead without doing `request(1)`

PR #738, opened 2020-01-29
Add attributes context  for RSocket

linux-china:
RSocket is a persistent connection with long lifecycle, for broker side & load balance implementation to manage lots of RSocket objects, some important attributes, for example connection URL, consumed services, exposed services, last health check timestamp etc,  should be saved  as attributes related to the RSocket for further usage.

The following code from RSocketExchange.java of spring-cloud-rsocket project. 

```
public interface AttributesContext {

  /**
   * Return a mutable map of attributes for the current context.
   *
   * @return current attributes.
   */
  Map<String, Object> getAttributes();

  /**
   * Return the request attribute value if present.
   *
   * @param name the attribute name
   * @param <T> the attribute type
   * @return the attribute value
   */
  @SuppressWarnings("unchecked")
  @Nullable
  default <T> T getAttribute(String name) {
    return (T) getAttributes().get(name);
  }

  default void setAttribute(String name, @Nonnull Object value) {
    getAttributes().put(name, value);
  }

  /**
   * Return the attribute value, or a default, fallback value.
   *
   * @param name the attribute name
   * @param defaultValue a default value to return instead
   * @param <T> the attribute type
   * @return the attribute value
   */
  @SuppressWarnings("unchecked")
  default <T> T getAttributeOrDefault(String name, T defaultValue) {
    return (T) getAttributes().getOrDefault(name, defaultValue);
  }
}
```


rsocket-core/src/main/java/io/rsocket/RSocketRequester.java
```diff
@@ -25,29 +25,17 @@
...
-import io.rsocket.frame.FrameHeaderFlyweight;
-import io.rsocket.frame.FrameType;
-import io.rsocket.frame.MetadataPushFrameFlyweight;
-import io.rsocket.frame.PayloadFrameFlyweight;
-import io.rsocket.frame.RequestChannelFrameFlyweight;
-import io.rsocket.frame.RequestFireAndForgetFrameFlyweight;
-import io.rsocket.frame.RequestNFrameFlyweight;
-import io.rsocket.frame.RequestResponseFrameFlyweight;
-import io.rsocket.frame.RequestStreamFrameFlyweight;
+import io.rsocket.frame.*;
```
spencergibb: Stick to project defaults, no `*` imports.

rsocket-core/src/main/java/io/rsocket/metadata/WellKnownMimeType.java
```diff
@@ -73,6 +73,7 @@
 
   // ... reserved for future use ...
 
+  MESSAGE_RSOCKET_AUTHENTICATION("message/x.rsocket.authentication.v0", (byte) 0x7C),
```
spencergibb: Seems unrelated?

linux-china: yes.  Merged by https://github.com/rsocket/rsocket-java/pull/728

rsocket-core/src/main/java/io/rsocket/RSocketRequester.java
linux-china: fixed.

linux-china:
> Curious why you think this needs to live in rsocket-java?

Now we use RSocketManager to manage all connected RSocket objects, and lots of attributes related with the RSocket  implemented by Extra Maps.  If RSocket has attributes support, and it will be easy to manage the attributes with RSocket.   Hard to sync the attributes when RSocket status changed for Ops requirements.

OlegDokuka:
Personally I'm against modifying the rsocket interface and seeing the same feature could be added apart of it. I will prepare a sample that shows how we can achieve that without introducing new api methods

linux-china:
> Personally I'm against modifying the rsocket interface and seeing the same feature could be added apart of it. I will prepare a sample that shows how we can achieve that without introducing new api methods.
> 
> I think it can leave as a plugin and not the main feature since it does not deliver major value to all RSocket users

Just for discuss.  Most developers like context attributes, such as ServletRequest,  Netty ChannelHandlerContext etc.  With new proxy RSocket introduced, for broker side,  more objects will be created and maintained.   I'm doing some testing for 500K connections for a server.

OlegDokuka:
@linux-china my point is that the same feature could be achieved over composition (in that case patter decorator) rather than via direct extension of the RSocket API.
For example, what could be done is having ContextAwareRsocket which can decorate the default one. The decoration could be done over plugin api in the rsocket factory on both sides (client and server). Such API will prevent modifying of the RSocket interface and gives desired feature.

Contrary to Sevlet / Netty HTTP API, RSocket is a representation of a single connection, and all representations of logical request context can be stored in Reactor Context instead.
Just in case it is necessary achieving some contextual information, let's say, data in setup payload, then additional decorator can resolve that.

In any case, in order to get access to some specific RSocket you would have to maintain 'XXXManager' or some Map, thus it does not mater whether you store RSocket or ContextAwareRsocket.

OlegDokuka:
I'm not saying that context is redundant there. I'm saying that the same result can be achieved in a different way without introducing of new API in the central interface counting the fact the library gives everything needed in order to add that feature seamlessly over plugin API

yschimke:
I'd agree. Better to add via Reactor Context if possible.

linux-china:
Closed, and consider Reactor Context implementation.

PR #739, opened 2020-02-02
provides set of fixes and test for tuple byte buffs

OlegDokuka:
This PR provides tests and a couple of bugfixes for TupleByteBuf

0) Adopted default ByteBuf test suit for TupleByteBuf

1) Removes custom toString implementation in favor of built-in one https://github.com/rsocket/rsocket-java/pull/739/files#diff-639e901074347653a889b0c9c393f25dL364 
2) Fixes readXXX for LittleEndian (https://github.com/rsocket/rsocket-java/pull/739/files#diff-183345d04c26e98d76cdd661694c7e8dL200)
3) Proper endiance switch support (https://github.com/rsocket/rsocket-java/pull/739/files#diff-183345d04c26e98d76cdd661694c7e8dR217) by default it should be BigEndian. On the other hand, I removed override method https://github.com/rsocket/rsocket-java/pull/739/files#diff-183345d04c26e98d76cdd661694c7e8dL232 which causes illegal behavior when netty tries to switch byte ordering in some cases
4) Fixes behaviors of unwrap method  
5) Fixes behavior for setIndex(int, int) writerIndex(int) method (should change index for TupleByteBuf but should not let write data
6) Removes custom implementation of `slice()` method since it created incorrect slice which violates the base test suit

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/buffer/AbstractReadOnlyReferenceCountedByteBuf.java
+++ b/rsocket-core/src/main/java/io/rsocket/buffer/AbstractReadOnlyReferenceCountedByteBuf.java
@@ -1,0 +1,176 @@
+package io.rsocket.buffer;
+
+import io.netty.buffer.AbstractReferenceCountedByteBuf;
+import io.netty.buffer.ByteBuf;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.ReadOnlyBufferException;
+import java.nio.channels.FileChannel;
+import java.nio.channels.ScatteringByteChannel;
+
+abstract class AbstractReadOnlyReferenceCountedByteBuf extends AbstractReferenceCountedByteBuf {
+
+  AbstractReadOnlyReferenceCountedByteBuf(int maxCapacity) {
+    super(maxCapacity);
+  }
+
+  @Override
+  public boolean isReadOnly() {
+    return true;
+  }
+
+  @Override
+  public boolean isWritable() {
+    return false;
+  }
+
+  @Override
+  public boolean isWritable(int numBytes) {
+    return false;
+  }
+
+  @Override
+  public int ensureWritable(int minWritableBytes, boolean force) {
+    return 1;
+  }
+
+  @Override
+  public ByteBuf ensureWritable(int minWritableBytes) {
+    throw new ReadOnlyBufferException();
+  }
+
+  @Override
+  public ByteBuf discardReadBytes() {
+    throw new ReadOnlyBufferException();
+  }
+
+  @Override
+  public ByteBuf setBytes(int index, ByteBuf src, int srcIndex, int length) {
+    throw new ReadOnlyBufferException();
+  }
+
+  @Override
+  public ByteBuf setBytes(int index, byte[] src, int srcIndex, int length) {
+    throw new ReadOnlyBufferException();
+  }
+
+  @Override
+  public ByteBuf setBytes(int index, ByteBuffer src) {
+    throw new ReadOnlyBufferException();
+  }
+
+  @Override
+  public ByteBuf setByte(int index, int value) {
+    throw new ReadOnlyBufferException();
+  }
+
+  @Override
+  protected void _setByte(int index, int value) {
+    throw new ReadOnlyBufferException();
+  }
+
+  @Override
+  public ByteBuf setShort(int index, int value) {
+    throw new ReadOnlyBufferException();
+  }
+
+  @Override
+  protected void _setShort(int index, int value) {
+    throw new ReadOnlyBufferException();
+  }
+
+  @Override
+  public ByteBuf setShortLE(int index, int value) {
+    throw new ReadOnlyBufferException();
+  }
+
+  @Override
+  protected void _setShortLE(int index, int value) {
+    throw new ReadOnlyBufferException();
+  }
+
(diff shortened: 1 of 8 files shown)


rsocket-core/src/main/java/io/rsocket/buffer/AbstractTupleByteBuf.java
```diff
@@ -213,29 +207,19 @@ public int capacity() {
...
-
-  @Override
-  public ByteBuf order(ByteOrder endianness) {
-    return this;
+    return ByteOrder.BIG_ENDIAN;
   }
 
   @Override
   public ByteBuf unwrap() {
-    throw new UnsupportedOperationException();
```
yschimke: mistake?

rsocket-core/src/main/java/io/rsocket/buffer/AbstractReadOnlyReferenceCountedByteBuf.java
```diff
@@ -0,0 +1,176 @@
...
+
+import io.netty.buffer.AbstractReferenceCountedByteBuf;
+import io.netty.buffer.ByteBuf;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.ReadOnlyBufferException;
+import java.nio.channels.FileChannel;
+import java.nio.channels.ScatteringByteChannel;
+
+abstract class AbstractReadOnlyReferenceCountedByteBuf extends AbstractReferenceCountedByteBuf {
```
yschimke: The netty docs deprecated similar classes and suggest using Unpooled.wrappedUnmodifiableBuffer.  Also the docs call out slice as an issue, how is this handled?

https://netty.io/4.1/api/io/netty/buffer/Unpooled.html#wrappedUnmodifiableBuffer-io.netty.buffer.ByteBuf...-

yschimke:
Can you call out the fixes specifically?

rsocket-core/src/main/java/io/rsocket/buffer/AbstractReadOnlyReferenceCountedByteBuf.java
OlegDokuka: It is deprecated but there are no other options to extend. TupledByteBuf designed to be readOnly and I don't wanna create few more objects to runtime just in order to use what netty gives. therefore I mimic the same what readonly bytebuf does

rsocket-core/src/main/java/io/rsocket/buffer/AbstractTupleByteBuf.java
OlegDokuka: Looks like. Please see https://netty.io/4.0/api/io/netty/buffer/ByteBuf.html#unwrap--

> Return the underlying buffer instance if this buffer is a wrapper of another buffer.
> Returns:
> null if this buffer is not a wrapper

OlegDokuka:
> Can you call out the fixes specifically?

Added more explanation in the fix

In general, PR adopts netty bytebuf test-suite which ensures proper implementation of rsocket bytebuf. On the other hand, most of the fixes were discovered as the result of running adopted test-suit against the implementation

/cc  @yschimke @simonbasle

OlegDokuka:
> TupleByteBuf was originally _intended_ to reduce copying, but unfortunately with SSL handler on pipeline It can make performance worse: engine cant append Bytebufs (until fit single TLS record) because `Tuple` is not writable, so allocates new buffer and does copy anyways. Do you think It makes sense to optimise for plaintext over TLS?
> Also AFAIK buggy `TupleByteBuf` implementation was one of the reasons (together with #709) any release after 0.12.2-RC4 (transitively current `spring-boot` and `spring-integration`) is unstable & unusable with TLS enabled transports. SSLHandler is the only one example - custom implementation of huge `ByteBuf` interface may easily have more bugs revealed with other netty handlers.
> I suggest giving users choice between pragmatic and stable `CompositeByteBuf` and a custom one, move `TupleByteBuf` out of core to project - there were enough issues with It.

I'm not sure this comment ever applicable to this PR since it's main intent is to provide a test suite for `tupplebytebuf` rather than to doubt its need.
If you would like to rise that problem and run a discussion over, it is better to discuss that in a separate issue. In general, more than appreciate benchmarks that confirm insignificant performance impact.

In general, I was thinking of making TupleByteBuf writable, counting the fact that right now we have issues with TLS

mostroverkhov:
This is follow up on Yurii request reqarding justification for 5k LOC *bugfix*.
```
 I was thinking of making TupleByteBuf writable
``` 
please do, but with an option for people to select implementation (CompositeByteBuf is pragmatic and stable one), and move `TupleByteBuf` to project which you think will benefit from It :smile:

yschimke:
FWIW I'm providing advisory drive-by reviews, not much code recently. There is a cautious balance between the need to write optimised code to avoid every allocation and keeping the code simple.

I'm naturally inclined towards reuse of existing code from netty and reactor. So take my queries as mainly trying to keep that tension.

OlegDokuka:
As I said before, the justification for this PR is to provide a test-suit.

Justification for `TupleByteBuf` in general is to improve performance, which, for instance, gives around **100k RPS** for simple request-response communication _(that is about 10-15% from the baseline)_.

OlegDokuka:
Needs rethinking

Issue #713, opened 2019-11-06
rsocket-load-balancer first request always ends up with error


mmaszkie:
Hi,
I have a problem with rsocket-load-balancer. I have one instance of client application and 2 instances of server application. They are communicating via RSocket (request/response interaction model) using rsocket-load-balancer on the client side. Unfortunately first request always ends up with an error on the client side.

Scenario:
- I start server applications and then client application,
- during client application startup single instance of LoadBalancedRSocketMono is being created in order to manage RSocket connections between client and server instances; logs:
```
2019-11-06 13:10:40,661 13:10:40.661 [reactor-http-nio-2] DEBUG i.rsocket.client.RSocketSupplierPool - 
Updated active factories (size: 2)
 + RSocketSupplier{address=Address(host=xxx.yyy.zzz, port=31135)}
 + RSocketSupplier{address=Address(host=xxx.yyy.zzz, port=31842)}
Active sockets:
```
- I would expect that the first request from client to server ends up with success, but it's not; there are some NoAvailableRSocketException exceptions in logs:
```
2019-11-06 13:22:21,524 13:22:21.524 [DefaultDispatcher-worker-1] DEBUG i.r.client.LoadBalancedRSocketMono - aperture 0 is below target 10, adding 10 sockets
2019-11-06 13:22:21,524 13:22:21.524 [DefaultDispatcher-worker-1] DEBUG i.r.client.LoadBalancedRSocketMono - addSockets(10) restricted by the number of factories, i.e. addSockets(2)
2019-11-06 13:22:21,525 13:22:21.525 [DefaultDispatcher-worker-1] DEBUG i.rsocket.client.RSocketSupplierPool - Added RSocketSupplier{address=Address(host=xxx.yyy.zzz.www, port=31135)} to leasedSuppliers
2019-11-06 13:22:21,525 13:22:21.525 [DefaultDispatcher-worker-1] DEBUG i.r.client.LoadBalancedRSocketMono - Creating WeightedSocket WeightedSocket(median=0.0 quantile-low=0.0 quantile-high=0.0 inter-arrival=1000000.0 duration/pending=0.0 pending=0 availability= 0.0)-> from factory RSocketSupplier{address=Address(host=xxx.yyy.zzz.www, port=31135)}
[...]

PR #754, opened 2020-03-12
Report no active sockets as an error which enables users of

juri8:
LoadBalancedRSocketMono to correctly react on that.
Issue #713

Due to that change I can now use 
```
LoadBalancedRSocketMono.create(...)
   .retryBackoff(1, Duration.ofMillis(100))

```
which makes the first call also work from a user perspective


OlegDokuka:
@juri8, first of all thank you for the PR. 

Secondly, can you please provide a test which reproduce the case first before fixing the issue.

Regards,
Oleh

OlegDokuka:
closes this one in favor of #899

PR #768, opened 2020-04-05
Payload length validation

OlegDokuka:
This PR provide extra payload validation in order to avoid later exception throwing if frame length exists allowed size for with disabled fragmentation


rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationUtils.java
```diff
@@ -0,0 +1,34 @@
...
+import io.netty.buffer.ByteBuf;
+import io.rsocket.Payload;
+import io.rsocket.frame.FrameHeaderFlyweight;
+import io.rsocket.frame.FrameLengthFlyweight;
+
+public final class FragmentationUtils {
+  public static boolean isValid(int mtu, Payload payload) {
+    return payload.hasMetadata()
+        ? isValid(mtu, payload.data(), payload.metadata())
+        : isValid(mtu, payload.metadata());
```
simonbasle, line 12: /!\ passing `metadata()` to `isValid(mtu, data)`
(a unit test should probably be added)

rsocket-core/src/main/java/io/rsocket/RSocketRequester.java
```diff
@@ -183,6 +187,11 @@ public boolean isDisposed() {
       return Mono.error(err);
     }
 
+    if (!FragmentationUtils.isValid(this.mtu, payload)) {
+      payload.release();
+      return Mono.error(new IllegalArgumentException("Too big Payload size"));
```
simonbasle, line 192: I would tend to reverse the sentence in the error message: `Payload size too big`. Maybe also allude to where the limitation comes from (not necessarily giving the value of the `mtu`, more like how it is configured on the factory, ie `Payload size too big compared to the configured fragment size`, or something like that, since `mtu` is set by the `fragment(int)` method).

(this comment applies to all the instances of the message below as well)

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationUtils.java
OlegDokuka: Yeah. It is a bag. I was sure to test is covering that. Will improve

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationUtils.java
```diff
@@ -0,0 +1,34 @@
+package io.rsocket.fragmentation;
+
+import io.netty.buffer.ByteBuf;
+import io.rsocket.Payload;
+import io.rsocket.frame.FrameHeaderFlyweight;
+import io.rsocket.frame.FrameLengthFlyweight;
+
+public final class FragmentationUtils {
```
rstoyanchev, line 8: This class is for an internal check. It would be better to avoid a whole new public class. An alternative is to create a `BiPredicate<int, Payload> payloadLengthPredicate` in `RSocketFactory` and pass it into requester and responder instead of `int mtu`.

rsocket-core/src/main/java/io/rsocket/RSocketRequester.java
rstoyanchev: Agreed, perhaps something like:
> The payload is too big to send as a single frame with a 24-bit encoded length. Considering enabling fragmentation via RSocketFactory.

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationUtils.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/RSocketRequester.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationUtils.java
OlegDokuka: Instead, moved validator utils to core package and made it pkg private

OlegDokuka:
Deferred until #736

PR #774, opened 2020-04-09
reduce package cycles

rstoyanchev:
This resolves package cycles between:
- `lease` - `internal` due to `ClientSetup` and `ServerSetup` (both related to leasing).
- `exception` - `lease` due to references between `MissingLeaseException` and `Lease`.
- `exception` - `frame` due to references between `ErrorFrameFlyweight` and `RSocketException`.

This is mostly backwards compatible and functionality neutral except for moving `MissingLeasingException` to `lease`. It's unlikely to break anything and a worthwhile trade-off.


rsocket-core/src/main/java/io/rsocket/frame/ErrorFrameFlyweight.java
```diff
@@ -68,4 +76,16 @@ public static ByteBuf data(ByteBuf byteBuf) {
   public static String dataUtf8(ByteBuf byteBuf) {
     return data(byteBuf).toString(StandardCharsets.UTF_8);
   }
+
+  public interface RSocketErrorCode {
```
OlegDokuka, line 80: Lets create a class for toplevel RSocketException

rsocket-core/src/main/java/io/rsocket/core/ServerSetup.java
```diff
@@ -14,40 +14,53 @@
...
 import io.rsocket.keepalive.KeepAliveHandler;
 import io.rsocket.resume.*;
-import io.rsocket.util.ConnectionUtils;
 import java.time.Duration;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import reactor.core.publisher.Mono;
 
-public interface ServerSetup {
+abstract class ServerSetup {
```
OlegDokuka, line 36: Just curious, do we really need that class?

OlegDokuka: Alright, maybe we have to get rid of that one later

rstoyanchev: I think this can be refactored a further, but after an initial attempt I backed out because it turned out a little more involved. I wanted to keep this PR as minimal as possible. As the class is package private, it can be an internal refactoring separately.

rsocket-core/src/main/java/io/rsocket/frame/ErrorFrameFlyweight.java
rstoyanchev: Done.

rsocket-core/src/main/java/io/rsocket/exceptions/RSocketException.java
```diff
@@ -16,41 +16,47 @@
...
+   * Constructs a new exception with the specified message and cause and error code 0x201
+   * (Application error).
    *
    * @param message the message
    * @param cause the cause of this exception
-   * @throws NullPointerException if {@code message} is {@code null}
    */
   public RSocketException(String message, @Nullable Throwable cause) {
-    super(Objects.requireNonNull(message, "message must not be null"), cause);
+    super(0x00000201, message, cause);
```
OlegDokuka, line 49: can we use `ErrorType.APPLICATION_ERROR` here. In any case, we have ref to `ErrorType`
```suggestion
    super(ErrorType.APPLICATION_ERROR, message, cause);
```

rsocket-core/src/main/java/io/rsocket/exceptions/RSocketException.java
```diff
@@ -16,41 +16,47 @@
 
 package io.rsocket.exceptions;
 
-import java.util.Objects;
+import io.rsocket.RSocketErrorException;
 import reactor.util.annotation.Nullable;
```
OlegDokuka, line 20:
```suggestion
import io.rsocket.frame.ErrorType;
import reactor.util.annotation.Nullable;
```

rsocket-core/src/main/java/io/rsocket/exceptions/RSocketException.java
rstoyanchev: Yes, indeed, this should be a constant. I've suggested a further change to use the constants from `ErrorFrameFlyweight`.

PR #775, opened 2020-04-09
[Bugfix] Enforces reassembly to always be enabled

OlegDokuka:
As of now, we have a bug in RSocket, related to the fact that one side may not directly setup fragmentation but another may.
Such miss-setup leads to the fact that one side is not able to reassemble back fragmented payload.

This PR provide fragment reassembling enforcement so every duplex connection is wrapped by a separate `ReassemblyDuplexConnection` which takes care of reassembling back fragments even if fragmentation is not enabled


rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
```diff
@@ -138,23 +138,9 @@ private ByteBuf encode(ByteBuf frame) {
...
   @Override
   public Flux<ByteBuf> receive() {
-    return delegate
-        .receive()
-        .handle(
-            (byteBuf, sink) -> {
-              ByteBuf decode = decode(byteBuf);
-              frameReassembler.reassembleFrame(decode, sink);
-            });
+    return delegate.receive();
```
OlegDokuka, line 143: @rstoyanchev here we remove reassembling logic since we always have it enabled

rsocket-examples/src/main/java/io/rsocket/examples/transport/ws/WebSocketHeadersSample.java
```diff
@@ -61,7 +63,10 @@ public static void main(String[] args) {
                         (in, out) -> {
                           if (in.headers().containsValue("Authorization", "test", true)) {
                             DuplexConnection connection =
-                                new WebsocketDuplexConnection((Connection) in);
+                                new ReassemblyDuplexConnection(
```
OlegDokuka, line 66: Yeah... we always have to wrap every DuplexConnection like that to ensure correctness

OlegDokuka:
@rstoyanchev, in short, I moved reassembly logic from FragmentationDuplexConnection so, for now, Fragmentation does exactly fragmentation and now we have ReassemblyDuplexConnection which accompanies every other DuplexConnection in the know transports. 

Unfortunately, I'm not seeing simple way (without braking API) to built-in reassembly logic directly into the DuplexConnection interface.

Any suggestions are welcome

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
rstoyanchev: This looks good although `ReassemblyDuplexConnection` could move up as  a base class of `FragmentationDuplexConnection`. That way you can have either re-assembly only or re-assembly and fragmentation in one.

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/client/TcpClientTransport.java
```diff
@@ -104,13 +105,15 @@ public static TcpClientTransport create(TcpClient client) {
                 c -> {
                   if (mtu > 0) {
                     return new FragmentationDuplexConnection(
-                        new TcpDuplexConnection(c, false),
+                        new ReassemblyDuplexConnection(
+                            new TcpDuplexConnection(c, false), ByteBufAllocator.DEFAULT, true),
                         ByteBufAllocator.DEFAULT,
                         mtu,
                         true,
                         "client");
```
rstoyanchev, line 113: This would become a little nicer if `FragmentationDuplexConnection(...)` was a sub-class of `ReassemblyDuplexConnection`

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
OlegDokuka: fixed

rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/client/TcpClientTransport.java
OlegDokuka: fixed

Issue #776, opened 2020-04-10
Unify `ByteBufAllocator` usage among all the places 


OlegDokuka:
## Problem

As of now, we have a mess which allocator we use and where.

For the once glance we have an ability to specify an allocator (https://github.com/rsocket/rsocket-java/blob/develop/rsocket-core/src/main/java/io/rsocket/RSocketFactory.java#L88)
https://github.com/rsocket/rsocket-java/blob/develop/rsocket-core/src/main/java/io/rsocket/RSocketFactory.java#L229

but on the other hand, it is used just in few places and much of them just hard use the DEFAULT one (https://github.com/rsocket/rsocket-java/blob/develop/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/client/WebsocketClientTransport.java#L166)

## Solution

We have to make sure that in all place we use ByteBufAllocator properly so we can test the library on the memory leaks purpose and ensure that the user will not be surprised getting ByteBufs from unexpected places

PR #777, opened 2020-04-10
Track Discarded Payloads

OlegDokuka:
In this PR I wanna introduce extra hooks to ensure we capture and release all the discarded payloads.

In this case, I use `doOnDiscard` operator in order to release dropped elements for every stream we have.

Also, just for now, I created an extra class for `Queue` wrapping in order to ensure that `UnicastProcesssor` does not leak elements on cancellation 

**Please note**, this PR does not address all the cases and a user still MUST install `Hooks.onNextDropped` to ensure there are no leaks if a payload goes beyond rsocket internals scope.


OlegDokuka:
Blocked due to https://github.com/reactor/reactor-core/pull/2114

rsocket-core/src/main/java/io/rsocket/core/CleanOnClearQueueDecorator.java
```diff
@@ -0,0 +1,108 @@
+package io.rsocket.core;
+
+import io.netty.util.ReferenceCountUtil;
+import io.rsocket.Payload;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Queue;
+
+final class CleanOnClearQueueDecorator implements Queue<Payload> {
```
rstoyanchev: This is a temporary workaround for https://github.com/reactor/reactor-core/pull/2114, right? Maybe insert a brief implementation comment to that extent and with a reference to the issue.

rsocket-core/src/main/java/io/rsocket/core/CleanOnClearQueueDecorator.java
```diff
@@ -0,0 +1,108 @@
...
+    this.delegate = delegate;
+  }
+
+  @Override
+  public void clear() {
+    Payload p;
+    while ((p = delegate.poll()) != null) {
+      ReferenceCountUtil.safeRelease(p);
+    }
+  }
```
rstoyanchev, line 24: So this will clear anything remaining in the `UnicastProcessor` queue after cancellation. What about items that may be dropped, depending on exact timing, before they make it into the queue? Shouldn't we also have an `onNextDropped` hook installed as another part of a workaround? An application could also do it but they shouldn't have to.

rsocket-core/src/test/java/io/rsocket/buffer/LeaksTrackingByteBufAllocator.java
```diff
@@ -0,0 +1,192 @@
...
+
+        final LeaksTrackingByteBufAllocator instrumented = new LeaksTrackingByteBufAllocator(ByteBufAllocator.DEFAULT);
+        try {
+            Field field = ByteBufAllocator.class.getField("DEFAULT");
+            setFinalStatic(field, instrumented);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+        return instrumented;
+    }
```
rstoyanchev, line 28: Maybe add implementation comment that this is temporary/necessary until https://github.com/rsocket/rsocket-java/issues/776 is resolved?

rsocket-core/src/test/java/io/rsocket/core/RSocketRequesterTest.java
```diff
@@ -333,6 +346,124 @@ public void shouldThrownExceptionIfGivenPayloadIsExitsSizeAllowanceWithNoFragmen
...
+                      ByteBuf metadata = allocator.buffer();
+                      metadata.writeCharSequence("abc", CharsetUtil.UTF_8);
+                      ByteBuf data = allocator.buffer();
+                      data.writeCharSequence("def", CharsetUtil.UTF_8);
+                      int streamId = rule.getStreamIdForRequestType(REQUEST_CHANNEL);
+                      ByteBuf frame = PayloadFrameFlyweight.encode(allocator, streamId, false, false, true, metadata, data);
+
+                      RaceTestUtils.race(as::cancel, () -> rule.connection.addToReceivedBuffer(frame));
+                    }
+            )/*,*/
```
rstoyanchev: Should we instrument all tests in this class with the leak allocator, as well as `RSocketResponderTest` and others?

rsocket-core/src/main/java/io/rsocket/core/CleanOnClearQueueDecorator.java
OlegDokuka: I decided to simplify the mechanism. This queue directly specified for `Payload`s. Now, this queue extends AtomicBoolean so once clear called (and it is called only on terminal signal) we set the volatile field to `true` so all  the late Payloads coming into the queue will be immediately released

rsocket-core/src/test/java/io/rsocket/buffer/LeaksTrackingByteBufAllocator.java
OlegDokuka: Actually, I was planning to make that as instrumentation for testing to always check that we don't have any leaks

rsocket-core/src/test/java/io/rsocket/core/RSocketRequesterTest.java
OlegDokuka: Yes, working on the responder test but in any case, I ignore them since reactor is leaking so the test can not be passed because of that

rsocket-core/src/main/java/io/rsocket/core/CleanOnClearQueueDecorator.java
OlegDokuka: Yes, it makes sense. In any case, this is package-private so nobody should use it

rsocket-core/src/main/java/io/rsocket/core/CleanOnClearQueueDecorator.java
OlegDokuka: What I can do as well is hacking publishOn operator over reflection and provide cleanOnClearQueueDecorator for the default queue to a contractor to ensure that we can workaround racing-related leaks for now. What do you think about that @rstoyanchev?

rsocket-core/src/main/java/io/rsocket/core/CleanOnClearQueueDecorator.java
```diff
@@ -0,0 +1,114 @@
...
+final class CleanOnClearQueueDecorator extends AtomicBoolean implements Queue<Payload> {
+  final Queue<Payload> delegate;
+
+  CleanOnClearQueueDecorator(Queue<Payload> delegate) {
+    this.delegate = delegate;
+  }
+
+  @Override
+  public void clear() {
+    set(true);
```
OlegDokuka, line 19: @rstoyanchev here is kind of termination of the queue

rsocket-core/src/main/java/io/rsocket/core/CleanOnClearQueueDecorator.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  @Override
+  public boolean retainAll(Collection<?> c) {
+    return delegate.retainAll(c);
+  }
+
+  @Override
+  public boolean offer(Payload payload) {
+    if (get()) {
```
OlegDokuka, line 88: here we check that queue is terminated or not. If it is, then the payload will be released instead of being offered to delegate

rsocket-core/src/test/java/io/rsocket/buffer/LeaksTrackingByteBufAllocator.java
rstoyanchev: Right I guess the question is where the `ByteBufAllocator` will come from and how it will be instrumented. Hopefully not by changing `ByteBufAllocator.DEFAULT` via reflection but yeah for now it's perfectly fine.

rsocket-core/src/main/java/io/rsocket/core/CleanOnClearQueueDecorator.java
rstoyanchev: That's covering even more places, good.

PR #783, opened 2020-04-17
Fixes Payload#hasMetadata to strictly flow the flag in frame

OlegDokuka:
This PR alights with the https://github.com/rsocket/rsocket/pull/302 proposal.
Also, it ensures that ByteBufPayload throws an exception when one tries to use the released instance


OlegDokuka:
Also, following the discussion at the specification thread mentioned earlier, EmptyMetadata should be supported with `hasMetadata` returning `true`. Thus will provide another workaround

rstoyanchev:
Yes I like having `hasMetadata` map to the setting of the metadata flag, and nothing more. The Javadoc would need to be adjusted to reflect that.

PR #794, opened 2020-04-23
fixes requestChannel and ensure support of half-closed state

OlegDokuka:
This PR fixes a bug on the requester side and ensures that the requester is capable of handling requests when its publisher has already sent or received completion.

The Bug here is in the fact that RequestChannel on the requester side has 2 maps to track the state:
 - `receivers` which is there to receive OnNext/OnError/OnComplete signals 
 - `senders` -  requestChannel specific map to hold a subscription and receive the request/cancel signals from the remote.

This means if one is absent and another is present - the stream is in the half-closed state.

Originally, we had that -> https://github.com/rsocket/rsocket-java/blob/develop/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java#L562

```
Subscriber<Payload> receiver = receivers.get(streamId);
    if (receiver == null) {
      handleMissingResponseProcessor(streamId, type, frame);
    }
```

which means that if there is no receiver - it is impossible to handle any frames even though the sender map contains handling subscriptions.

Due to the spec, there is a half-closed state when the responder may send incomplete, but the requester may still continue its stream and send onXXX signals along with being able to receive demand (requestN) and cancel frames.

OUTER PUBLISHER COMPLETED - INNER PUBLISHER CAN SEND
OUTER PUBLISHER CANCELLED - INNER PUBLISHER CAN SEND
INNER PUBLISHER COMPLETED - OUTER PUBLISHER CAN SEND
INNER PUBLISHER CANCELLED - the WHOLE CHAIN IS TERMINATED

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
@@ -604,8 +604,8 @@ class RSocketRequester implements RSocket {
         {
           Subscription sender = senders.get(streamId);
           if (sender != null) {
-            long n = RequestNFrameFlyweight.requestN(frame);
-            sender.request(n);
+            int n = RequestNFrameFlyweight.requestN(frame);
+            sender.request(n >= Integer.MAX_VALUE ? Long.MAX_VALUE : n);
           }
           break;
         }
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
@@ -560,46 +560,58 @@ class RSocketRequester implements RSocket {
 
   private void handleFrame(int streamId, FrameType type, ByteBuf frame) {
     Subscriber<Payload> receiver = receivers.get(streamId);
-    if (receiver == null) {
-      handleMissingResponseProcessor(streamId, type, frame);
-    } else {
-      switch (type) {
-        case ERROR:
-          receiver.onError(Exceptions.from(streamId, frame));
-          receivers.remove(streamId);
-          break;
-        case NEXT_COMPLETE:
-          receiver.onNext(payloadDecoder.apply(frame));
-          receiver.onComplete();
-          break;
-        case CANCEL:
-          {
-            Subscription sender = senders.remove(streamId);
-            if (sender != null) {
-              sender.cancel();
-            }
-            break;
+    switch (type) {
+      case NEXT:
+        if (receiver == null) {
+          handleMissingResponseProcessor(streamId, type, frame);
+          return;
+        }
+        receiver.onNext(payloadDecoder.apply(frame));
+        break;
+      case NEXT_COMPLETE:
+        if (receiver == null) {
+          handleMissingResponseProcessor(streamId, type, frame);
+          return;
+        }
+        receiver.onNext(payloadDecoder.apply(frame));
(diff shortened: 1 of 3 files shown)


rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -560,46 +560,58 @@ private void handleStreamZero(FrameType type, ByteBuf frame) {
...
-              int n = RequestNFrameFlyweight.requestN(frame);
-              sender.request(n >= Integer.MAX_VALUE ? Long.MAX_VALUE : n);
-            }
-            break;
+        }
+      case REQUEST_N:
+        {
+          Subscription sender = senders.get(streamId);
+          if (sender != null) {
+            long n = RequestNFrameFlyweight.requestN(frame);
```
simonbasle, line 607: is that the bug you're mentioning in the PR body? ie if the `RequestNFrameFlyweight.requestN` already used to return a `long` that was cast into an `int`, meaning that any request in the range `Integer.MAX_VALUE..Long.MAX_VALUE` was previously propagated as `Long.MAX_VALUE` ? 

In any case, I would suggest creating a dedicated issue describing this bug for reference (this PR can fix both the partial close and the request bug, but at least the later becomes searchable in issues, which is likely needed)

OlegDokuka: That was pulled from a different PR acidentially

OlegDokuka: The Bug if that RequestChannel on the requester side has 2 maps to track the state:
 - `receivers` which is there to receive OnNext/OnError/OnComplete signals 
 - `senders` -  requestChannel specific map to hold a subscription and receive the request/cancel signals from the remote.

Originally, we had that -> https://github.com/rsocket/rsocket-java/blob/develop/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java#L562

```
Subscriber<Payload> receiver = receivers.get(streamId);
    if (receiver == null) {
      handleMissingResponseProcessor(streamId, type, frame);
    }
```

which means that if there is no receiver - it is impossible to handle any frames.

However, due to the spec, there is half-closed state when the responder may send incomplete, but the requester may still continue its stream and send onXXX signals along with being able to receive demand (requestN) and cancel frames.

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -560,46 +560,58 @@ private void handleStreamZero(FrameType type, ByteBuf frame) {
 
   private void handleFrame(int streamId, FrameType type, ByteBuf frame) {
     Subscriber<Payload> receiver = receivers.get(streamId);
-    if (receiver == null) {
```
OlegDokuka, line 563: this statement is basically a bug which prevents handling the `requestN` frames when the receiver is completed

OlegDokuka: cc/ @simonbasle @rstoyanchev

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
simonbasle: yeah, I figured this was a separate issue. ok then, separate PR is good 👍

PR #796, opened 2020-04-24
Refactors how `ByteBufAllocator` is used and setup

OlegDokuka:
Ensures that Transport setups and provides access to Allocator whereas the RSocketRequester/RSocketResponder may get access to set up allocator over DuplexConnection#alloc method


OlegDokuka:
Yeah, I'm up!

PR #807, opened 2020-04-30
provides supportive class to hide concurrency complexity

OlegDokuka:
This PR includes extra supportive class which should be used only for requestXXX interactions via `Operators.lift` in order to simplify business logic inside `RSocketRequester` (e.g. wipLoops and `cancel` / `request` signals serialization)

***Also***, this PR, ensures that `streamId` will be generated exactly during the first request phase so we will not waste sequence in case stream was not subscribed. The same is relevant to 

```java

  private final IntObjectMap<Subscription> senders;
  private final IntObjectMap<Processor<Payload, Payload>> receivers;
```

holders which are going to get logical stream `put` in it only in case of successful subscription to the generated interaction with the subsequent `request(n)` call on it. 

Note. `StreamId` should not be issued until the first `requestN` as well as the holders should not store anything prior to the first frame sent (it means that if cancel signal happened prior the first request, the only thing we have to do is releasing given payload)

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
@@ -393,118 +393,129 @@ class RSocketRequester implements RSocket {
 
   private Flux<? extends Payload> handleChannel(Payload initialPayload, Flux<Payload> inboundFlux) {
     final UnboundedProcessor<ByteBuf> sendProcessor = this.sendProcessor;
-    final int streamId = streamIdSupplier.nextStreamId(receivers);
 
     final UnicastProcessor<Payload> receiver = UnicastProcessor.create();
-    final BaseSubscriber<Payload> upstreamSubscriber =
-        new BaseSubscriber<Payload>() {
 
-          boolean first = true;
+    return receiver.transform(
+        Operators.<Payload, Payload>lift(
+            (s, actual) ->
+                new RequestOperator(actual) {
+                  int streamId;
 
-          @Override
-          protected void hookOnSubscribe(Subscription subscription) {
-            // noops
-          }
+                  final BaseSubscriber<Payload> upstreamSubscriber =
+                      new BaseSubscriber<Payload>() {
 
-          @Override
-          protected void hookOnNext(Payload payload) {
-            if (first) {
-              // need to skip first since we have already sent it
-              // no need to release it since it was released earlier on the request establishment
-              // phase
-              first = false;
-              request(1);
-              return;
-            }
-            if (!PayloadValidationUtils.isValid(mtu, payload)) {
-              payload.release();
-              cancel();
-              final IllegalArgumentException t =
-                  new IllegalArgumentException(INVALID_PAYLOAD_ERROR_MESSAGE);
-              errorConsumer.accept(t);
-              // no need to send any errors.
-              sendProcessor.onNext(CancelFrameFlyweight.encode(allocator, streamId));
-              receiver.onError(t);
-              return;
-            }
-            final ByteBuf frame =
-                PayloadFrameFlyweight.encodeNextReleasingPayload(allocator, streamId, payload);
+                        boolean first = true;
 
-            sendProcessor.onNext(frame);
-          }
-
-          @Override
-          protected void hookOnComplete() {
-            ByteBuf frame = PayloadFrameFlyweight.encodeComplete(allocator, streamId);
-            sendProcessor.onNext(frame);
-          }
-
-          @Override
-          protected void hookOnError(Throwable t) {
-            ByteBuf frame = ErrorFrameFlyweight.encode(allocator, streamId, t);
-            sendProcessor.onNext(frame);
-            receiver.onError(t);
-          }
-
-          @Override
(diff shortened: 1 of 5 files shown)


rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -257,21 +251,41 @@ public boolean isDisposed() {
...
-                      sendProcessor.onNext(CancelFrameFlyweight.encode(allocator, streamId));
-                    }
-                    removeStreamReceiver(streamId);
-                  })
+              .transform(
+                  Operators.<Payload, Payload>lift(
+                      (s, actual) ->
+                          new RequestOperator(actual) {
+
+                            int streamId;
```
OlegDokuka, line 259: @rstoyanchev maybe it worth moving that field to base class as well to avoid identical on every anonymous class

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -305,62 +314,52 @@ public boolean isDisposed() {
...
-                      }
-                    }
-                    removeStreamReceiver(streamId);
-                  })
+              .transform(
+                  Operators.<Payload, Payload>lift(
+                      (s, actual) ->
+                          new RequestOperator(actual) {
+
+                            int streamId;
```
OlegDokuka, line 322: see https://github.com/rsocket/rsocket-java/pull/807/files#r418297084

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -394,141 +393,124 @@ public void accept(long n) {
...
     final UnicastProcessor<Payload> receiver = UnicastProcessor.create();
-    final BaseSubscriber<Payload> upstreamSubscriber =
-        new BaseSubscriber<Payload>() {
 
-          boolean first = true;
+    return receiver.transform(
+        Operators.<Payload, Payload>lift(
+            (s, actual) ->
+                new RequestOperator(actual) {
+                  int streamId;
```
OlegDokuka, line 403: see https://github.com/rsocket/rsocket-java/pull/807/files#r418297084

rsocket-core/src/main/java/io/rsocket/core/RequestOperator.java
```diff
@@ -0,0 +1,189 @@
...
+import reactor.core.publisher.SignalType;
+import reactor.util.context.Context;
+
+/**
+ * This class is a supportive class to simplify requests' logic implementation and aimed to be used
+ * with {@link Operators#lift}
+ *
+ * <p>This class incorporates extra logic which ensures serial cancel and first request execution as
+ * well as provides extra logic to separate first {@link Subscription#request} invocation from the
+ * rest of them
```
rstoyanchev, line 18:
```suggestion
 * This is a support class for handling of request input, intended for use with {@link Operators#lift}.
 * It ensures serial execution of cancellation vs first request signals and also provides hooks for separate
 * handling of first vs subsequent {@link Subscription#request} invocations.
```

rsocket-core/src/main/java/io/rsocket/core/RequestOperator.java
```diff
@@ -0,0 +1,189 @@
...
+   *
+   * <p><b>Note</b>: this hook may not be invoked if cancellation happened before this invocation
+   */
+  void hookOnFirstRequest(long n) {}
+
+  /**
+   * Optional hook executed after the {@link Subscription#request} was propagated to the upstream
+   * subscription and excludes the first {@link Subscription#request} invocation.
+   */
+  void hookOnRestRequests(long n) {}
```
rstoyanchev, line 50: `hookOnRemainingRequests` or `hookOnNextRequests` perhaps?

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
rstoyanchev: Good idea. That class is for handling requests.

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -257,21 +251,41 @@ public boolean isDisposed() {
...
+                            }
+
+                            @Override
+                            void hookOnCancel() {
+                              if (receivers.remove(streamId, receiver)) {
+                                sendProcessor.onNext(
+                                    CancelFrameFlyweight.encode(allocator, streamId));
+                              } else {
+                                payload.release();
+                              }
```
rstoyanchev, line 281: Nice one! Avoiding the reliance on the refCnt to check if the request was sent.

rsocket-core/src/main/java/io/rsocket/core/RequestOperator.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
OlegDokuka: fixed

PR #811, opened 2020-05-02
provides ordered stream id issuing

OlegDokuka:
This PR ensures that stream ids are issued in order even under high racing between logical streams in the way of scheduling stream id issuing on the event loop scheduler. 

This is done on the level of RSocket because of the incorrect Transport level design and should be considered as a workaround until DuplexConnection will be revised in 1.1


OlegDokuka:
I guess you are right. We can sacrifice a new API method in favor of just solving the problem

OlegDokuka:
@rstoyanchev fixed. Followed your suggestion to use `Scheduler.single(Schedulers.parallel())` Since this is a workaround, this should never be public API until one request for that

PR #813, opened 2020-05-04
Add Javadoc to RSocketConnector and RSocketServer

### LINKED ISSUE #769: Add Javadoc to RSocketFactory

rstoyanchev:
Closes #769

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
@@ -22,6 +22,7 @@ import io.rsocket.DuplexConnection;
 import io.rsocket.Payload;
 import io.rsocket.RSocket;
 import io.rsocket.SocketAcceptor;
+import io.rsocket.fragmentation.FragmentationDuplexConnection;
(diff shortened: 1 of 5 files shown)


rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
```diff
@@ -72,53 +94,107 @@
...
   public static Mono<RSocket> connectWith(ClientTransport transport) {
     return RSocketConnector.create().connect(() -> transport);
   }
 
+  /**
+   * Provide a {@code Payload} with data and/or metadata for the initial {@code SETUP} frame.
+   *
+   * <p>For metadata encoding, consider using one of the following encoders:
+   *
+   * <ul>
```
OlegDokuka, line 124: I guess it worth moving that section to `metadataMimeType` method since `CompositeMetadata` works only in case the specific metadata mime-type was set. 

WDYT on that?

rstoyanchev: Good idea.

PR #824, opened 2020-05-07
provides support for TracingMetadata

### LINKED ISSUE #583: Add Zipkin Tracing Flyweight

OlegDokuka:
closes #583 

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/metadata/TracingMetadata.java
+++ b/rsocket-core/src/main/java/io/rsocket/metadata/TracingMetadata.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2015-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.rsocket.metadata;
+
+public final class TracingMetadata {
+
+  final long traceIdHigh;
+  final long traceId;
+  private final boolean hasParentId;
+  final long parentId;
+  final long spanId;
+  final boolean isEmpty;
+  final boolean isNotSampled;
+  final boolean isSampled;
+  final boolean isDebug;
+
+  TracingMetadata(
+      long traceIdHigh,
+      long traceId,
+      long spanId,
+      boolean hasParentId,
+      long parentId,
+      boolean isEmpty,
+      boolean isNotSampled,
+      boolean isSampled,
+      boolean isDebug) {
+    this.traceIdHigh = traceIdHigh;
+    this.traceId = traceId;
+    this.spanId = spanId;
+    this.hasParentId = hasParentId;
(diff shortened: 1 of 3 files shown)


OlegDokuka:
cc\ @adriancole appreciate your feedback (also see https://github.com/rsocket/rsocket/blob/master/Extensions/Tracing-Zipkin.md for more info)

adriancole:
I'd suggest focusing on the outputs vs the inputs. Ex instead of describing features like isSamplingEnabled and isDebuggingEnabled (knowing there may be different sampling algos etc).. focus only on the resulting data that you'd take action on.


isReported -> isSampled: https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/SamplingFlags.java#L83

if you want to support debug you should also know the difference between this and sampled
isDebug: https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/SamplingFlags.java#L108

Not now, but just fyi, there's a third option which is that data is always sampled, but it is not propagated in trace headers/frames. This is for recording everything for local processing such as metrics or [secondary sampling](https://github.com/openzipkin-contrib/zipkin-secondary-sampling)

isSampledLocal: 
https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/SamplingFlags.java#L100


Hope this helps!

spencergibb:
He's trying to implement the extension spec he mentioned.

adriancole:
sorry missed that link.. I say "Add Zipkin Tracing Flyweight" and had no idea what it meant.

In that case, just retain comments about isSampled and isDebug and forget about isSampledLocal as that's not propagated.

OlegDokuka:
Alright, by looking more into all the stuff I tend to say that current RSocket spec is incorrect and misaligned with today's implementation of TraceContext that I found at brave. Fixing that first and then continue on this impl

OlegDokuka:
@adriancole do you have a few minutes to chat on the TraceContext topic and discuss all the flags that it has and their importance / behavior?

adriancole:
ok find me here https://gitter.im/openzipkin/zipkin

rsocket-core/src/main/java/io/rsocket/metadata/TracingMetadata.java
```diff
@@ -0,0 +1,104 @@
...
+  final long traceId;
+  private final boolean hasParentId;
+  final long parentId;
+  final long spanId;
+  final boolean isEmpty;
+  final boolean isNotSampled;
+  final boolean isSampled;
+  final boolean isDebug;
+
+  TracingMetadata(
```
rstoyanchev, line 31: Some minimal Javadoc with an `@since 1.0` tag would be good.

rsocket-core/src/main/java/io/rsocket/metadata/TracingMetadataCodec.java
```diff
@@ -0,0 +1,166 @@
...
+public class TracingMetadataCodec {
+
+  static final int FLAG_EXTENDED_TRACE_ID_SIZE = 0b0000_1000;
+  static final int FLAG_INCLUDE_PARENT_ID = 0b0000_0100;
+  static final int FLAG_NOT_SAMPLED = 0b0001_0000;
+  static final int FLAG_SAMPLED = 0b0010_0000;
+  static final int FLAG_DEBUG = 0b0100_0000;
+  static final int FLAG_IDS_SET = 0b1000_0000;
+
+  public static ByteBuf encodeEmpty(ByteBufAllocator allocator, Flags flag) {
```
rstoyanchev, line 15: Likewise some minimal Javadoc with an `@since 1.0` tag. Maybe a bit on the encode and decode methods but it doesn't need to hold up 1.0.

PR #832, opened 2020-05-10
removes deprecated errorConsumer

OlegDokuka:

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
+++ b/rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
@@ -111,7 +111,7 @@ public final class RSocketFactory {
     private Resume resume;
 
     public ClientRSocketFactory() {
-      this(RSocketConnector.create().errorConsumer(Throwable::printStackTrace));
+      this(RSocketConnector.create());
     }
 
     public ClientRSocketFactory(RSocketConnector connector) {
@@ -395,7 +395,6 @@ public final class RSocketFactory {
 
     /** @deprecated this is deprecated with no replacement. */
     public ClientRSocketFactory errorConsumer(Consumer<Throwable> errorConsumer) {
-      connector.errorConsumer(errorConsumer);
       return this;
     }
 
@@ -417,7 +416,7 @@ public final class RSocketFactory {
     private Resume resume;
 
     public ServerRSocketFactory() {
-      this(RSocketServer.create().errorConsumer(Throwable::printStackTrace));
+      this(RSocketServer.create());
     }
 
     public ServerRSocketFactory(RSocketServer server) {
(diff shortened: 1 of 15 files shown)


rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
```diff
@@ -395,7 +395,6 @@ public ClientRSocketFactory fragment(int mtu) {
 
     /** @deprecated this is deprecated with no replacement. */
     public ClientRSocketFactory errorConsumer(Consumer<Throwable> errorConsumer) {
-      connector.errorConsumer(errorConsumer);
       return this;
     }
```
rstoyanchev, line 399: Let's mention in the Javadoc that as of 1.0 this method is a no-op.

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
```diff
@@ -499,7 +498,6 @@ public ServerRSocketFactory fragment(int mtu) {
 
     /** @deprecated this is deprecated with no replacement. */
     public ServerRSocketFactory errorConsumer(Consumer<Throwable> errorConsumer) {
-      server.errorConsumer(errorConsumer);
       return this;
     }
```
rstoyanchev, line 502: Same here, mention the method is a no-op.

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -610,9 +611,9 @@ private void handleStreamZero(FrameType type, ByteBuf frame) {
         break;
       default:
         // Ignore unknown frames. Throwing an error will close the socket.
-        errorConsumer.accept(
-            new IllegalStateException(
-                "Client received supported frame on stream 0: " + frame.toString()));
+        if (LOGGER.isDebugEnabled()) {
+          LOGGER.debug("Requester received unsupported frame on stream 0: " + frame.toString());
+        }
```
rstoyanchev, line 616: I wonder if this shouldn't be visible in production logs, perhaps `INFO` level?

rstoyanchev, line 616: I wonder if this shouldn't be visible in production logs, perhaps `INFO` level?

OlegDokuka: Interesting question. This is non-harmful thing that might happen. In general, it means that there is something wrong but it does not break the general work

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/RSocketFactory.java
OlegDokuka: fixed

PR #833, opened 2020-05-10
fixes onClose behaviour to not error on shutdown

### LINKED ISSUE #819: Interaction between RSocket onClose and dispose
In 1.0.0-RC7 it seems like for `onClose()` to be subscribable, we have to do something with the "error" resulting from the `dispose()` that previously was not required. This is a _passing_ test written in `RSocketTest` that demonstrates what seems to be required:

```java
@Test
public void disposeAndOnClose() throws InterruptedException {
  CountDownLatch latch = new CountDownLatch(1);
  rule.crs.onClose()
          .map(v -> 1)
          .onErrorReturn(1) // I only care to know that it was closed, not why
          .subscribe(closed -> latch.countDown());
  rule.crs.dispose();

  assertThat(latch.await(1, TimeUnit.SECONDS)).isTrue();
}
```

Otherwise you get the exception:

```
reactor.core.Exceptions$ErrorCallbackNotImplemented: java.util.concurrent.CancellationException: Disposed
Caused by: java.util.concurrent.CancellationException: Disposed
	at io.rsocket.core.RSocketRequester.lambda$dispose$1(RSocketRequester.java:188)
	at io.rsocket.core.RSocketRequester.tryTerminate(RSocketRequester.java:690)
	at io.rsocket.core.RSocketRequester.dispose(RSocketRequester.java:188)
```

I did the weird `map(v -> 1).onErrorReturn(1)` thing because for the latch to count down we need a non-empty `Mono<Void>`, but paradoxically cannot create a `Mono<Void>` with null, the only valid value for `Void`. I'm probably just not seeing how to do this more idiomatically.

OlegDokuka:
closes #819 

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
@@ -50,7 +50,6 @@ import io.rsocket.keepalive.KeepAliveHandler;
 import io.rsocket.keepalive.KeepAliveSupport;
 import io.rsocket.lease.RequesterLeaseHandler;
 import java.nio.channels.ClosedChannelException;
-import java.util.concurrent.CancellationException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.function.Consumer;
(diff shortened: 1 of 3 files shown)


jkschneider:
Thanks @OlegDokuka

PR #836, opened 2020-05-15
Apply FragmentationDuplexConnection from a single place

rstoyanchev:
In b15a1fccae8a482dbf1a7e0d5cbdb538785b7e5e transports started applying ` FragmentationDuplexConnection`. This was probably done so that TCP transports can give up control of dealing with the frame length to the fragmentation layer. However on closer look there is no obvious reason to not leave the frame length to transports to deal with.

The two commits here restore fragmentation being applied from a single place and make transports unaware of that logic eliminating duplication in the process.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
@@ -23,6 +23,7 @@ import io.rsocket.Payload;
 import io.rsocket.RSocket;
 import io.rsocket.SocketAcceptor;
 import io.rsocket.fragmentation.FragmentationDuplexConnection;
+import io.rsocket.fragmentation.ReassemblyDuplexConnection;
 import io.rsocket.frame.SetupFrameCodec;
 import io.rsocket.frame.decoder.PayloadDecoder;
 import io.rsocket.internal.ClientServerInputMultiplexer;
@@ -400,14 +401,7 @@ public class RSocketConnector {
    *     and Reassembly</a>
    */
   public RSocketConnector fragment(int mtu) {
-    if (mtu > 0 && mtu < FragmentationDuplexConnection.MIN_MTU_SIZE || mtu < 0) {
-      String msg =
-          String.format(
-              "The smallest allowed mtu size is %d bytes, provided: %d",
-              FragmentationDuplexConnection.MIN_MTU_SIZE, mtu);
-      throw new IllegalArgumentException(msg);
-    }
-    this.mtu = mtu;
+    this.mtu = FragmentationDuplexConnection.assertMtu(mtu);
     return this;
   }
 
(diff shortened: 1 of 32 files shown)


rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
```diff
@@ -468,8 +469,19 @@ public RSocketConnector payloadDecoder(PayloadDecoder decoder) {
    * @return a {@code Mono} with the connected RSocket
    */
   public Mono<RSocket> connect(Supplier<ClientTransport> transportSupplier) {
+
     Mono<DuplexConnection> connectionMono =
-        Mono.fromSupplier(transportSupplier).flatMap(t -> t.connect(mtu));
+        Mono.fromSupplier(transportSupplier)
+            .flatMap(
+                transport -> {
+                  Mono<DuplexConnection> errorMono = FragmentationDuplexConnection.checkMtu(mtu);
```
OlegDokuka, line 477: This is redundant since we have already checked `mtu` validity on the factory level

rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
```diff
@@ -272,8 +273,12 @@ public RSocketServer payloadDecoder(PayloadDecoder decoder) {
 
           @Override
           public Mono<T> get() {
+            Mono<T> errorMono = FragmentationDuplexConnection.checkMtu(mtu);
+            if (errorMono != null) {
+              return errorMono;
+            }
```
OlegDokuka, line 279: Same here

rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
```diff
@@ -294,6 +299,7 @@ public RSocketServer payloadDecoder(PayloadDecoder decoder) {
    * @see io.rsocket.examples.transport.ws.WebSocketHeadersSample
    */
   public ServerTransport.ConnectionAcceptor asConnectionAcceptor() {
+    FragmentationDuplexConnection.assertMtu(mtu);
```
OlegDokuka, line 302: Same here

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
```diff
@@ -49,6 +49,23 @@
...
+  public FragmentationDuplexConnection(DuplexConnection delegate, int mtu, String type) {
+    this(delegate, mtu, false, type);
+  }
+
+  /**
+   * @deprecated as of 1.0.1 in favor of {@link #FragmentationDuplexConnection(DuplexConnection,
+   *     int, String)} and hence {@code decodeLength} should always be false.
+   */
+  @Deprecated
   public FragmentationDuplexConnection(
```
OlegDokuka, line 69: I believe we can remove that one at all

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
```diff
@@ -79,7 +96,7 @@ private boolean shouldFragment(FrameType frameType, int readableBytes) {
     }
   }
 
-  private static int assertMtu(int mtu) {
+  public static int assertMtu(int mtu) {
```
OlegDokuka, line 99: redundant method. Lets remove it

rsocket-core/src/main/java/io/rsocket/fragmentation/ReassemblyDuplexConnection.java
```diff
@@ -37,6 +37,20 @@
...
+   */
+  public ReassemblyDuplexConnection(DuplexConnection delegate) {
+    this(delegate, false);
+  }
+
+  /**
+   * @deprecated as of 1.0.1 in favor of {@link #ReassemblyDuplexConnection(DuplexConnection)} and
+   *     hence {@code decodeLength} should always be false.
+   */
+  @Deprecated
```
OlegDokuka, line 53: Let's remove deprecation at all

rstoyanchev:
I've cleaned up further as suggested.

PR #847, opened 2020-05-21
fixes ReconnectMono behaviour when racing invalidate and subscribe #847

OlegDokuka:
Found that there is a bug in ReconnectMono when it appears that `Reconnect#invalidate` is raced with `Reconnect#subscribe` method which is possible when e.g.

```
Mono<RSocket> rsocketMono =
  RSocketConnector.create()
        .reconnect(Retry.fixedDelay(3, Duration.ofSeconds(1)))
        .connect(transport)
        .flatMap(rSocket -> {
            return rsocket.requestStream(...);
        })
        .retryWhen(Retry.backoff(10000, Duration.ofMillis(1)))
        .subscribe();
```

subscriber retried, subscribed, resolved connection while the thread that invalidating RSocket has executed CAS operation and in the process of setting `this.value = null`.

In general, including a few more omitted cases, there can be 3 outcomes

1. The subscriber that observed state `RSocket is READY` can observe the actual value as null because of racing with invalidate
2. The subscriber may observe state e.g `INVALIDATED and need to subscribe to the source again`, hence subscribe to source, but racing between resolving value and invalidation may end up with null value after all 
3. Everything can be fine

All tests expose those case and fix is pretty straightforward

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/ReconnectMono.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/ReconnectMono.java
@@ -112,12 +112,25 @@ final class ReconnectMono<T> extends Mono<T> implements Invalidatable, Disposabl
     final ReconnectInner<T> inner = new ReconnectInner<>(actual, this);
     actual.onSubscribe(inner);
 
-    final int state = this.add(inner);
+    for (; ; ) {
+      final int state = this.add(inner);
 
-    if (state == READY_STATE) {
-      inner.complete(this.value);
-    } else if (state == TERMINATED_STATE) {
-      inner.onError(this.t);
+      T value = this.value;
+
+      if (state == READY_STATE) {
+        if (value != null) {
+          inner.complete(value);
+          return;
+        }
+        // value == null means racing between invalidate and this subscriber
+        // thus, we have to loop again
+        continue;
+      } else if (state == TERMINATED_STATE) {
+        inner.onError(this.t);
+        return;
+      }
+
+      return;
     }
   }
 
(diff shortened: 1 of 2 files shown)


rsocket-core/src/main/java/io/rsocket/core/ReconnectMono.java
```diff
@@ -355,6 +400,11 @@ void remove(ReconnectInner<T> ps) {
     }
   }
 
+  /**
+   * Subscriber that subscribes to the source {@link Mono} in order to resolve value. <br>
+   * Note, implementation of this subscribers exclude possibility to receive onComplete only and if
+   * such happens, terminates execution with an error
+   */
```
rstoyanchev, line 407:
```suggestion
  /**
   * Subscriber that subscribes to the source {@link Mono} to receive its value. <br>
   * Note that the source is not expected to complete empty, and if this happens,
   * execution will terminate with an {@code IllegalStateException}.
   */
```

rsocket-core/src/main/java/io/rsocket/core/ReconnectMono.java
```diff
@@ -389,7 +439,7 @@ public void onComplete() {
       }
 
       if (value == null) {
-        p.terminate(new IllegalStateException("Unexpected Completion of the Upstream"));
+        p.terminate(new IllegalStateException("Unexpected empty source"));
```
rstoyanchev, line 442:
```suggestion
        p.terminate(new IllegalStateException("Source completed empty"));
```

PR #849, opened 2020-05-25
provides draft implementation for the base RSocketPool

OlegDokuka:
This PR provides a base implementation for RSocketPool which is thread-safe and non-blocking from the modification perspective.

It uses ResolvingOperator as a base-class.
This PR provides an example implementation of the RoundRobinPool in order to show how BaseRSocketPool can be extended


OlegDokuka:
cc\ @pmacho

PMacho:
Hi @OlegDokuka,

I am not yet done, checking your code. However the for loop in BaseRSocketPool.onNext() could be replaced with:

      HashSet<RSocket> set = new HashSet<>(sockets);
      HashSet<RSocket> newSockets = new HashSet<>(sockets.size());

      for (RSocket rSocket : activeSockets) {
        if (set.contains(rSocket)){
          newSockets.add(rSocket);
        } else {
          // todo:
          // rSocket.setPending()
        }
      }
      newSockets.addAll(sockets);

      if (ACTIVE_SOCKETS.compareAndSet(this, activeSockets, (RSocket[]) newSockets.toArray())) {
        break;
      }

But actually it would be even enough to just do:

      HashSet<RSocket> set = new HashSet<>(sockets);

      for (RSocket rSocket : activeSockets) {
        if (!set.contains(rSocket)){
          // todo:
          //  rSocket.setPending()
        }
      }

      if (ACTIVE_SOCKETS.compareAndSet(this, activeSockets, (RSocket[]) sockets.toArray())) {
        break;
      }

... since there is no difference between the old and the new RSocket. 

Plus: Your code would add not disposed old sockets to the new once. I think, this is not good, since they stay available for selection.

PR #855, opened 2020-06-01
ensures lease is used exactly on request frame sending

OlegDokuka:
This PR ensures that lease is used right before when the actual request frame is being sent.

That said that right now lease is used even though the `Publisher` given as the result of the `requestXXX` method execution is not subscribed (which of course may break lease counting on the client- vs server-side)

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
@@ -230,6 +230,7 @@ class RSocketRequester implements RSocket {
 
               RequesterLeaseHandler lh = leaseHandler;
               if (!lh.useLease()) {
+                payload.release();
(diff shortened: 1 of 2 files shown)


rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -228,6 +228,11 @@ public boolean isDisposed() {
                 return Mono.error(t);
               }
 
+              RequesterLeaseHandler lh = leaseHandler;
+              if (!lh.useLease()) {
+                return Mono.error(lh.leaseError());
```
rstoyanchev, line 233: Also release payload here.

OlegDokuka: fixed

PR #876, opened 2020-06-24
provides setup for `maxInboundPayloadSize` and `maxFrameSize`

### LINKED ISSUE #864: Client and server are trivially subject to memory overflow on frame reassembly
Client and server are subject to OOM because `FrameReassembler` does not have upper bound on received frame size. Reproducible with code below

Server
```
    CloseableChannel server =
        RSocketServer.create()
            .acceptor((setup, sendingSocket) -> Mono.just(new RSocket() {}))
            .bind(TcpServerTransport.create(7077))
            .block();
    logger.info("Server bound on {}", server.address());
    server.onClose().block();
```

Client
```
    InetSocketAddress address = new InetSocketAddress("localhost", 7077);
    TcpClientTransport.create(address)
        .connect()
        .flatMap(
            duplexConnection -> {
              logger.info("Client connected to {}", address);

              MonoProcessor<Void> onClose = MonoProcessor.create();
              UnicastProcessor<ByteBuf> outbound = UnicastProcessor.create();

              duplexConnection
                  .send(outbound)
                  .subscribe(unused -> onClose.onComplete(), err -> onClose.onError(err));
              duplexConnection
                  .receive()
                  .subscribe(
                      inbound -> inbound.release(),
                      err -> onClose.onError(err),
                      () -> onClose.onComplete());

              outbound.onNext(
                  SetupFrameCodec.encode(
                      ByteBufAllocator.DEFAULT,
                      false,
                      10_000,
                      100_000,
                      "application/binary",
                      "application/binary",
                      EmptyPayload.INSTANCE));
              outbound.onNext(
                  RequestResponseFrameCodec.encode(
                      ByteBufAllocator.DEFAULT,
                      1,
                      true,
                      Unpooled.wrappedBuffer(METADATA),
                      Unpooled.EMPTY_BUFFER));
              for (int i = 0; i < FRAGMENTS_COUNT; i++) {
                outbound.onNext(
                    PayloadFrameCodec.encode(
                        ByteBufAllocator.DEFAULT,
                        1,
                        true,
                        false,
                        true,
                        Unpooled.wrappedBuffer(METADATA),
                        Unpooled.EMPTY_BUFFER));
              }
              logger.info(
                  "Sent {} fragments of total size {} bytes",
                  FRAGMENTS_COUNT,
                  FRAGMENT_SIZE * FRAGMENTS_COUNT);
              return onClose;
            })
        .block();

```

It emulates RSocket client sending infinite sequence of frame fragments for particular request.

OlegDokuka:
closes #864

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
@@ -15,6 +15,8 @@
  */
 package io.rsocket.core;
 
+import static io.rsocket.core.PayloadValidationUtils.assertValidateSetup;
+
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.rsocket.ConnectionSetupPayload;
@@ -92,7 +94,7 @@ public class RSocketConnector {
   private Supplier<Leases<?>> leasesSupplier;
 
   private int mtu = 0;
-  private int maxReassemblySize = Integer.MAX_VALUE;
+  private int maxInboundPayloadSize = Integer.MAX_VALUE;
   private PayloadDecoder payloadDecoder = PayloadDecoder.DEFAULT;
 
   private RSocketConnector() {}
@@ -419,14 +421,18 @@ public class RSocketConnector {
    * <p>By default this is not set in which case maximum reassembled payloads size is not
    * controlled.
    *
-   * @param maxReassemblySize the threshold size for reassembly, must be no less than 16,777,215
+   * @param maxInboundPayloadSize the threshold size for reassembly, must no be less than 64 bytes.
+   *     Please note, {@code maxInboundPayloadSize} must always be greater or equal to {@link
+   *     io.rsocket.transport.Transport#maxFrameLength()}, otherwise inbound frame can exceed the
+   *     {@code maxInboundPayloadSize}
    * @return the same instance for method chaining
    * @see <a
    *     href="https://github.com/rsocket/rsocket/blob/master/Protocol.md#fragmentation-and-reassembly">Fragmentation
    *     and Reassembly</a>
    */
-  public RSocketConnector reassemble(int maxReassemblySize) {
-    this.maxReassemblySize = ReassemblyDuplexConnection.assertMaxReassemblySize(maxReassemblySize);
+  public RSocketConnector maxInboundPayloadSize(int maxInboundPayloadSize) {
+    this.maxInboundPayloadSize =
+        ReassemblyDuplexConnection.assertInboundPayloadSize(maxInboundPayloadSize);
     return this;
   }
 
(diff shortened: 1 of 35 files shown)


OlegDokuka:
@rstoyanchev right now this implementation terminates the connection if the reassembly limit is overflowed. Should we consider logical stream termination instead?

rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
```diff
@@ -395,6 +397,23 @@ public RSocketConnector lease(Supplier<Leases<? extends LeaseStats>> supplier) {
...
+   * @return the same instance for method chaining
+   * @see <a
+   *     href="https://github.com/rsocket/rsocket/blob/master/Protocol.md#fragmentation-and-reassembly">Fragmentation
+   *     and Reassembly</a>
+   */
+  public RSocketConnector reassemble(int maxReassemblySize) {
+    this.maxReassemblySize = ReassemblyDuplexConnection.assertMaxReassemblySize(maxReassemblySize);
+    return this;
+  }
+
```
rstoyanchev, line 416: I think this should apply to any input payload size regardless of whether it is fragmented or not. In fact by default fragmentation is off and in any case once there is a limit in place for fragmented input, it would be expected to have the same for non-fragmented input. I do understand that the rest is at the transport level. Nevertheless from an application perspective, conceptually, it's one limit.

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameReassembler.java
```diff
@@ -151,6 +173,19 @@ void cancelAssemble(int streamId) {
...
+        int currentPayloadSize = getMetadataSize(streamId) + getDataSize(streamId);
+        if (currentPayloadSize + frame.readableBytes() - FrameHeaderCodec.size()
+            > maxReassemblySize) {
+          header.release();
+          frame.release();
+          cancelAssemble(streamId);
+          throw new IllegalStateException("Reassembled payload went out of allowed size");
+        }
+      }
+
```
rstoyanchev, line 188: If the limit is reached on the very last fragment I think it should be allowed through. It is in memory at this point anyway and the purpose of the limit should be more to prevent further growth after the limit is exceeded or otherwise there would be cases where it fail for even 1 byte over. It should be lenient about that.

rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
OlegDokuka: That is another concern. Spec specifies that Max-Frame size if 16MB, thus it should be clear that RSocket should support that size. Now the question is - if one applies a 1byte memory limit, how a client, which expects that allowed frame size is 16MB, should behave in that case.

From the fragmentation perspective, minimal frame size is 64bytes, so reassembly limit should be more than that at least, but if we wanna send something bigger than that, all tries will fail, so what is the benefit of such reassembly then? 

From my point of view, the minimal storage per stream should be at least the default max frame size. Or at least minimal fragment size specified by spec

rsocket-core/src/main/java/io/rsocket/fragmentation/FrameReassembler.java
OlegDokuka: well. Disagree. This will open another spot of the attack. Since now anyone can flash anything in the pipe, it can configure fragmentation to send the very last frame == MAX FRAME SIZE (16MB) so if the limit of in-memory storage is 16KB, then the last frame will do 16MB + 16KB. Which means - one can use it as a vulnerability

OlegDokuka:
> Could it perhaps release the accumulated fragments, send an error through the `Sink`, and then continue to discard further fragments received for that stream?

I don't sink discarding is a good idea since we will not be able to respond back with any signal.

Also, Sink for ReassemblyDuplexConnection means connection sink so if we send error it automatically terminates transport connection and disposes of RSocket.

If we don't wanna terminate the connection, then we must do it according to the spec - Send error to the responder (depends on the stream type - if it is request-response - ignore payload and send nothing to the responder, but then send an error back to the requester saying something in the error message). If reassembly of an ordinary payload - cancel upstream and error downstream

rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
```diff
@@ -91,6 +94,7 @@
   private Supplier<Leases<?>> leasesSupplier;
 
   private int mtu = 0;
+  private int maxPayloadSize = Integer.MAX_VALUE;
```
OlegDokuka, line 97: Leaves that as unbounded for 1.0.x

OlegDokuka: for 1.1 will be 16 mb

rsocket-core/src/main/java/io/rsocket/transport/Transport.java
```diff
@@ -16,5 +16,22 @@
...
+
+  /**
+   * Configurations that exposes the maximum frame size that a {@link DuplexConnection} can bring up
+   * to RSocket level.
+   *
+   * <p>This number should not exist the 16,777,215 (maximum frame size specified by RSocket spec)
+   *
+   * @return return maximum configured frame size limit
+   */
+  default int maxFrameLength() {
```
OlegDokuka, line 34: moved maxFrameLength property to the transport, since it describes transport behavior and from my point of view checking that when a connection has already wired is too late

OlegDokuka: Also, this one is a default property, so it should not be a breaking change for anybody

rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
```diff
@@ -395,6 +399,23 @@ public RSocketConnector lease(Supplier<Leases<? extends LeaseStats>> supplier) {
...
+   * <p>By default this is not set in which case maximum reassembled payloads size is not
+   * controlled.
+   *
+   * @param maxPayloadSize the threshold size for reassembly, must no be less than 64 bytes
+   * @return the same instance for method chaining
+   * @see <a
+   *     href="https://github.com/rsocket/rsocket/blob/master/Protocol.md#fragmentation-and-reassembly">Fragmentation
+   *     and Reassembly</a>
+   */
+  public RSocketConnector maxPayloadSize(int maxPayloadSize) {
```
OlegDokuka, line 414: Still doubt whether that is the best name. 

What about `maxReceivedPayloadSize` since this is applicable for received payloads, but not for the sent. WDYT @rstoyanchev?

OlegDokuka: Also, I would add linking to Transport#maxFrameSize saying that maxReceivedPayloadSize must be >= to maxFrameSize, otherwise received frame can exit the maxReceivedPayloadSize

rsocket-core/src/main/java/io/rsocket/fragmentation/ReassemblyDuplexConnection.java
```diff
@@ -36,14 +37,26 @@
...
+    if (maxReassemblySize < FragmentationDuplexConnection.MIN_MTU_SIZE) {
+      String msg =
+          String.format(
+              "The smallest allowed maxReassemblySize size is %d bytes, provided: %d",
+              FrameLengthCodec.FRAME_LENGTH_MASK, maxReassemblySize);
+      throw new IllegalArgumentException(msg);
+    } else {
+      return maxReassemblySize;
+    }
+  }
```
rstoyanchev, line 58: It says `max` but actually asserts for min size. Do we even need to assert this?

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
```diff
@@ -53,16 +55,14 @@
    * @param mtu the fragment size, greater than {@link #MIN_MTU_SIZE}
    * @param type a label to use for logging purposes
    */
-  public FragmentationDuplexConnection(DuplexConnection delegate, int mtu, String type) {
-    super(delegate);
+  public FragmentationDuplexConnection(
+      DuplexConnection delegate, int mtu, int maxReassemblySize, String type) {
+    super(delegate, maxReassemblySize);
```
rstoyanchev, line 60: Maybe we can call this `maxPayloadSize` to align the terminology, and also in the base class?

rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
```diff
@@ -198,6 +201,23 @@ public RSocketServer lease(Supplier<Leases<?>> supplier) {
...
+   * @param maxReassemblySize the threshold size for reassembly, must no be less than 64 bytes
+   * @return the same instance for method chaining
+   * @see <a
+   *     href="https://github.com/rsocket/rsocket/blob/master/Protocol.md#fragmentation-and-reassembly">Fragmentation
+   *     and Reassembly</a>
+   */
+  public RSocketServer maxPayloadSize(int maxReassemblySize) {
+    this.maxPayloadSize = ReassemblyDuplexConnection.assertMaxReassemblySize(maxReassemblySize);
+    return this;
+  }
```
rstoyanchev, line 219: Rename argument to `maxPayloadSize`.

rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
```diff
@@ -262,10 +282,41 @@ public RSocketServer payloadDecoder(PayloadDecoder decoder) {
...
           @Override
           public Mono<T> get() {
+            int maxFrameLength = transport.maxFrameLength();
+
+            if (maxFrameLength > FRAME_LENGTH_MASK) {
+              return Mono.error(
+                  new IllegalArgumentException(
+                      "Configured maxFrameLength["
+                          + maxFrameLength
+                          + "] exits maxFrameLength limit "
```
rstoyanchev, line 296: This is meant to be `exceeds`? Search and replace here and in RSocketConnector.

rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
```diff
@@ -262,10 +282,41 @@ public RSocketServer payloadDecoder(PayloadDecoder decoder) {
...
+            if (mtu != 0 && mtu > maxFrameLength) {
+              return Mono.error(
+                  new IllegalArgumentException(
+                      "Configured maximumTransmissionUnit["
+                          + mtu
+                          + "] exits configured maxFrameLength["
+                          + maxFrameLength
+                          + "]"));
+            }
+
```
rstoyanchev, line 319: Could this be encapsulated some place and shared with RSocketConnector? PayloadValidationUtils maybe?

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
```diff
@@ -208,7 +211,7 @@ public boolean isDisposed() {
       return Mono.error(t);
     }
 
-    if (!PayloadValidationUtils.isValid(this.mtu, payload)) {
+    if (!PayloadValidationUtils.isValid(this.mtu, payload, maxFrameLength)) {
```
rstoyanchev, line 214: Shouldn't this be called `maxPayloadLength`? That's what's passed in. Same in `RSocketResponder`.

OlegDokuka: That is different from `maxPayloadSize` fragmentation is aimed to maximum frame length that can be sent over the transport. If transport limits frame size to 128 bytes, then validation should be aligned with that. In contrast, `maxPayloadSize` is more general, since from 10 frames we can gather a single payload. `MaxPayloadSize` should always be >= to `maxFrameLength`

rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
OlegDokuka: it is not directly related to Payload. But I agree that it is duplicated

rsocket-core/src/main/java/io/rsocket/core/RSocketConnector.java
rstoyanchev: Good point. Perhaps `maxInboundPayloadSize` and `maxInboundFrameSize`?

rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
OlegDokuka: yeah...

rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
OlegDokuka: sure

rsocket-core/src/main/java/io/rsocket/fragmentation/FragmentationDuplexConnection.java
OlegDokuka: agree

rsocket-core/src/main/java/io/rsocket/fragmentation/ReassemblyDuplexConnection.java
OlegDokuka: agree. need to fix that

rsocket-core/src/main/java/io/rsocket/core/RSocketRequester.java
OlegDokuka: I could make a mistake passing in payload size instead of maxFrameLenght configured in the transport

rstoyanchev: Ok

PR #881, opened 2020-07-01
Add option to provide Mono<Payload> for the SETUP payload

rstoyanchev:
Currently `RSocketConnector` expects a `Payload` instance for the SETUP frame. However, if the input is from a higher level object, it may be serialized to `Mono<Payload>`. So that needs to be resolved first, then the `setupPayload` can be set on `RSocketConnector`, and only then can `connect` be called. This makes it difficult to avoid re-creating the `Mono<RSocket>` from the connect method per subscriber, which in turn precludes use of the reconnect feature.

This change adds an extra option on `RSocketConnector` to accept a `Mono<Payload>`.


rstoyanchev:
Test added.

PR #885, opened 2020-07-07
refactors Lease API

OlegDokuka:
This PR refactors leases API.

key spots:

1. LeaseTracker - a new interface that allows to track all lease cycles including rejection/acceptance / lease-releasing
2. LeaseSender/LeaseReceiver interfaces as a more certain API way which can be extended later on.

More changes will go into 1.1. As of now, this PR's intention is to deliver minimal useful changes in order to implement a working server-side concurrency control mechanism 


rstoyanchev (commented):
My initial impression, this is a step in the right direction, but there is quite a bit of infrastructure in the advanced sample that should be in the framework. We can take this further to make the Lease API easy to consume so that the "advanced" sample becomes straight forward and the main lease sample. By comparison the current "simple" example really is not at all simple, which is due to the way the existing API is, and it should become unnecessary. 

On the requester side, I think we should consider to incorporate the `DeferringLeaseReceiver` and the resulting `LeaseWaitingRSocket` into framework code and apply transparently. So when leasing is enabled, I can just use the resulting `RSocket`. Maybe I don't even need to supply a `LeaseReceiver`.

On the responding side, the framework could provide a general purpose `RequestInterpceptor` such as the `LimitBasedLeaseSender` but without any dependency on an external library like Netflix Concurrency Limit. Instead what is currently the `LeaseManager` could become an interface with implementations encapsulating the limit algorithm logic while also exposing the necessary methods.

OlegDokuka:
@rstoyanchev on the one hand you are totally right. On the other, I can imagine one doing something unordinary which requires access to sender-receiver logic. What if we expose all that via smart builder API and let one do what is needed?

rsocket-core/src/main/java/io/rsocket/plugins/RequestInterceptor.java
```diff
@@ -76,4 +78,27 @@
...
+                        ri instanceof CompositeRequestInterceptor
+                            ? Arrays.stream(((CompositeRequestInterceptor) ri).requestInterceptors)
+                            : ri instanceof CompositeRequestInterceptor.SafeRequestInterceptor
+                                ? Stream.of(
+                                    ((CompositeRequestInterceptor.SafeRequestInterceptor) ri)
+                                        .requestInterceptor)
+                                : Stream.of(ri))
+                .toArray(RequestInterceptor[]::new));
+    }
+  }
```
rstoyanchev, line 103: This could move to `InitializingInterceptorRegistry` which is already advertised as "not for direct use" and is used by `RSocketConnector` and `RSocketServer`.

rsocket-core/src/main/java/io/rsocket/lease/Lease.java
```diff
@@ -18,93 +18,91 @@
...
-   * Number of milliseconds that this lease is valid from the time it is received.
+   * Number of requests allowed by this lease.
    *
-   * @return Number of milliseconds that this lease is valid from the time it is received.
+   * @return The number of requests allowed by this lease.
    */
-  int getTimeToLiveMillis();
+  public int allowedRequests() {
+    return allowedRequests;
+  }
```
rstoyanchev, line 67: While we're changing these, let's calls this `numberOfRequests` to match the spec (and on `create` above).

rsocket-core/src/main/java/io/rsocket/lease/Lease.java
```diff
@@ -18,93 +18,91 @@
...
 
   /**
    * Absolute time since epoch at which this lease will expire.
    *
    * @return Absolute time since epoch at which this lease will expire.
    */
-  long expiry();
+  public long expireAt() {
+    return expireAt;
+  }
```
rstoyanchev, line 85: `expirationTime()`? or otherwise `expiresAt`.

rsocket-core/src/main/java/io/rsocket/core/LeaseConfig.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  public LeaseConfig deferOnNoLease() {
+    return deferOnNoLease(Queues.SMALL_BUFFER_SIZE);
+  }
+
+  public LeaseConfig deferOnNoLease(int maxPendingRequests) {
+    this.maxPendingRequests = maxPendingRequests;
+    return this;
+  }
```
rstoyanchev, line 54: Perhaps replace the `failOnLease` and `deferOnLease` methods with one that accepts the queue size, like `maxPendingRequests(int)` which can be 0 or more and using some default like `Queues.SMALL_BUFFER_SIZE`.

rsocket-core/src/main/java/io/rsocket/core/LeaseConfig.java
```diff
@@ -0,0 +1,55 @@
...
+
+  public LeaseConfig sender(LeaseSender sender) {
+    this.sender = sender;
+    return this;
+  }
+
+  public LeaseConfig statsCollector(RequestInterceptor requestInterceptor) {
+    this.statsCollector = requestInterceptor;
+    return this;
+  }
```
rstoyanchev, line 40: I'd like to suggest consolidating these into one. For example leave only `sender(LeaseSender)` and check if the `LeaseSender` also implements `RequestInterceptor`. In other words, in all cases I start by implementing `LeaseSender` and if my sender works by tracking requests, I can also implement `RequestInterceptor`. We could even add a `TrackingLeaseSender` sub-interface that is both. 

Later on, if the need arises, we can provide a convenience 2-arg method that takes `LeaseSender` and `RequestInterceptor` as separate instances but I suspect that might not be necessary.

rsocket-core/src/main/java/io/rsocket/core/LeaseConfig.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package io.rsocket.core;
+
+import io.rsocket.lease.LeaseSender;
+import io.rsocket.plugins.RequestInterceptor;
+import reactor.core.publisher.Flux;
+import reactor.util.concurrent.Queues;
+
+public final class LeaseConfig {
```
rstoyanchev, line 24: I think `LeaseConfig` would be better in the `lease` package next to `LeaseSender` and `Lease`.

rstoyanchev (changes requested):
Overall looks good. I'll look some more in the core package. For now sending a few comments identified so far.

rsocket-core/src/main/java/io/rsocket/core/LeaseHandler.java
```diff
@@ -0,0 +1,8 @@
+package io.rsocket.core;
+
+interface LeaseHandler {
+
+  void handleLease();
+
+  void handleError(Throwable t);
+}
```
rstoyanchev, line 8: Maybe rename to `LeasePermitHandler` to make it more clear it handles 1 permit rather than a LEASE frame. The methods could be `handlePermit()` and `handlePermitError(Throwable)` with the latter breaking the overlap with `handleError` from `FrameHandler` which can be difficult to spot otherwise.

rsocket-core/src/main/java/io/rsocket/core/StateUtils.java
```diff
@@ -13,27 +13,36 @@
...
       0b000000000000000000000000000000001_0000000000000000000000000000000L;
+  /** Bit Flag that indicates that the first payload in RequestChannel scenario is received */
+  static final long FIRST_PAYLOAD_RECEIVED_FLAG =
+      0b000000000000000000000000000000010_0000000000000000000000000000000L;
+  /**
+   * Bit Flag that indicates that the logical stream is prepared for the first initial frame sending
+   * (applicable for requester only)
+   */
+  static final long PREPARED_FLAG =
+      0b000000000000000000000000000000100_0000000000000000000000000000000L;
```
rstoyanchev, line 24: Maybe `READY_TO_SEND_FIRST_FRAME_FLAG` and also if the Javadoc could reflect that this is set either when a lease permit is acquired, or otherwise when `request` is received, if leasing is not enabled.

rstoyanchev (approved):
Approved with a few final comments below.

rsocket-examples/src/main/java/io/rsocket/examples/transport/tcp/loadbalancer/RoundRobinRSocketLoadbalancerExample.java
```diff
@@ -54,56 +55,39 @@ public static void main(String[] args) {
...
-            .bindNow(TcpServerTransport.create(8082));
+    //    CloseableChannel server3 =
+    //        RSocketServer.create(
+    //                SocketAcceptor.forRequestResponse(
+    //                    p -> {
+    //                      System.out.println("Server 3 got fnf " + p.getDataUtf8());
+    //                      return Mono.just(DefaultPayload.create("Server 3 response"))
+    //                          .delayElement(Duration.ofMillis(100));
+    //                    }))
+    //            .bindNow(TcpServerTransport.create(8082));
```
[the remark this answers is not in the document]
OlegDokuka: done

rsocket-core/src/main/java/io/rsocket/core/StateUtils.java
OlegDokuka: applied suggestion

rsocket-core/src/main/java/io/rsocket/core/RequestResponseRequesterMono.java
```diff
@@ -134,15 +136,34 @@ public final void request(long n) {
...
+    sendFirstPayload(this.payload);
   }
 
-  void sendFirstPayload(Payload payload, long initialRequestN) {
+  @Override
+  public void handleLease() {
+    final long previousState = markPrepared(STATE, this);
+
+    if (isTerminated(previousState)) {
+      return;
```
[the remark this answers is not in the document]
OlegDokuka: good catchup. Fixed

rsocket-core/src/main/java/io/rsocket/core/LeaseHandler.java
OlegDokuka: applied suggestion

PR #891, opened 2020-07-16
Add RSocket to RSocketClient adapter

rstoyanchev:


rstoyanchev:
@OlegDokuka, note I didn't opt for a `wrap` static factory method in `RSocketClient` because it's more of an auxiliary component and didn't want it to be the only such method in `RSocketClient`. 

Perhaps it can better live in the core package though. WDYT?

OlegDokuka:
LGTM. Minor. Can we use dedicated operators from the `DefaultRSocketClient` class?

e.g. 
https://github.com/rsocket/rsocket-java/blob/master/rsocket-core/src/main/java/io/rsocket/core/DefaultRSocketClient.java#L70

Issue #895, opened 2020-07-24
RSocket server is subject to empty frames flood denial-of-service


mostroverkhov:
RSocket `FrameReassembler` [1](https://github.com/rsocket/rsocket-java/blob/3b2883f9941beda5749812d675df4f793c132e48/rsocket-core/src/main/java/io/rsocket/fragmentation/FrameReassembler.java#L270), [2](https://github.com/rsocket/rsocket-java/blob/3b2883f9941beda5749812d675df4f793c132e48/rsocket-core/src/main/java/io/rsocket/fragmentation/FrameReassembler.java#L244) silently accepts empty `Payload` frames with `follows` flag set - ones with empty both data and metadata.

 In this case RSocket stream does zero progress while consuming non-zero CPU, memory and network resources. Because empty frames rate is limited only by network connection width that can be quite large, this may be exploited by external rogue clients to reduce server capacity, responsiveness and lead to elevated costs caused by such parasite traffic.

This vulnerability is similar in spirit to [CVE-2019-9518 - flood of empty frames](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-9518)

OlegDokuka:
Good catchup 👍

OlegDokuka:
We will be adding validation on the frame size > 64 as well as going to check that if empty or frame size is less than 64 bytes than this frame has to be the last one. Otherwise, the connection will be terminated.

For 1.1 the stream will be terminated instead of the whole connection

mostroverkhov:
I am not sure why this is marked as "enhancement" given It is covered by CVE - anyways, here another "enhancement" candidate - https://github.com/rsocket/rsocket-java/issues/898

OlegDokuka:
@rstoyanchev also, we need to add one more check to ensure that size of all fragments are equal and if the new one is less/greater than the last one, we have to consider it as a terminal, since an attacker can send frames different in size, thus a check for homogenity should be added in place in order to detect a corruption

PR #903, opened 2020-08-03
Reject empty fragments

rstoyanchev:
Proposed fix for #895.


rstoyanchev:
Last commit was just a rebase on top of the latest 1.0.x.

PR #909, opened 2020-08-06
reworks multiplexer and improves tests to check async interceptors #909

OlegDokuka:
This PR reworks `ClientServerInputMultiplexer` to avoid async enqueueing/dequeuing of the incoming frames when interceptors offload subscription/publication on the different `Scheduler`, so frame appears released at the time of its consumption (`FluxReceiver` sends `ByteBuf` and releases it once `onNext` is returned -> https://github.com/reactor/reactor-netty/blob/master/reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java#L264).

Basically this PR gets rid of `groupBy` operator in favor of its drain-free version.

***FYI. This PR as well as changes related to refCoutning mechanics brake resumability***.
It can wait since there is no usage of this feature in Spring-Messaging, though we should revise the whole resumability implementation before the final 1.1 release.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/ClientServerInputMultiplexer.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/ClientServerInputMultiplexer.java
@@ -1,0 +1,399 @@
+/*
+ * Copyright 2015-2020 the original author or authors.
+ *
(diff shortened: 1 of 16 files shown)


OlegDokuka:
@bclozel @rwinch this should fix the mentioned issue related to async interceptors

PR #923, opened 2020-08-31
improves DuplexConnection api and reworks Resumability

OlegDokuka:
This PR enhances DuplexConnection API to unify the way how frames should be sent over it


rsocket-core/src/main/java/io/rsocket/internal/UnboundedProcessor.java
```diff
@@ -40,6 +40,7 @@
  *
  * @param <T> the input and output type
  */
+@Deprecated
```
rstoyanchev, line 43: This could be made package private right away. It's an internal class anyway. However see related comments on `BaseDuplexConnection` as well.

rsocket-core/src/main/java/io/rsocket/internal/BaseDuplexConnection.java
```diff
@@ -1,16 +1,29 @@
...
 import io.rsocket.DuplexConnection;
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.MonoProcessor;
 
 public abstract class BaseDuplexConnection implements DuplexConnection {
-  private MonoProcessor<Void> onClose = MonoProcessor.create();
+  protected MonoProcessor<Void> onClose = MonoProcessor.create();
+
+  @SuppressWarnings("deprecation")
+  protected UnboundedProcessor<ByteBuf> sender = new UnboundedProcessor<>();
```
rstoyanchev, line 12: To make `UnboundedProcessor` completely internal, this should be a private field as well. Perhaps expose a `Publisher<ByteBuf>` protected method for sub-classes to use?

rsocket-core/src/main/java/io/rsocket/DuplexConnection.java
```diff
@@ -19,40 +19,21 @@
...
-import reactor.core.publisher.Mono;
 
 /** Represents a connection with input/output that the protocol uses. */
 public interface DuplexConnection extends Availability, Closeable {
 
   /**
-   * Sends the source of Frames on this connection and returns the {@code Publisher} representing
-   * the result of this send.
+   * Delivers the given frame to the underlying transport connection. This method is non-blocking
+   * and can be safely executed from multiple threads.
```
rstoyanchev, line 30: This mentions non-blocking and thread safety. It would be useful to mention backpressure as well.

rsocket-core/src/main/java/io/rsocket/DuplexConnection.java
```diff
@@ -79,6 +60,12 @@
    */
   Flux<ByteBuf> receive();
 
+  /**
+   * @param errorFrame
+   * @return
+   */
+  void terminate(ByteBuf frame, RSocketErrorException terminalError);
```
rstoyanchev, line 67: It looks like the frame is always `ErrorFrameCodec.encode(duplexConnection.alloc(), 0, exception)`. Maybe this could become a one-argument method `terminate(RSocketErrorException)`.

rstoyanchev (changes requested):

rsocket-core/src/main/java/io/rsocket/DuplexConnection.java
```diff
@@ -19,40 +19,21 @@
...
-   * @return {@code Publisher} that completes when the frame is written on the connection
-   *     successfully and errors when it fails.
+   * @param streamId to which the given frame relates
+   * @param frame with the encoded content
+   * @param prioritize whether the given frame should be sent with priority to others
    */
-  default Mono<Void> sendOne(ByteBuf frame) {
-    return send(Mono.just(frame));
-  }
+  void sendFrame(int streamId, ByteBuf frame, boolean prioritize);
```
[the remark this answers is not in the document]
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/internal/UnboundedProcessor.java
OlegDokuka: Figured out that this Processor is needed in a few more places, thus removed this annotation for now and kept this impl public

rsocket-core/src/main/java/io/rsocket/DuplexConnection.java
OlegDokuka: Removed Exception at all and left ByteBuf frame only

rsocket-core/src/main/java/io/rsocket/DuplexConnection.java
OlegDokuka: fixed

OlegDokuka:
@rstoyanchev finalized API and reworked Resumability implementation

rsocket-core/src/main/java/io/rsocket/core/ClientServerInputMultiplexer.java
```diff
@@ -49,67 +46,40 @@
...
-    return source;
-  }
-
-  public DuplexConnection asServerConnection() {
+  DuplexConnection asResponderConnection() {
     return serverConnection;
   }
 
-  public DuplexConnection asClientConnection() {
+  DuplexConnection asRequesterConnection() {
```
rstoyanchev, line 79: Really minor but the terminology `requester` and `responder` doesn't align with the rest of this class which is `client` and `server`, including the class name. `DuplexConnectionInterceptor.Type` also uses `client` and `server`, and so does the spec when describing `client` and `server` stream id's. I believe that is what they are, and this is why they flip between `RSocketConnector` and `RSocketServer` where one creates the requester `asRequesterConnection()` while the other with `asResponderConnection()`. 

I think `asClientConnection()` and `asServerConnection` are correct.

rsocket-core/src/main/java/io/rsocket/plugins/DuplexConnectionInterceptor.java
```diff
@@ -27,6 +27,8 @@
     extends BiFunction<DuplexConnectionInterceptor.Type, DuplexConnection, DuplexConnection> {
 
   enum Type {
+    /** @deprecated since 1.1.0-M2. Will be removed in 1.2 */
+    @Deprecated
```
rstoyanchev, line 31:
```suggestion
    /**
     * @deprecated as of 1.1 this type is no longer available for interception and will be removed
     *     in 1.2. As an alternative, intercept via {@link Type#SOURCE} and look for frames of type
     *     {@code SETUP}, {@code RESUME}, or {@code RESUME_OK}.
     */
    @Deprecated
```

rstoyanchev (changes requested):
This is a partial review with some feedback identified so far.

PR #924, opened 2020-09-01
Refactors the input for LoadbalanceRSocketClient

rstoyanchev:
The current input is a `Publisher` of a list of `Mono<RSocket>` each of which is created with `RSocketConnector`. However there are some options that should not vary (e.g. metadata type, datatype, etc) since loadbalancing should be transparent.

This change switches the input to a single `RSocketConnector` to use as a template along with a list of `ClientTransport` that can be used to produce the `Mono<RSocket>` for each target. This should provide more guidance and should also be a little easier.


rsocket-core/src/main/java/io/rsocket/loadbalance/LoadbalanceTarget.java
```diff
@@ -0,0 +1,76 @@
...
+ * list of target to use. The transport is used to connect to the target.
+ *
+ * @since 1.1
+ */
+public class LoadbalanceTarget {
+
+  final String key;
+  final ClientTransport transport;
+
+  private LoadbalanceTarget(String key, ClientTransport transport) {
```
spencergibb, line 32: The routing broker won't have a `ClientTransport` but an `RSocket`.

rstoyanchev: IIRC the broker only depends on `LoadBalanceStrategy` or do I have that wrong?

spencergibb: There hasn't been integration yet. So maybe that's all we'll use just some discussion about. If that is the case, then this won't be a problem. Maybe @OlegDokuka remembers. I can definitly see using this on the routing client side though.

OlegDokuka: We do not need transport for broker. Only  strategy

spencergibb: thanks for the remindert

PR #925, opened 2020-09-02
RSocketClient refactoring

rstoyanchev:
`RSocketClient` is not yet in production so we've an opportunity to revisit things like the packaging, Javadoc, as well as where and how the client can be created. 

One high level goal of this change is to draw a more clear contrast between `RSocketClient` and `RSocket`. There is a major update to the Javadoc of `RSocketClient` related to that as well precising the language to reflect the addition of `LoadbalanceRSocketClient` as a second alternative implementation since it was first created.

Secondly to emphasize them as as two distinct layers, in the proposed changes `RSocketConnector` is now left primarily for creating `Mono<RSocket>`, i.e. connecting to establish a live, transient connection. Separately `RSocketClient.from(Mono<RSocket)` allows wrapping a `Mono<RSocket>` source and uses it to establish (and re-establish) a shared connection for multiplexed use as needed.

The change also moves `RSocketClient` from the top-level package, where `RSocket` is for example, to the `io.rsocket.core` sub-package. For once `RSocketClient` is not on the same level as `RSocket`. It's only for sending, it works on a higher level than RSocket, etc. On a practical level this makes it possible to have static factory methods for the default implementation and for adapting a live `RSocket`.


rsocket-core/src/main/java/io/rsocket/core/RSocketClient.java
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * <p>The below configures retry logic to use when a shared {@code RSocket} connection is obtained:
+ *
+ * <pre class="code">{@code
+ * Mono<RSocket> source =
+ *         RSocketConnector.create()
+ *                 .metadataMimeType("message/x.rsocket.composite-metadata.v0")
+ *                 .dataMimeType("application/cbor")
+ *                 .reconnect(Retry.fixedDelay(3, Duration.ofSeconds(1)))
+ *                 .toRSocketClient(TcpClientTransport.create("localhost", 7000));
```
spencergibb, line 68: `toRSocketClient` should probably be something different.

rstoyanchev: indeed

OlegDokuka (approved):
LGTM

Issue #743, opened 2020-02-16
Whitelist client IP address


kacper-emersoft:
As per title, is there any way to whitelist client IP address?

spencergibb:
Whitelist typically means always allow, is there something blocking your ip address. Can you explain what problem you are experiencing and why you believe a whitelist is needed?

kacper-emersoft:
I have created a rsocket server and few clients. I want to restrict access to my rsocket server that only clients with given IP address should be able to connect to my server. . Is it possible to achive that on rsocket server level?

OlegDokuka:
I think rsocket has no relation to that. Usually, such a configuration happens on the infrastructure level. 

Are there any reasons to put it on the rsocket level?

kacper-emersoft:
I agree that infrastructure level is a good place to handle this. I was just wondering if such feature can be obtained on setup phase. If no. I'm fine with it. Thanks!

OlegDokuka:
Theoretically, your client can include its IP address, however, in many cases, it is impossible to do. 
You can imagine:

* A Docker container calling another docker container. 
* A client calling another over a local, in-memory transport (so, technically there is no any opened port at all)
* A client is connected to a server over a transport that does not support IP-transmitting 

So, in general, there are many limitations and if such logic is a requirement i is better to implement some handshake on your own.

Closing this issue.

Regards,
Oleh

linux-china:
@kacper-emersoft   you can build one interceptor  to verify whitelist clent IP.   https://github.com/rsocket/rsocket-java/issues/735

kacper-emersoft:
@linux-china Thanks for great tip! Two questions regarding implementation:
1. As far as I can see, connection field is private. Is there any other way than access it via reflection? 
2. How to react in my DuplexConnectionInterceptor in case of IP missmatch? Throw an exception?

linux-china:
>> As far as I can see, connection field is private. Is there any other way than access it via reflection?
you can use reflection reflection to access connection field

 >> How to react in my DuplexConnectionInterceptor in case of IP missmatch? Throw an exception?
Yes. you can send an error with REJECTED_SETUP to requester.  

```
REJECTED_SETUP | 0x00000003 | The server rejected the setup, it can specify the reason in the payload. Stream ID MUST be 0.
```

The example code as following(not tested):

```
.addConnectionPlugin(new DuplexConnectionInterceptor() {
                    @Override
[...]

Issue #735, opened 2020-01-16
Ability to intercept requests and access channel information such as remote address


NizGh:
As in webflux we can retrieve the remote address from serverWebExchange.getRequest().getRemoteAddress(), what are alternatives to extract connection/channel information in rsocket?

linux-china:
Now  you can get the remote address by tricky way. 

* TcpDuplexConnection has a  connection attribute with reactor.netty.Connection type, and you can use   connection.channel().remoteAddress() to get remote address
* Write a DuplexConnectionInterceptor plugin on server to intercept DuplexConnection,  if  duplexConnection is instance of WebsocketDuplexConnection or TcpDuplexConnection, and you can get Netty connection attribute.

PR #929, opened 2020-09-10
Expose remote address via DuplexConnection

rstoyanchev:
Pull request for #735 and #743.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/DuplexConnection.java
+++ b/rsocket-core/src/main/java/io/rsocket/DuplexConnection.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2018 the original author or authors.
+ * Copyright 2015-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package io.rsocket;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
+import java.net.SocketAddress;
 import java.nio.channels.ClosedChannelException;
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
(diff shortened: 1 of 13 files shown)


rsocket-core/src/test/java/io/rsocket/test/util/LocalSocketAddress.java
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.rsocket.test.util;
+
+import java.net.SocketAddress;
+import java.util.Objects;
+
+public final class LocalSocketAddress extends SocketAddress {
```
yschimke, line 21: Suggest putting Test in the name of the class maybe toString to make it non-confusing with https://github.com/rsocket/rsocket-java/blob/d903e9635a159285b6943ea93156c31aa406ba5d/rsocket-transport-local/src/main/java/io/rsocket/transport/local/LocalSocketAddress.java

rsocket-core/src/main/java/io/rsocket/DuplexConnection.java
```diff
@@ -86,6 +87,17 @@
    */
   ByteBufAllocator alloc();
 
+  /**
+   * Return the remote address that this connection is connected to. The returned {@link
+   * SocketAddress} varies by transport type and should be downcast to obtain more detailed
+   * information. For TCP and WebSocket, the address type is {@link java.net.InetSocketAddress}. For
+   * local transport, it is {@link io.rsocket.transport.local.LocalSocketAddress}.
```
yschimke, line 94: Can we clarify here that it is null for disconnected connections? Do we want to mark as @Nullable?

rstoyanchev: I think for disconnected, the channel still returns the address. It can return null before it is connected is my understanding but that the `DuplexConnection` is created only after a connection is ready.

OlegDokuka: Yup. I think soo. I guess once the connection is resolved, the address should be available.

OlegDokuka (approved):
LGTM

PR #931, opened 2020-09-15
migrates from Reactor UnicastProcessor to new Sinks API

JWood48:


OlegDokuka:
@JWood48 thank you for PR. LGTM. But, can you please apply a googleJavaFormat task to make the build passing?

rsocket-examples/src/main/java/io/rsocket/examples/transport/tcp/fnf/TaskProcessingWithServerSideNotificationsExample.java
```diff
@@ -213,7 +213,7 @@ public RSocketTaskHandler(
       @Override
       public Mono<Void> fireAndForget(Payload payload) {
         logger.info("Received a Task[{}] from Client.ID[{}]", payload.getDataUtf8(), id);
-        tasksToProcess.onNext(new Task(id, payload.getDataUtf8()));
+        tasksToProcess.tryEmitNext(new Task(id, payload.getDataUtf8()));
```
simonbasle, line 216: the `tryEmitNext` method provides immediate feedback and could reject the emission. the return value should be checked, at a minimum, for `isFailed()` status.

simonbasle (dismissed):

rsocket-examples/src/main/java/io/rsocket/examples/transport/tcp/fnf/TaskProcessingWithServerSideNotificationsExample.java
```diff
@@ -213,9 +213,11 @@ public RSocketTaskHandler(
       @Override
       public Mono<Void> fireAndForget(Payload payload) {
         logger.info("Received a Task[{}] from Client.ID[{}]", payload.getDataUtf8(), id);
-        tasksToProcess.onNext(new Task(id, payload.getDataUtf8()));
+        Sinks.Emission emission = tasksToProcess.tryEmitNext(new Task(id, payload.getDataUtf8()));
```
bsideup, line 216: FYI some emissions (like the non-serialized access) can be retried (although make sure to always check whether it still worth retrying)

simonbasle (commented):
I'll defer to @OlegDokuka see if that example should include more complex handling of `Emission` error codes, but otherwise looks good.

Issue #887, opened 2020-07-07
Server memory overflow due to stream receivers retain payloads regardless requestN


mostroverkhov:
Rogue clients may not respect requestN from server, send frames firehose which are queued by receivers -  [UnicastProcessors](https://github.com/rsocket/rsocket-java/blob/2dbf61d2b214185a492bba45725edce9000e8df4/rsocket-core/src/main/java/io/rsocket/core/RSocketResponder.java#L540), backed by unbounded [queue](https://github.com/reactor/reactor-core/blob/232fe2e32bd6a99327b0f0f7b4efb4ca84f8e0e6/reactor-core/src/main/java/reactor/core/publisher/UnicastProcessor.java#L100).

Affects server `Responder` channel; also `Requester` stream & channel - if server sends requests to client.

Reproducible with test below, It roughly models system that processes batches of messages asynchronously, calls new batch with requestN once current one is completed

```java
  @Test
  void serverOom() {
    EmitterProcessor<Void> completed = EmitterProcessor.create();

    TestServerTransport testTransport = new TestServerTransport();
    Closeable server =
        RSocketServer.create()
            .acceptor(
                (setup, sendingSocket) ->
                    Mono.just(
                        new RSocket() {
                          @Override
                          public Flux<Payload> requestChannel(Publisher<Payload> payloads) {
                            Flux.from(payloads)
                                .doFinally(s -> completed.onComplete())
                                .subscribe(new BatchProcessingSubscriber());
                            return completed.cast(Payload.class);
                          }
                        }))
            .bind(testTransport)
            .block();

    byte[] bytes = bytes(1_000_000);
    TestDuplexConnection testDuplexConnection = testTransport.connect();

    /*setup*/
    testDuplexConnection.addToReceivedBuffer(
        SetupFrameCodec.encode(
            ByteBufAllocator.DEFAULT,
            false,
            1111111,
            1111111,
            Unpooled.EMPTY_BUFFER,
            "application/binary",
            "application/binary",
            EmptyPayload.INSTANCE));

    /*request-channel which sends many frames regardless received requestN*/
    testDuplexConnection.addToReceivedBuffer(
        RequestChannelFrameCodec.encode(
            ByteBufAllocator.DEFAULT,
            1,
            false,
            false,
            1111111,
            Unpooled.EMPTY_BUFFER,
            Unpooled.EMPTY_BUFFER));

    int framesCount = 8000;
    for (int i = 0; i < framesCount; i++) {
[...]

PR #932, opened 2020-09-22
Avoid queueing in UnicastProcessor receivers

rstoyanchev:
Pull request for #887 to disallow unbounded input.

### DIFF
--- a/rsocket-core/src/test/java/io/rsocket/core/RSocketTest.java
+++ b/rsocket-core/src/test/java/io/rsocket/core/RSocketTest.java
@@ -16,8 +16,6 @@
 
 package io.rsocket.core;
 
-import static io.rsocket.frame.FrameLengthCodec.FRAME_LENGTH_MASK;
-
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.rsocket.Payload;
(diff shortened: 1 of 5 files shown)


rsocket-core/src/test/java/io/rsocket/core/RSocketTest.java
```diff
@@ -556,6 +558,7 @@ protected void init() {
                       .subscribe();
 
                   return Flux.range(1, 10)
+                      .subscribeOn(Schedulers.boundedElastic())
```
rstoyanchev, line 561: @OlegDokuka the purpose of these is to decouple the responding since otherwise, with local transports, the responder messages get sent too early before the requester has fully set up the subscriber side.

OlegDokuka: Yeah. That is fine. Actually it has to bee fixed in Reactor, but I postponed that. Thus, once I have it fixed, we can remove it

rsocket-core/src/test/java/io/rsocket/core/RSocketTest.java
```diff
@@ -543,6 +543,7 @@ protected void init() {
                 @Override
                 public Flux<Payload> requestStream(Payload payload) {
                   return Flux.range(1, 10)
+                      .delaySubscription(Duration.ofMillis(100))
```
yschimke, line 546: If these are pending a fix in reactor (IIUC), can we comment on these with a TODO?

OlegDokuka: I send to @rstoyanchev a small patch which should eliminate the reason for adding this delay. Thus once it is going to be applied, all these will be removed as well

Issue #865, opened 2020-06-12
Metadata-pushes rate is not capped on responder, may lead to server resources starvation


mostroverkhov:
External clients can start hundreds of thousands of metadata-pushes over multiple connections.

If metadata-push handler is present and It does not implement rate limiting on its own, then server is subject to resource starvation attack.

If metadata-push handler is absent, then [exception with stacktrace](https://github.com/rsocket/rsocket-java/blob/9e12fb35a906e6d6fffeaa9197d2d8138f11031b/rsocket-core/src/main/java/io/rsocket/RSocket.java#L83) is created for every metadata-push - this just consumes cpu cycles. With sufficiently large number of metadata-pushes RSocket becomes unresponsive.

Problem worsens with websocket transport of  this repository. By default It [sets](https://github.com/rsocket/rsocket-java/blob/9e12fb35a906e6d6fffeaa9197d2d8138f11031b/rsocket-transport-netty/src/main/java/io/rsocket/transport/netty/server/WebsocketServerTransport.java#L54) frame size limit to  ~16MB - this opens opportunity for memory overflow given there may be arbitrarily large number of concurrent metadata-pushes with size close to said size limit.

OlegDokuka:
Hey!

First of all, **we appreciate issues and bug reports**! 

**However**, sometimes it is **_HARD_** to figure out what the issue is about. In certain: **how to reproduce it**; **what is the root cause**; **what is possible solutions to mitigate it**!  

Having no such info just makes it **challenging** for maintainers to help submitter / improve project by resolving the issue.
 
_**Therefore**_, we created issues' templates for that purpose so the issue submitter can express all important points that maintainers need in order to resolve an issue. Otherwise, if thoughts were not expressed clearly just make the issue useless and waste maintainers' time on figuring out what was said. 

**Therefore, please respect maintainers if you submitting an issue!**

Thanks in advance.

-----

1. Any code samples to reproduce that? 

2. I believe the same issue can happen with fnf call. Is not it?

3. **_The issue is not critical and a workaround for everything lower than 1.0.2  will be overriding of all RSocket methods with a static response error_**

4. The solution for this issue will be creating static Mono.error/Flux.error for `RSocket.class` so stack trace will not be filled on every call.

5.
> Problem worsens with websocket transport of this repository. By default It sets frame size limit to ~16MB - this opens opportunity for memory overflow given there may be arbitrarily large number of concurrent metadata-pushes with size close to said size limit.

[...]

PR #933, opened 2020-09-22
Use static errors in RSocket default method implementations

rstoyanchev:
Proposed changes for #865.

I think limiting the rate of any individual request type is up to RSocket responder implementations. By default we can at least ensure minimal processing overhead.


spencergibb (approved):

rsocket-core/src/main/java/io/rsocket/RSocketAdapter.java
```diff
@@ -0,0 +1,67 @@
...
+import reactor.core.publisher.Mono;
+
+/**
+ * Package private class with default implementations for use in {@link RSocket}. The main purpose
+ * is to hide static {@link UnsupportedOperationException} declarations.
+ */
+class RSocketAdapter {
+
+  private static final Mono<Void> UNSUPPORTED_FIRE_AND_FORGET =
+      Mono.error(new UnsupportedOperationException("Fire-and-Forget not implemented."));
```
robotmrv, line 30: @rstoyanchev  
JFY  
There were several issues related to static exceptions.  
For example https://github.com/reactor/reactor-core/issues/1710 and https://github.com/reactor/reactor-core/issues/1872

rstoyanchev: Yes I am aware, thanks for the reminder @rsocket/rsocket-rb and pointers. I've taken a similar approach as Netty, turning off suppressed exceptions. I think that's a perfectly legitimate trade-off, in this case there isn't much use for the Reactor debugging feature that relies on suppressed exceptions.

OlegDokuka (approved):
LGTM

PR #934, opened 2020-09-27
ensures that resumability implementation is stable

OlegDokuka:
This PR ensures that resumability implementation is stable and does not have any mem leaks in it

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/resume/ServerRSocketSession.java
+++ b/rsocket-core/src/main/java/io/rsocket/resume/ServerRSocketSession.java
@@ -48,6 +48,11 @@ public class ServerRSocketSession
   final ByteBufAllocator allocator;
   final boolean cleanupStoreOnKeepAlive;
 
+  /**
+   * All incoming connections with the Resume intent are enqueued in this queue. Such an approach
+   * ensure that the new connection will affect the resumption state anyhow until the previous
+   * (active) connection is finally closed
+   */
   final Queue<Runnable> connectionsQueue;
 
   volatile int wip;
--- a/rsocket-core/src/main/java/io/rsocket/resume/ServerRSocketSession.java
+++ b/rsocket-core/src/main/java/io/rsocket/resume/ServerRSocketSession.java
@@ -26,6 +26,8 @@ import io.rsocket.frame.ResumeFrameCodec;
 import io.rsocket.frame.ResumeOkFrameCodec;
 import io.rsocket.keepalive.KeepAliveSupport;
 import java.time.Duration;
+import java.util.Queue;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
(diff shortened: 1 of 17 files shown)


rsocket-core/src/main/java/io/rsocket/resume/ServerRSocketSession.java
```diff
@@ -45,6 +48,12 @@
   final ByteBufAllocator allocator;
   final boolean cleanupStoreOnKeepAlive;
 
+  final Queue<Runnable> connectionsQueue;
```
rstoyanchev, line 51: This looks like it is meant to allow multiple concurrent attempts to resume but I'm not sure. It would be useful to clarify the case in an implementation comment.

rsocket-test/src/main/java/io/rsocket/test/LeaksTrackingByteBufAllocator.java
```diff
@@ -201,4 +223,60 @@ public int calculateNewCapacity(int minNewCapacity, int maxCapacity) {
...
+  @SuppressWarnings("unchecked")
+  static Set<Object> resolveTrackingInfo(ByteBuf byteBuf) throws Exception {
+    if (ResourceLeakDetector.getLevel().ordinal()
+        >= ResourceLeakDetector.Level.ADVANCED.ordinal()) {
+      if (simpleLeakAwareCompositeByteBufClass.isInstance(byteBuf)) {
+        return (Set<Object>) allLeaksField.get(leakFieldForComposite.get(byteBuf));
+      } else if (simpleLeakAwareByteBufClass.isInstance(byteBuf)) {
+        return (Set<Object>) allLeaksField.get(leakFieldForNormal.get(byteBuf));
+      }
+    }
```
rstoyanchev, line 278: Maybe we should ask Netty if they would consider making it easier to obtain a leak report. Or alternatively, probably better yet, to provide support for tracking leaks in tests.

rstoyanchev (changes requested):

rsocket-core/src/main/java/io/rsocket/resume/ServerRSocketSession.java
OlegDokuka: added docs

rsocket-test/src/main/java/io/rsocket/test/LeaksTrackingByteBufAllocator.java
OlegDokuka: I guess I can port such a tool to Netty, so possibly they accept it eventually and we will use it directly from there

rstoyanchev: To be clear I'm not suggesting that it should be a PR to start. For now it would be valuable enough to raise the idea by creating a well defined issue, also perhaps describing own experience.

PR #939, opened 2020-10-03
Update URL of RSocket Demo Client in Readme

jexp:
The URL has changed, see: https://rsocket-demo.herokuapp.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ import java.net.URI;
 
 public class ExampleClient {
     public static void main(String[] args) {
-        WebsocketClientTransport ws = WebsocketClientTransport.create(URI.create("ws://rsocket-demo.herokuapp.com/ws"));
+        WebsocketClientTransport ws = WebsocketClientTransport.create(URI.create("wss://rsocket-demo.herokuapp.com/rsocket"));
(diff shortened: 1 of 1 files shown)


yschimke (approved):

jexp:
Also added the M2 and the repository spring-milestone for reactor to the readme

fixes https://github.com/rsocket/rsocket-java/issues/940

README.md
```diff
@@ -40,6 +42,7 @@ Example:
 ```groovy
 repositories {
 	maven { url 'https://oss.jfrog.org/oss-snapshot-local' }
+	maven { url 'https://repo.spring.io/milestone' }
```
OlegDokuka, line 45:
```suggestion
	maven { url 'https://repo.spring.io/snapshot' }
```

We usually depend on the snapshots or reactor, thus if such occasion happens, it is better to specify a snapshot repo

OlegDokuka: Also, this does not fix #940 since this section is related to the snapshots, but M1 now specified as a release one, so I would suggest adding a similar section for release and keep it until we have a GA for 1.1.0

rstoyanchev:
This should be addressed now. Thanks for the PR.

Issue #914, opened 2020-08-11
NullPointerException on connection disposal


SerCeMan:
Hey, folks!

We've noticed the following error periodically when the connections are being disposed. If I'm reading the exception correctly, it seems like the collection stores `null` at some point.
```
java.lang.NullPointerException: null
    at java.lang.Iterable.forEach(Iterable.java:75)
    at io.rsocket.core.RSocketResponder.cleanUpSendingSubscriptions(RSocketResponder.java:274)
    at io.rsocket.core.RSocketResponder.cleanup(RSocketResponder.java:264)
    at io.rsocket.core.RSocketResponder.tryTerminate(RSocketResponder.java:166)
    at io.rsocket.core.RSocketResponder.tryTerminateOnConnectionClose(RSocketResponder.java:159)
    at reactor.core.publisher.LambdaMonoSubscriber.onComplete(LambdaMonoSubscriber.java:132)
    at reactor.core.publisher.MonoProcessor$NextInner.onComplete(MonoProcessor.java:518)
    at reactor.core.publisher.MonoProcessor.onNext(MonoProcessor.java:308)
    at reactor.core.publisher.MonoProcessor.onComplete(MonoProcessor.java:265)
    at io.rsocket.internal.BaseDuplexConnection.dispose(BaseDuplexConnection.java:23)
    at ...
```

## Steps to Reproduce

It seems that the error isn't easily reproducible. We're only seeing very rarely, once or twice a day. 

## Your Environment

* RSocket version(s) used: 1.0.1
* Other relevant libraries versions (eg. `netty`, ...): none
* Platform (eg. JVM version (`javar -version`) or Node version (`node --version`)): 13
* OS and version (eg `uname -a`): ubuntu 20.04

OlegDokuka:
Hey @SerCeMan!

Thank you for reporting that!
I will have a look. 

On the root-cause... Not sure we put null somewhere, it is more likely it is racing between cleanup during disposal and a stream which does selfremoval when it is terminated as well. So it may be that iterator has seen an element but a collection has already been shrunken so once a next is called, the element being observed is null, hence we see NPE.

rstoyanchev:
This should be resolved now.

Cesarla:
Hi @OlegDokuka, @rstoyanchev  can these improvements be back-ported to rsocket 1.1.X?

OlegDokuka:
@Cesarla any reproducer to share?

Cesarla:
[...]

PR #941, opened 2020-10-06
Safer iteration when cancelling subscriptions

rstoyanchev:
Indeed as suggested in https://github.com/rsocket/rsocket-java/issues/914#issuecomment-671742260, since the removal of an individual sending subscription is not synchronized, when iterating over all subscription we can run into null values. This change iterates explicitly and continues despite errors on a best case effort.


OlegDokuka (approved):
LGTM. Need to be ported to the master branch as well

PR #944, opened 2020-10-12
provides request intercepting api

OlegDokuka:
This PR provides a `RequestInterceptor` interface API and an appropriate implementation in every RequestOperator implementation.

This interceptor is a building block to build Stats/Metrics tracking without providing a wrapper for every call.


rsocket-core/src/main/java/io/rsocket/plugins/InterceptorRegistry.java
```diff
@@ -30,16 +31,54 @@
...
+
+  /**
+   * Variant of {@link #forResponderRequests(Supplier)} with access to the list of existing
+   * registrations.
+   */
+  public InterceptorRegistry forResponderRequests(
+      Consumer<List<Supplier<? extends RequestInterceptor>>> consumer) {
+    consumer.accept(responderRequestInterceptors);
+    return this;
+  }
```
rstoyanchev, line 75: Since the naming convention is not so much about the type of interceptor but about the what the interceptor is installed on, perhaps we can go with overloaded methods, e.g. `forRequester(Supplier<? extends RequestInterceptor>)`. We could also probably avoid the variants with `Consumer<List<Supplier>>` for now until we know it really is needed?

rsocket-core/src/main/java/io/rsocket/plugins/InterceptorRegistry.java
```diff
@@ -30,16 +31,54 @@
...
+  private List<Supplier<? extends RequestInterceptor>> requesterRequestInterceptors =
+      new ArrayList<>();
+  private List<Supplier<? extends RequestInterceptor>> responderRequestInterceptors =
+      new ArrayList<>();
+  private List<RSocketInterceptor> requesterRSocketInterceptors = new ArrayList<>();
+  private List<RSocketInterceptor> responderRSocketInterceptors = new ArrayList<>();
   private List<SocketAcceptorInterceptor> socketAcceptorInterceptors = new ArrayList<>();
   private List<DuplexConnectionInterceptor> connectionInterceptors = new ArrayList<>();
 
+  /** Add an {@link RequestInterceptor} that will hook into Requester RSocket requests' phases. */
```
rstoyanchev, line 43: It would be mention in the Javadoc to explain that unlike an `RSocketInterceptor` which is a singleton, here a `RequestInterceptor` is per `RSocket` and the `Supplier` is used to create a new instance on each new connection. 

The same comment applies to the other registration methods as well and I think a similar comment can be added in the type level Javadoc of `RequestInterceptor` that it's meant to be used with a separate instance per `RSocket` although technically nothing prevents always supplying the same instance too if you want a singleton.

Another way to achieve this would be to take a `Function<RSocket, ? extends RequestInterceptor>` which would make it more clear and also allow potentially a singleton to be used.

rsocket-core/src/main/java/io/rsocket/core/FireAndForgetRequesterMono.java
```diff
@@ -98,21 +103,41 @@ public void subscribe(CoreSubscriber<? super Void> actual) {
       return;
     }
 
+    final RequestInterceptor interceptor = this.requestInterceptor;
+    if (interceptor != null) {
+      interceptor.onStart(streamId, FrameType.REQUEST_FNF, p.sliceMetadata());
```
rstoyanchev, line 108: Sometimes (many times?) an interceptor might not care about the metadata. I'm wondering if we can avoid the slice in that case? We could for example pass the request `Payload` and then the interceptor can access the metadata from that if it needs it.

rsocket-core/src/main/java/io/rsocket/plugins/RequestInterceptor.java
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Method which is being invoked on the request rejection.
+   *
+   * @param streamId used for the request
+   * @param requestType of the request. Must be one of the following types {@link
+   *     FrameType#REQUEST_FNF}, {@link FrameType#REQUEST_RESPONSE}, {@link
+   *     FrameType#REQUEST_STREAM} or {@link FrameType#REQUEST_CHANNEL}
+   * @param metadata provided in the request frame
+   */
+  void onReject(int streamId, FrameType requestType, @Nullable ByteBuf metadata);
```
rstoyanchev, line 41: Javadoc here could elaborate a little more that this is a rejection from the lease handler if it has run out of leases. That means this happens before the request is even sent so it's either `onStart` + `onEnd` or just `onReject`.

rsocket-core/src/main/java/io/rsocket/plugins/SafeCompositeRequestInterceptor.java
```diff
@@ -0,0 +1,64 @@
+package io.rsocket.plugins;
+
+import io.netty.buffer.ByteBuf;
+import io.rsocket.frame.FrameType;
+import reactor.core.publisher.Operators;
+import reactor.core.publisher.SignalType;
+import reactor.util.context.Context;
+
+public class SafeCompositeRequestInterceptor implements RequestInterceptor {
```
rstoyanchev, line 9: Perhaps just `CompositeRequestInterceptor` as the name? I'm not sure it needs to have `Safe` in the name and there are no other (unsafe) implementation.

rstoyanchev (changes requested):
Looks good overall, some comments below.

rsocket-core/src/main/java/io/rsocket/plugins/InitializingInterceptorRegistry.java
```diff
@@ -18,13 +18,40 @@
...
+      List<Supplier<? extends RequestInterceptor>> interceptors) {
+    switch (interceptors.size()) {
+      case 0:
+        return null;
+      case 1:
+        return new SafeRequestInterceptor(interceptors.get(0).get());
+      default:
+        return new SafeCompositeRequestInterceptor(
+            interceptors.stream().map(Supplier::get).toArray(RequestInterceptor[]::new));
+    }
```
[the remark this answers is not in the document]
OlegDokuka: checking for null is better than doing a dynamic invocation, though it should be eliminated by JIT if we are lucky

rsocket-core/src/main/java/io/rsocket/plugins/InterceptorRegistry.java
OlegDokuka: It clashes with the existing one, we can not have Consumer of List of RSocket as well as Consumer of List of RequestInterceptors

rsocket-core/src/main/java/io/rsocket/core/FireAndForgetRequesterMono.java
OlegDokuka: We can not do that when it comes to reassembling fragmented payload. Thus, the best is Bytebuf for metadata only. Also, if the fragment is rejected because of leasing, we will never have the full payload

OlegDokuka: I guess we can get rid of metadata, for now, until we see a use case for that

rsocket-core/src/main/java/io/rsocket/plugins/InterceptorRegistry.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/plugins/RequestInterceptor.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/plugins/SafeCompositeRequestInterceptor.java
OlegDokuka: Fixed

rsocket-core/src/main/java/io/rsocket/core/FireAndForgetRequesterMono.java
OlegDokuka: revised. Do `p.metadata()` now

rsocket-core/src/main/java/io/rsocket/plugins/InterceptorRegistry.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/plugins/InitializingInterceptorRegistry.java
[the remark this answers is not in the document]
OlegDokuka: keep null check for now

PR #946, opened 2020-10-22
migrates weighted loadbalance strategy to user new  RequestInterceptors for Stats tracking

OlegDokuka:
This PR refactors current weighted rsocket to another mechanism which makes Stats to be hidden from the outside world and make it accessible only for WeighetedLoadbalanceStrategy.

This PR introduces StatsRequestInterceptor, which is a public class, hence can be used not only for Client-side but as well as for the server one (e.g. broker case).

The WeightedLoadbalanceStrategy, gets a new constructor API which lets one using selected RSocket to extract associated StatsRequestInterceptor and then calculate the weight.


rsocket-core/src/main/java/io/rsocket/loadbalance/StatsRequestInterceptor.java
```diff
@@ -0,0 +1,87 @@
+package io.rsocket.loadbalance;
+
+import io.netty.buffer.ByteBuf;
+import io.rsocket.frame.FrameType;
+import io.rsocket.plugins.RequestInterceptor;
+import reactor.util.annotation.Nullable;
+
+public class StatsRequestInterceptor implements RequestInterceptor {
```
rstoyanchev, line 8: Maybe `WeightedStatsRequestInterceptor` and likewise `Stats` --> `WeightedStats` since those are associated with the `WeightedLoadbalanceStrategy`.

rsocket-core/src/main/java/io/rsocket/loadbalance/LoadbalanceRSocketClient.java
```diff
@@ -179,4 +186,12 @@ private LoadbalanceStrategy initLoadbalanceStrategy() {
...
     }
   }
+
+  /**
+   * Extension for {@link LoadbalanceStrategy} which allows pre-setup {@link RSocketConnector} for
+   * {@link LoadbalanceStrategy} needs
+   */
+  public interface InitializingLoadbalanceStrategy extends LoadbalanceStrategy {
+    void init(RSocketConnector connector);
+  }
```
rstoyanchev, line 196: Let's make this a top-level interface. It has become a bit long otherwise. `ClientLoadbalanceStrategy` as we discussed previously is a good name because it's for use from the client side.

rstoyanchev (commented):
Looks great overall.

One consequence is that `WeightedLoadBalanceStrategy` can no longer be instantiated directly for the purpose of being used with `LoadbalanceRSocketClient`. I think `WeightedLoadbalanceStrategy` should be an `InitializingLoadbalanceStrategy` but the management of that `Map` as well as the logic to initialize the connector could be in an inner class (rather than a sub-class) which implements `Function<RSocket, StatsRequestInterceptor>`. So then if `WeightedLoadbalanceStrategy` is created without an externally provided `Function` it falls back on an instance of that inner class.

spencergibb (approved):

rsocket-core/src/main/java/io/rsocket/loadbalance/StatsRequestInterceptor.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/loadbalance/LoadbalanceRSocketClient.java
OlegDokuka: fixed

PR #948, opened 2020-10-25
Add capnp & flatbuffers serialization support

linux-china:
Cap’n Proto and FlatBuffers  are high performance serialization library, and they should be included in WellKnownType, and especially useful to per stream DataMimeType, and text MimeType is too long for them because of performance.


OlegDokuka (changes requested):
I guess this should be added first into the spec, and then into the certain implementation.

https://github.com/rsocket/rsocket/blob/master/Extensions/WellKnownMimeTypes.md

yschimke:
Can we merge this now? https://github.com/rsocket/rsocket/commits/master/Extensions/WellKnownMimeTypes.md

PR #949, opened 2020-10-26
Update WeightedLoadbalanceStrategy to use a Builder

rstoyanchev:
The use of a `Builder` helps to isolate a couple of optional properties and keeps the potential to expose more of its internal parameters for configuration in the future.


OlegDokuka (approved):
LGTM

PR #950, opened 2020-10-26
Rename RequestInterceptor registration methods

rstoyanchev:
Avoiding the overload with `RSocketInterceptor` registrations helps to avoid ambiguity when using lambdas.


OlegDokuka (approved):
LGTM

Issue #961, opened 2020-11-19
RSocketResponder does not remove Subscriber from sendingSubscriptions when response is empty


hadix-lin:
io.rsocket.core.RSocketResponder#handleRequestResponse  

```java
private void handleRequestResponse(int streamId, Mono<Payload> response) {
    final BaseSubscriber<Payload> subscriber =
        new BaseSubscriber<Payload>() {
          private boolean isEmpty = true;

          @Override
          protected void hookOnNext(Payload payload) {
            if (isEmpty) {
              isEmpty = false;
            }

            if (!PayloadValidationUtils.isValid(mtu, payload, maxFrameLength)) {
              payload.release();
              cancel();
              final IllegalArgumentException t =
                  new IllegalArgumentException(INVALID_PAYLOAD_ERROR_MESSAGE);
              handleError(streamId, t);
              return;
            }

            ByteBuf byteBuf =
                PayloadFrameCodec.encodeNextCompleteReleasingPayload(allocator, streamId, payload);
            sendProcessor.onNext(byteBuf);
          }

          @Override
          protected void hookOnError(Throwable throwable) {
            if (sendingSubscriptions.remove(streamId, this)) {
              handleError(streamId, throwable);
            }
          }

          @Override
          protected void hookOnComplete() {
            if (isEmpty) {
              if (sendingSubscriptions.remove(streamId, this)) {
                sendProcessor.onNext(PayloadFrameCodec.encodeComplete(allocator, streamId));
              }
            }
          }
        };

    sendingSubscriptions.put(streamId, subscriber); 
    response.doOnDiscard(ReferenceCounted.class, DROPPED_ELEMENTS_CONSUMER).subscribe(subscriber);
  }
```

set isEmpty=false in hookOnNext, but there is no code to set it to true afterwards.   ```sendingSubscriptions.remove(streamId, this)``` in hookOnComplete has no chance to be executed. cause the subscriber object cannot be recycle.

Why not put ```sendingSubscriptions.remove(streamId, this)``` into hookFinally?

linux-china:
Then link is:  https://github.com/rsocket/rsocket-java/blob/d047c59809dd65db86ebd3d76924e692b402c6e5/rsocket-core/src/main/java/io/rsocket/core/RSocketResponder.java#L402

or remove directly  in hookOnComplete like logic for requestStream

```
 protected void hookOnComplete() {
            if (sendingSubscriptions.remove(streamId, this)) {
[...]

PR #962, opened 2020-11-20
Ensure Subscriber is removed from sendingSubscriptions

rstoyanchev:
Fix for #961.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketResponder.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketResponder.java
@@ -406,22 +406,20 @@ class RSocketResponder implements RSocket {
 
           @Override
           protected void hookOnNext(Payload payload) {
-            if (isEmpty) {
-              isEmpty = false;
-            }
+            isEmpty = false;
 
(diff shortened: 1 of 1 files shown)


OlegDokuka (changes requested):
Let's do it a little differently. Let's remove the subscriber right immediately at the `onNext` method. Spec says that in the case of `NEXT_COMPLETE` no frames should be sent.

Thus, we should consider only 3 possible outcomes:
```
onNext -> remove subscriber and send the NEXT_COMPLETE frame
only onComplete -> remove subscriber and send the COMPLETE frame
only onError -> remove subscriber and send the ERROR frame
```

if 

```
onNext + onComplete (do nothing at onComplete)
onNext + onError (drop error at onError)
```

rstoyanchev:
> Let's do it a little differently.

Good idea, I've updated the PR accordingly.

OlegDokuka (approved):
LGTM

OlegDokuka (approved):
LGTM

linux-china:
@OlegDokuka  could you release 1.0.4 to include this bug fix?   it's very important for Spring Boot 2.3.X.

PR #976, opened 2021-01-29
adds proxy RSocket for WeightedStatsRequestInterceptor

spencergibb:
Signed-off-by: Spencer Gibb sgibb@pivotal.io

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/loadbalance/WeightedStats.java
+++ b/rsocket-core/src/main/java/io/rsocket/loadbalance/WeightedStats.java
@@ -1,7 +1,9 @@
 package io.rsocket.loadbalance;
 
+import io.rsocket.RSocket;
+
 /**
- * Representation of stats used by the {@link WeightedLoadbalanceStrategy}
+ * Representation of stats used by the {@link WeightedLoadbalanceStrategy}.
  *
  * @since 1.1
  */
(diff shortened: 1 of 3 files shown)


rstoyanchev (changes requested):
Sorry for the slow response. 

I'm not sure if this needs to be tied to the interceptor. Perhaps something like this would make it more generally useful:
```java
public class WeightedStatsRSocketProxy extends RSocketProxy implements WeightedStats {

	private final WeightedStats stats;


	public WeightedStatsRSocketProxy(RSocket source, WeightedStats stats) {
		super(source);
		this.stats = stats;
	}

	@Override
	public double higherQuantileLatency() {
		return this.stats.higherQuantileLatency();
	}

	@Override
	public double lowerQuantileLatency() {
		return this.stats.lowerQuantileLatency();
	}

	@Override
	public int pending() {
		return this.stats.pending();
	}

	@Override
	public double predictedLatency() {
		return this.stats.predictedLatency();
	}

	@Override
	public double weightedAvailability() {
		return this.stats.weightedAvailability();
	}
}
```

spencergibb:
In conversations with @OlegDokuka, the idea was that the class wouldn't be public.

spencergibb:
@rstoyanchev maybe a package protected class, but with the static creator in the interceptor still?

rstoyanchev:
> maybe a package protected class, but with the static creator in the interceptor still?

Isn't that what's currently proposed? 

How about exposing via default method on `WeightedStats`, something like:
```java
default RSocket wrap(RSocket rsocket) {
    return new WeightedStatsRSocketProxy(rsocket, this); // instantiate package private class
}
```

spencergibb:
Done @rstoyanchev

rstoyanchev (approved):

PR #983, opened 2021-02-18
fixes LoadbalanceTest issues

rstoyanchev:
I think, on occasion, not all requests had finished executing when the counter is checked.

### DIFF
--- a/rsocket-core/src/test/java/io/rsocket/loadbalance/LoadbalanceTest.java
+++ b/rsocket-core/src/test/java/io/rsocket/loadbalance/LoadbalanceTest.java
@@ -12,6 +12,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 import org.reactivestreams.Publisher;
@@ -26,10 +28,18 @@ import reactor.util.context.Context;
 
 public class LoadbalanceTest {
 
-  @Test
-  public void shouldDeliverAllTheRequestsWithRoundRobinStrategy() {
+  @BeforeEach
+  void setUp() {
     Hooks.onErrorDropped((__) -> {});
+  }
 
+  @AfterAll
+  static void afterAll() {
+    Hooks.resetOnErrorDropped();
+  }
+
+  @Test
+  public void shouldDeliverAllTheRequestsWithRoundRobinStrategy() throws Exception {
     final AtomicInteger counter = new AtomicInteger();
     final ClientTransport mockTransport = Mockito.mock(ClientTransport.class);
     final RSocketConnector rSocketConnectorMock = Mockito.mock(RSocketConnector.class);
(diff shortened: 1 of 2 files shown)


rstoyanchev:
It looks like some requests are failing like this:
```
2021-02-18T11:23:19.6328178Z STD_ERR: java.lang.ClassCastException: class io.rsocket.loadbalance.PooledRSocket cannot be cast to class io.rsocket.loadbalance.WeightedStats (io.rsocket.loadbalance.PooledRSocket and io.rsocket.loadbalance.WeightedStats are in unnamed module of loader 'app')
2021-02-18T11:23:19.6332679Z STD_ERR: 	at io.rsocket.loadbalance.LoadbalanceTest.lambda$shouldDeliverAllTheRequestsWithWeightedStrategy$7(LoadbalanceTest.java:109)
2021-02-18T11:23:19.6336141Z STD_ERR: 	at io.rsocket.loadbalance.WeightedLoadbalanceStrategy.select(WeightedLoadbalanceStrategy.java:72)
2021-02-18T11:23:19.6338775Z STD_ERR: 	at io.rsocket.loadbalance.RSocketPool.doSelect(RSocketPool.java:217)
```

The test sets up a `weightedStatsResolver` that simply downcasts the `RSocket` which is `PooledRSocket`. So when there are two loadbalance targets, and `RSocketPool` needs to make a choice, it fails.

rstoyanchev:
I've pushed an update that should address the issue.

rsocket-core/src/test/java/io/rsocket/loadbalance/LoadbalanceTest.java
```diff
@@ -26,10 +29,13 @@
 
 public class LoadbalanceTest {
 
-  @Test
-  public void shouldDeliverAllTheRequestsWithRoundRobinStrategy() {
+  @BeforeEach
+  void setUp() {
     Hooks.onErrorDropped((__) -> {});
```
OlegDokuka, line 34: I guess I forgot to do `Hooks.resetOnErrorDropped()` can you please add that one in the `@AfterEach`

rsocket-core/src/test/java/io/rsocket/loadbalance/LoadbalanceTest.java
```diff
@@ -51,12 +57,14 @@ public void shouldDeliverAllTheRequestsWithRoundRobinStrategy() {
...
       final RSocketPool rSocketPool =
           new RSocketPool(rSocketConnectorMock, source, new RoundRobinLoadbalanceStrategy());
+      final CountDownLatch requestLatch = new CountDownLatch(1000);
 
       RaceTestUtils.race(
           () -> {
             for (int j = 0; j < 1000; j++) {
               Mono.defer(() -> rSocketPool.select().fireAndForget(EmptyPayload.INSTANCE))
                   .retry()
+                  .doFinally(s -> requestLatch.countDown())
```
OlegDokuka, line 67: retry() is a sync op, thus all execution should be within `RaceTestUtils.race` and once the racing is done, nothing should be working anymore, hence, I suspect, the bug is in a different place ( can we try 100k cycles on CI to ensure it is not reproing even with CDL (potentially CDL slowdowns racing, hence tests are passing)

rstoyanchev: I can remove this part. I think the bug is as I explained in https://github.com/rsocket/rsocket-java/pull/983#issuecomment-781310234. That is a clear issue.

rstoyanchev:
I undid the `CountDownLatch` and force pushed in order to see the actual change that remains. Basically fixing a `ClassCastException` from the `weightedStatsResolver`.

OlegDokuka (approved):
LGTM

Issue #970, opened 2020-12-30
SocketAcceptorInterceptor receive an unexpected payload object rather than receive a setup payload object


KaimingWan:
We create a client and server  with reconnection feature.  The server register a interceptor to receive the setup payload and then deserialized for authentication. 

Client have another daemon thread called Thread A and send data by rsocket periodically. The sended payload data we call it payload-A. 

After server is stopped and we can see the priodically sended data is failed as what we expected. After the server is started again, client begin to reconnect to the server. We can see the socket acceptor receive the setup payload data, but it is not the setup data we generated by method getAuthInfo() method. The server sidecar acceptor receive an unexpected normal payload data which is very strange. 

Related server and client code:
```java
// client 
        requesterMono = rsocketRequesterBuilder.setupRoute(SETUP_ROUTE)
            .setupData(new ObjectMapper().writeValueAsString(getAuthInfo()))
            .dataMimeType(MimeTypeUtils.parseMimeType(WellKnownMimeType.APPLICATION_JSON.getString()))
            .rsocketConnector(connector -> connector.acceptor(responder).fragment(MTU_BYTE_SIZE)
                .keepAlive(Duration.ofSeconds(KEEP_ALIVE_INTERVAL_SEC), Duration.ofSeconds(KEEP_ALIVE_MAX_TIME_SEC))
                .reconnect(Retry.fixedDelay(Integer.MAX_VALUE, Duration.ofSeconds(RSOCKET_RETRY_INTERVAL_SECONDS))
                    .doAfterRetry(retrySignal -> log.warn("RSocket client is reconnecting to get the newest connection...."))))
            .connect(TcpClientTransport.create(TcpClient.create()
                .option(ChannelOption.CONNECT_TIMEOUT_MILLIS, TCP_CONN_TIMEOUT)
                .option(ChannelOption.TCP_NODELAY, true)
                .option(ChannelOption.SO_KEEPALIVE, true)
                .host(consoleDomain)
                .port(SERVER_PORT)
                .secure(ssl -> ssl.sslContext(sslContext))));

// register a socket acceptor interceptor to receive the authentication setup payload object in the server side
    protected SocketAcceptorInterceptor genAcceptorAuthInterceptor() {
        SocketAcceptorInterceptor interceptor = acceptor -> (setup, sendingSocket) -> {
            try {
                log.info("receive an task connection try to init.");
[...]

PR #993, opened 2021-03-02
Ensure DeafultPayload#create methods copy ByteBuf content

rstoyanchev:
The RSocketConnector method that takes a setupPayload as `Payload`, as documented, tries to make a copy of a `ByteBufPayload` in order to ensure it can be re-used (e.g. on reconnect). However, `DefaultPayload#create` wasn't really making a copy. This PR fixes that and ensures that `DefaultPayload` copies data and metadata from ByteBuf's.

This is likely a fix for #970.


OlegDokuka (approved):
LGTM

Issue #752, opened 2020-03-11
Replace UnboundedProcessor with FlatMap Like Processor


OlegDokuka:
## The Problem

Right now in RSocket we use `UnboundedProcessor` which simply is an `MultiProducerSingleConsumerArrayLinkedQueue`.

It means that all frames from all possible interactions are enqueued into that queue with NO backpressure control and then underlying connection drains elements with the speed it can:
![image](https://user-images.githubusercontent.com/5380167/76427571-29d0c680-63b5-11ea-9c6b-f85521b57cfe.png)

Such a design may lead to many problems including memory leaks and general application instability 

## Proposal 

In order to have fine-grained control on all the levels, I propose to provide a customizable processor and introduce FlatMappingProcessor instead.

In a few words, FlatMappingProcessor should behave identically to FlatMap operator in ProjectReactor / RxJava

![image](https://user-images.githubusercontent.com/5380167/76428396-4a4d5080-63b6-11ea-95ee-6c301dccdee6.png)

### Main Characteristic

1. Prefetch Mode for Flux 
2. Concurrency Control (RSocket user can specify the number of active streams)
3. Small SpScBoundedArrayQueue for every Flux (`requestChannel` / `requestStream`  cases) and general scalar `MpScUnboundedLinkedArrayQueue` for `Mono` cases

PR #1003, opened 2021-04-20
replaces use of deprecated Reactor Processor API

rstoyanchev:


rstoyanchev:
The remaining uses of Reactor Processor in src/main are in `io.rsocket.internal.UnboundedProcessor` which will be done in #752 as well as in deprecated classes in `rsocket-load-balancer` module.

`ResolvingOperatorTests`  did not convert easily but is due to be replaced with JC Stress tests.

`TestDuplexConnection` also does not convert easily. It uses `DirectProcessor` + `FluxSink` and if changed to `Sinks.many().multicast().directBestEffort()` + `sink.asFlux()`, there are failures in `RSocketResponderTest`  in a couple of the `checkNoLeaksOnRacingXxx` tests. I'll need your help on the last one, @OlegDokuka.

In addition I'm not sure if the use of `tryEmitNext` might need more checks, in tests with `Sinks.Many` for example.

PR #1014, opened 2021-05-31
reworks InMemoryResumableFramesStore and improves its tests coverage

OlegDokuka:
# Main Changes Description

At the moment Resumability implementation is unstable due to its non-well-covered reworked impl.

As it was uncovered, the main root-cause of the problem is the incorrect state synchronization in InMemoryResumableFrameStore when it comes to the connection reestablishment (e.g. the previous connection was lost so we need to go through the resume handshake phase). 

As it was observed, the local producers may produce more elements while the new subscriber iterating over the cached value may miss some updates.

To resolve the mentioned problem, InmemoryResumableFrameStore is reworked again with the thought of **backpressure (via ASYNC fusion)** and fully **sequentially signals processing** (using WorkInProgress pattern). 

The first improvements allow draining elements from the upfront `Publisher` (_at the moment **UnboundedProcessor** which does not support backpressure but we should have a proper once #752 is implemented_) only when there is an active connection. In general, relying on `queue.poll` only (instead of handling data from `onNext`) decreases the amount of potentially concurrent signals we have to deal with and ensures that the new connection does not have to mess with duplicates-checking logic. 

The second improvement eliminates the need for `synchronized` keyword and replaces it with `volatile long state` machine over which we can expose various changes without introducing an expensive MpScQueue

# Side Changes Description

There are some LocalDuplexConnection and UnboundedProcessor modifications as a part of this PR. These modifications are mainly to ensure that e2e resumability tests pass with LocalTransport. `LocalDuplexConnection` (used in LocalClient/ServerTransport) embraces UnboundedProcessor to exchange frames between peers. Before these changes, it was not possible to track when all the frames are delivered / discarded. Thus, it was not possible to notify about the real termination of the DuplexConnection. The modifications to UnboundedProcessor provide a `ondisposehook` handle which allows tracking when the queue is cleaned and closed - hence notify the duplex connection that it can be terminated as well.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumableFramesStore.java
+++ b/rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumableFramesStore.java
@@ -19,118 +19,286 @@ package io.rsocket.resume;
 import static io.rsocket.resume.ResumableDuplexConnection.isResumableFrame;
 
 import io.netty.buffer.ByteBuf;
-import java.util.ArrayList;
-import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import io.netty.util.CharsetUtil;
+import java.util.ArrayDeque;
+import java.util.Queue;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 import org.reactivestreams.Subscription;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.CoreSubscriber;
+import reactor.core.Fuseable;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.Operators;
 import reactor.core.publisher.Sinks;
+import reactor.util.annotation.Nullable;
 
 /**
  * writes - n (where n is frequent, primary operation) reads - m (where m == KeepAliveFrequency)
  * skip - k -> 0 (where k is the rare operation which happens after disconnection
  */
 public class InMemoryResumableFramesStore extends Flux<ByteBuf>
-    implements CoreSubscriber<ByteBuf>, ResumableFramesStore, Subscription {
+    implements ResumableFramesStore, Subscription {
 
+  private FramesSubscriber framesSubscriber;
   private static final Logger logger = LoggerFactory.getLogger(InMemoryResumableFramesStore.class);
 
   final Sinks.Empty<Void> disposed = Sinks.empty();
-  final ArrayList<ByteBuf> cachedFrames;
-  final String tag;
+  final Queue<ByteBuf> cachedFrames;
+  final String side;
+  final String session;
   final int cacheLimit;
 
   volatile long impliedPosition;
   static final AtomicLongFieldUpdater<InMemoryResumableFramesStore> IMPLIED_POSITION =
       AtomicLongFieldUpdater.newUpdater(InMemoryResumableFramesStore.class, "impliedPosition");
 
-  volatile long position;
-  static final AtomicLongFieldUpdater<InMemoryResumableFramesStore> POSITION =
-      AtomicLongFieldUpdater.newUpdater(InMemoryResumableFramesStore.class, "position");
+  volatile long firstAvailableFramePosition;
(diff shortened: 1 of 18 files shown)


rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumableFramesStore.java
```diff
@@ -19,118 +19,240 @@
...
+      0b0100_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000L;
+  static final long FLAG_TERMINATED =
+      0b0010_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000L;
+  static final long FLAG_CONNECTED =
+      0b0001_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000L;
+  static final long FLAG_PENDING_CONNECTION =
+      0b0000_1000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000L;
+  static final long FLAG_REMOTE_IMPLIED_POSITION_CHANGED =
+      0b0000_0100_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000L;
+  static final long FLAG_HAS_FRAME =
```
rstoyanchev, line 86: `FLAG` should be at the end (e.g. `FINALIZED_FLAG`).

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumableFramesStore.java
```diff
@@ -19,118 +19,240 @@
...
+      0b0000_0010_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000L;
```
rstoyanchev, line 87: `FLAG` should be at the end (e.g. `FINALIZED_FLAG`).

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumableFramesStore.java
```diff
@@ -19,118 +19,240 @@
...
+  static final long MAX_WORK_IN_PROGRESS =
+      0b0000_0000_0000_0000_0000_0000_0000_0000_1111_1111_1111_1111_1111_1111_1111_1111L;
```
rstoyanchev, line 89: A brief explanation of each state and a diagram would serve a good purpose and should be relatively easy to add.

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumableFramesStore.java
```diff
@@ -298,30 +412,357 @@ public void request(long n) {}
...
-      actual.onSubscribe(this);
-      synchronized (this) {
-        for (final ByteBuf frame : cachedFrames) {
-          actual.onNext(frame.retainedSlice());
-        }
+    //    logger.debug(
+    //        "Tag: {}. Subscribed at Position[{}] and ImpliedPosition[{}]",
+    //        tag,
+    //        firstAvailableFramePosition,
+    //        impliedPosition);
```
rstoyanchev, line 424: Is this meant to be commented in, or perhaps done at trace?

rstoyanchev (approved):

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumableFramesStore.java
OlegDokuka: fixed

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumableFramesStore.java
OlegDokuka: improved logging

rsocket-core/src/main/java/io/rsocket/resume/InMemoryResumableFramesStore.java
OlegDokuka: done

rstoyanchev (approved):

PR #1027, opened 2021-11-09
adds first frame handling timeout

### LINKED ISSUE #1025: Limit max time for initial SETUP or RESUME frame to be received
The spec says:

> Immediately upon successful connection, the client MUST send either a SETUP or RESUME frame with Stream ID of 0. Any other frame received that is NOT a SETUP|RESUME frame or a SETUP|RESUME frame with a Stream ID > 0, MUST cause the server to send an ERROR[INVALID_SETUP] and close the connection.

The server should have a configurable limit that determines how long to wait for the SETUP or RESUME frame to materialize or close the connection otherwise. This timeout should probably have some default value that is enforced out of the box but can be customized.

**Note:** originally reported in the Spring Framework issue tracker, as https://github.com/spring-projects/spring-framework/issues/27373, along with a [repro project](https://github.com/mostroverkhov/springboot-repros) (see `/dos_idle_connections` and the `server.sh` script).

OlegDokuka:
fixes #1025 

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
@@ -233,7 +233,7 @@ public final class RSocketServer {
    * @param timeout duration
    * @return the same instance for method chaining
    */
-  public RSocketServer setupHandlingTimeout(Duration timeout) {
+  public RSocketServer maxTimeToFirstFrame(Duration timeout) {
     if (timeout.isNegative() || timeout.isZero()) {
       throw new IllegalArgumentException("Setup Handling Timeout should be greater than zero");
     }
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
@@ -41,6 +41,7 @@ import io.rsocket.plugins.InterceptorRegistry;
 import io.rsocket.plugins.RequestInterceptor;
 import io.rsocket.resume.SessionManager;
 import io.rsocket.transport.ServerTransport;
+import java.time.Duration;
 import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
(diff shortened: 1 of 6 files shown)


rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
```diff
@@ -223,6 +225,22 @@ public RSocketServer maxInboundPayloadSize(int maxInboundPayloadSize) {
     return this;
   }
 
+  /**
+   * Specifies timeout for the first incoming frame from the accepted connection.
```
rstoyanchev, line 229:
```suggestion
   * Specify the max time to wait for the first frame (e.g. {@code SETUP}) on an accepted connection.
```

rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
```diff
@@ -223,6 +225,22 @@ public RSocketServer maxInboundPayloadSize(int maxInboundPayloadSize) {
...
+   *
+   * <p>By default this is set to 1 minute.
+   *
+   * @param timeout duration
+   * @return the same instance for method chaining
+   */
+  public RSocketServer setupHandlingTimeout(Duration timeout) {
```
rstoyanchev, line 236: How about calling this `maxTimeToFirstFrame`?

rstoyanchev (commented):
Looks good. A couple of minor comments.

rsocket-core/src/main/java/io/rsocket/core/RSocketServer.java
OlegDokuka: Sound good to me!

OlegDokuka:
```suggestion
  public RSocketServer maxTimeToFirstFrame(Duration timeout) {
```

rsocket-core/src/test/java/io/rsocket/core/RSocketServerTest.java
```diff
@@ -60,6 +62,32 @@ public void unexpectedFramesBeforeSetupFrame() {
         .hasNoLeaks();
   }
 
+  @Test
+  public void timeoutOnNoFirstFrame() {
+    final VirtualTimeScheduler scheduler = VirtualTimeScheduler.getOrSet();
+    try {
+      TestServerTransport transport = new TestServerTransport();
+      RSocketServer.create().setupHandlingTimeout(Duration.ofMinutes(2)).bind(transport).block();
```
OlegDokuka, line 70:
```suggestion
      RSocketServer.create().maxTimeToFirstFrame(Duration.ofMinutes(2)).bind(transport).block();
```

PR #1029, opened 2021-11-12
eliminate boxing in RequesterResponderSupport when using IntObjectMap

olme04:
`IntObjectMap` implements `Map` and as so it has methods:
* [`default V putIfAbsent(K key, V value)`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#putIfAbsent-K-V-)
* [`default boolean remove(Object key, Object value)`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#remove-java.lang.Object-java.lang.Object-)

Those declarations uses `Object` for keys, and not specified for usage of `int` primitive as key, and as so it autobox them.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/RequesterResponderSupport.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RequesterResponderSupport.java
@@ -7,6 +7,7 @@ import io.rsocket.DuplexConnection;
 import io.rsocket.RSocket;
 import io.rsocket.frame.decoder.PayloadDecoder;
 import io.rsocket.plugins.RequestInterceptor;
+import java.util.Objects;
 import java.util.function.Function;
 import reactor.util.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


rsocket-core/src/main/java/io/rsocket/core/RequesterResponderSupport.java
```diff
@@ -118,8 +119,12 @@ public int addAndGetNextStreamId(FrameHandler frameHandler) {
...
   public synchronized boolean add(int streamId, FrameHandler frameHandler) {
-    final FrameHandler previousHandler = this.activeStreams.putIfAbsent(streamId, frameHandler);
-
+    final IntObjectMap<FrameHandler> activeStreams = this.activeStreams;
+    // copy of Map.putIfAbsent(key, value) without `streamId` boxing
+    FrameHandler previousHandler = activeStreams.get(streamId);
+    if (previousHandler == null) {
+      previousHandler = activeStreams.put(streamId, frameHandler);
+    }
     return previousHandler == null;
```
OlegDokuka, line 128:
```suggestion
    final IntObjectMap<FrameHandler> activeStreams = this.activeStreams;
    // copy of Map.putIfAbsent(key, value) without `streamId` boxing
    FrameHandler previousHandler = activeStreams.get(streamId);
    if (previousHandler == null) {
      activeStreams.put(streamId, frameHandler);
      return true;
    }
    return false;
```

OlegDokuka (approved):
LGTM with minor suggestion

rsocket-core/src/main/java/io/rsocket/core/RequesterResponderSupport.java
olme04: thx, done

PR #1036, opened 2022-02-09
First pass on Micrometer Observation API

marcingrzejszczak:
- added instrumentation for the requester and the responder
- if you run `./gradlew generateObservabilityDocs` we will generate metrics and span docs in `root/build` folder


gradle.properties
```diff
@@ -11,5 +11,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-version=1.1.2
+version=2.0.0-SNAPSHOT
```
jonatan-ivanov, line 14: Isn't this the version of rsocket-java? :)

jonatan-ivanov (approved):
I'm not an rsocket expert but it looks good.

gradle.properties
marcingrzejszczak: Yeah I thought that this might be released as RSocket 2.0.0 cause we're introducing micrometer that introduces breaking changes. We'll see if it actually remains 2.0.0 or 1.2

OlegDokuka:
superseded by #1056

PR #1063, opened 2022-08-30
introduces `onClose` listener for RSocketClient and `connect` method

### LINKED ISSUE #1048: Support close in RSocketClient that notifies of the result
`RSocketClient` implements `Disposable` which provides a way to initiate a shutdown. However, there is no good way to hold for the result, and it is important in some cases to wait for that. For example, in components that expose start/stop lifecycle methods and could be started and stopped multiple times. Likewise in tests to ensure resources are released before continuing.

By comparison `RSocket` implements `Closeable` so technically possible to access the underlying source `RSocket` and call `close()` on that but that could cause a session to be started if not already connected. It would be better to align the two and expose a similar option at the level of `RSocketClient` as well.

OlegDokuka:
closes #1048 

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketClient.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketClient.java
@@ -77,6 +77,17 @@ import sun.reflect.generics.reflectiveObjects.NotImplementedException;
  */
 public interface RSocketClient extends Closeable {
 
+  /**
+   * Connect to the remote rsocket endpoint, if not yet connected. This method is a shortcut for
+   * {@code RSocketClient#source().subscribe()}.
+   *
+   * @return {@code true} if an attempt to connect was triggered or if already connected, or {@code
+   *     false} if the client is terminated.
+   */
+  default boolean connect() {
+    throw new NotImplementedException();
+  }
+
   default Mono<Void> onClose() {
     return Mono.error(new NotImplementedException());
   }
--- a/rsocket-core/src/main/java/io/rsocket/core/RSocketClient.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/RSocketClient.java
@@ -15,12 +15,13 @@
  */
 package io.rsocket.core;
 
+import io.rsocket.Closeable;
 import io.rsocket.Payload;
 import io.rsocket.RSocket;
 import org.reactivestreams.Publisher;
-import reactor.core.Disposable;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
 
 /**
  * Contract for performing RSocket requests.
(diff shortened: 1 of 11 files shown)


rstoyanchev (commented):
Looks good. Only a couple of small comments for the `connect` method, which could also be split out into a separate commit as it is orthogonal to the `onClose` I think.

rsocket-core/src/main/java/io/rsocket/core/RSocketClient.java
```diff
@@ -74,7 +75,15 @@
  * @since 1.1
  * @see io.rsocket.loadbalance.LoadbalanceRSocketClient
  */
-public interface RSocketClient extends Disposable {
+public interface RSocketClient extends Closeable {
+
+  default boolean connect() {
+    return false;
+  }
```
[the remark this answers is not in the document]
OlegDokuka: fixed

OlegDokuka:
@rstoyanchev added 2 separate commits as it was suggested

rsocket-core/src/main/java/io/rsocket/core/RSocketClient.java
```diff
@@ -74,7 +75,21 @@
  * @since 1.1
  * @see io.rsocket.loadbalance.LoadbalanceRSocketClient
  */
-public interface RSocketClient extends Disposable {
+public interface RSocketClient extends Closeable {
+
+  /**
+   * Tries to connect to remote rsocket endpoint if not yet connected. This method is a
+   * shortcut to {@code RSocketClient#source().subscribe()}
```
rstoyanchev, line 82:
```suggestion
   * Connect to the remote rsocket endpoint, if not yet connected. This method is a
   * shortcut for {@code RSocketClient#source().subscribe()}.
```

rsocket-core/src/main/java/io/rsocket/core/RSocketClient.java
```diff
@@ -74,7 +75,21 @@
...
+   *
+   * @return {@code true} if successfully activated connection process
```
rstoyanchev, line 84:
```suggestion
   * @return {@code true} if an attempt to connect was triggered, or {@code false}
   * if already connected or if the client is terminated.
```

rsocket-core/src/main/java/io/rsocket/core/RSocketClientAdapter.java
```diff
@@ -41,11 +41,22 @@ public RSocket rsocket() {
     return rsocket;
   }
 
+  @Override
+  public boolean connect() {
+    throw new UnsupportedOperationException(
+        "Serverside RSocketClientAdapter can not proactively open a connection");
```
rstoyanchev, line 47:
```suggestion
        "Connect does not apply to a server side RSocket");
```

rstoyanchev (approved):

PR #1076, opened 2022-10-26
allows continuation of observations

marcingrzejszczak:
without this change we're not taking into account a parent observation into account. Nor do we set in the reactor context the current observation for the users to use

with this change whenever an observation is being created we put it into the reacto context under a well-known micrometer.observation key (for more information look at the ObservationThreadLocalAccessor class from micrometer-core)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -11,5 +11,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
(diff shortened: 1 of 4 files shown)


OlegDokuka:
as discussed with the team we will roll it into 1.2.0 instead. We will revert this commit for 1.1.x

PR #1078, opened 2022-11-11
Provides GracefulShutdown functionality 

### LINKED ISSUE #988: Support for graceful shutdown
Is there a way to wait for RSocket requests to finish before cutting them off at the knees?

> Oleh Dokuka  < 1 minute ago

We need to add that. Spec has such a feature, however we still does not support it.

OlegDokuka:
closes #988 


rstoyanchev (commented):
Great changes! 

There are 5 sinks passed around for coordination, in addition to 2 more `Mono`s for when both sides are completed or closed. It takes some effort to trace their usages and I'm thinking we encapsulate this behavior a bit to make the code more readable and intuitive.

As far as I can see this is primarily requester side driven, but the responder does need to know when the graceful shutdown starts, as well as to indicate back to the requester when it's done with its part. 

Let's say we have a package private contract like this:
```java
interface GracefulShutdownTracker {

  Mono<Void> onGracefulShutdownStarted();

  Sinks.Empty<Void> responderCompletedSink();

  Sinks.Empty<Void> responderClosedSink();

}
```

Then inside `RSocketRequester`:

```java
class RSocketRequester extends RequesterResponderSupport implements RSocket {

  private final DefaultGracefulShutdownTracker shutdownTracker = new DefaultGracefulShutdownTracker();

  // ...

  public GracefulShutdownTracker getGracefulShutdownTracker() {
    return gracefulShutdownTracker;
  }


  private final static class DefaultGracefulShutdownTracker implements GracefulShutdownTracker {

    private final Sinks.Empty<Void> startedSink = Sinks.unsafe().empty();

    private final Sinks.Empty<Void> completedSink = Sinks.unsafe().empty();
    private final Sinks.Empty<Void> closedSink = Sinks.unsafe().empty();

    private final Sinks.Empty<Void> responderCompletedSink = Sinks.unsafe().empty();
    private final Sinks.Empty<Void> responderClosedSink = Sinks.unsafe().empty();

    public Mono<Void> onGracefulShutdownStarted() {
      return startedSink.asMono();
    }

    @Override
    public Sinks.Empty<Void> responderCompletedSink() {
      return responderCompletedSink;
    }

    @Override
    public Sinks.Empty<Void> responderClosedSink() {
      return responderClosedSink;
    }

    Mono<Void> onBothCompleted() {
      return Mono.whenDelayError(
              responderCompletedSink.asMono(),
              completedSink.asMono());
    }

    Mono<Void> onBothClosed() {
      return Mono.whenDelayError(
              responderClosedSink.asMono(),
              closedSink.asMono());
    }
  }

}
````

Now `RSocketRequester` has all the Sinks and Monos it needs access internally to do its work, and also has a getter for `GracefulShutdownControl` that exposes all that's needed for the responder to do its work.

rstoyanchev (commented):
One more note. Currently, `RequestResponderSupport` has an `onGracefulShutdownSink` as a constructor initiailized field. 

In my proposal above, this could be replaced with an abstract method, something like `void GracefulShutdownCompleted()` which the requester and the responder would implemented accordingly based on what's available to them.

Issue #1077, opened 2022-11-01
SendUtils onDiscard handler class cast exception


Svenskunganka:
I'm using RSocket Java with Spring Webflux and RSocket JS in the front-end, using `REQUEST_STREAM` with an infinite Kafka source. The returned Flux receives quite a lot of small messages and does quite a bit of filtering and buffering. [The relevant code in `SendUtils`](https://github.com/rsocket/rsocket-java/blob/e31038f04e8bfc7b9c3a9a2b3a12fd578ed02a40/rsocket-core/src/main/java/io/rsocket/core/SendUtils.java#L41-L51) does an unchecked cast of each discarded object to a `ReferenceCounted` which seems to throw a class cast exception that is caught and ignored.  
The problem is that due to the sheer amount of objects being discarded (due to filtering and discarded buffers on disconnect), this uses up a substantial amount of CPU resources looking at the profiler.

![profiler](https://user-images.githubusercontent.com/2657513/199223616-fb5f8086-566b-4554-b57c-af3605ce1b62.png)

The returned Fluxes are scheduled on `Schedulers.boundedElastic()`, and for some reason entire threads are started up just for this exception:
![flamegraph](https://user-images.githubusercontent.com/2657513/199225004-be4db466-4e3d-4ad7-8da6-a42274009945.png)

## Expected Behavior
Not use up so much CPU resources

## Actual Behavior
Uses a lot of CPU resources

## Possible Solution
Not exactly sure, but perhaps an `instanceof` check:
```java
if (data instanceof ReferenceCounted) ((ReferenceCounted) data).release();
```

## Your Environment

* RSocket version(s) used: rsocket-core:1.1.3, rsocket-transport-netty:1.1.3
* Other relevant libraries versions (eg. `netty`, ...): Spring Webflux 5.3.23
* Platform (eg. JVM version (`javar -version`) or Node version (`node --version`)): 17
* OS and version (eg `uname -a`): Happens on all tested platforms (Debian, Mac)

OlegDokuka:
@Svenskunganka! Sounds relevant! 

Feel free to contribute if you see the direct solution, otherwise, I will investigate it!

Best,
Oleh

Svenskunganka:
Hi Oleh!

If my possible solution is acceptable I can definitely contribute it! But if you perhaps know of another way this could be solved without needing to perform an `instanceof` check for every discarded object that might be a better route to explore.  
[...]

PR #1079, opened 2022-12-01
#1077 type safe discard element consumer

luczus:
Discard element consumer is accepting other elements than ReferenceCounted

### Motivation:
When I use DefaultRSocketClient in logs files I see warning log entries like:

```
] WARN  reactor.core.publisher.Operators        : Error in discard hook
java.lang.ClassCastException: class org.springframework.core.io.buffer.NettyDataBuffer cannot be cast to class io.netty.util.ReferenceCounted (org.springframework.core.io.buffer.NettyDataBuffer and io.netty.util.ReferenceCounted are in unnamed module of loader 'app')
```

### Modifications:
Make discard consumer accept other types than ReferenceCounted parameter and avoid having warnings in logs

### Result:

The warning log entries are not created by DefaultRSocketClient.

### DIFF
--- a/rsocket-core/src/main/java/io/rsocket/core/DefaultRSocketClient.java
+++ b/rsocket-core/src/main/java/io/rsocket/core/DefaultRSocketClient.java
@@ -22,6 +22,7 @@ import io.rsocket.RSocket;
 import io.rsocket.frame.FrameType;
 import java.util.AbstractMap;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


kmccarp:
Is there any plan to integrate this? This is heavily polluting our logs.

OlegDokuka:
close in favor of #1091

PR #1106, opened 2024-02-23
updates versions and fixes memleak in UnboundedProcessor

OlegDokuka:
This PR updates dependencies and makes minor modifications to `UnboundedProcessor` due to repeating failures of UnboundedProcessorJCStreassTest, which started reproducing some unspotted issues. 

### Motivation:

UnboundedProcessor is a critical component in the RSocket-Java ecosystem and must work properly. After analysis of its internal state machine, it was spotted that sometimes:
1. Request may not be delivered due to natural concurrency
2. Terminal signal may not be delivered since it checks for demand which might be consumed already (due to natural concurrency) 
3. Final value could be delivered violating reactive-streams spec 

### Modifications:

This PR adds a minimal set of changes, preserving old implementation but eliminating the mentioned bugs 

### Result:

_[After your change, what will change.]_


whyoleg (approved):
LGTM

PR #1118, opened 2025-01-31
ensures connection is closed on keepalive timeout

### LINKED ISSUE #1099: Server does not close and re-connect on missed keep alive ack
Using the server setup found [here](https://github.com/rsocket/rsocket-java/issues/987#issuecomment-796176003) and using rsocket-core 1.1.3 my server behaves as expected. If for some reason no acks get delivered during the keep-alive timeout it will restart the connection. In 1.1.4 this no longer works.

## Expected Behavior
After:

```
io.rsocket.exceptions.ConnectionErrorException: No keep-alive acks for x ms
```

I expect that the connection is closed and my client can reconnect.

## Actual Behavior

In 1.1.4 the client RSocketRequester does not properly reset the connection and the next call fails with the same `ConnectionErrorException`

## Steps to Reproduce
I've created a [small reproducing example](https://github.com/pstreef/rsocket-reconnect-issue). The test in this projects runs green using rsocket-core 1.1.3, but fails when using 1.1.4

## Possible Solution
I believe this is regression from [this](https://github.com/rsocket/rsocket-java/pull/1085) change.

## Your Environment
See sample project

* RSocket version(s) used: 1.1.4 (and 1.1.3)
* Other relevant libraries versions (eg. `netty`, ...): spring-boot-starter-rsocket 3.2.0
* Platform (eg. JVM version (`javar -version`) or Node version (`node --version`)): zulu 17.0.8
* OS and version (eg `uname -a`): macos 14.2 (23C64)

OlegDokuka:
closes #1099

### DIFF
--- a/rsocket-transport-netty/src/test/java/io/rsocket/integration/KeepaliveTest.java
+++ b/rsocket-transport-netty/src/test/java/io/rsocket/integration/KeepaliveTest.java
@@ -1,9 +1,5 @@
 package io.rsocket.integration;
 
-import java.time.Duration;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.function.Function;
-
 import io.rsocket.Payload;
 import io.rsocket.RSocket;
 import io.rsocket.core.RSocketClient;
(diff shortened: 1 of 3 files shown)


rsocket-transport-netty/src/test/java/io/rsocket/integration/KeepaliveTest.java
```diff
@@ -0,0 +1,188 @@
...
+    //				.flatMap(RSocket::onClose)
+    //				.doOnError(err -> LOG.error("Error during onClose.", err))
+    //				.retryWhen(reconnectSpec.apply("client-close"))
+    //				.doFirst(() -> LOG.info("Connected on client side."))
+    //				.doOnTerminate(() -> LOG.info("Connection closed on client side."))
+    //				.repeat()
+    //				.subscribe();
+
+    //		return client;
+  }
```
rstoyanchev, line 179: The createClient methods maybe can also be combined into one with a `lazy` input argument.

rstoyanchev (changes requested):
A couple of improvements suggested below for the test class, but otherwise looks good.

